PR #186, opened 2012-09-12
kvm plugins improvements

duritong:
I added a few improvements to the kvm plugins:
- Make most of them also work on RedHat based systems
- Add an additional kvm_net_rh Plugin that works on RedHat bases systems. It is more or less a copy of the existing kvm_net plugin, which does not work and which needed to many changes which I'm not sure would have broken things further.

### DIFF
--- a/plugins/virtualization/kvm_cpu
+++ b/plugins/virtualization/kvm_cpu
@@ -38,7 +38,7 @@ graph_args --base 1000 -r --lower-limit 0 --upper-limit %d""" % percent
         print "%s_cpu.type DERIVE" % vm
         print "%s_cpu.draw %s" % (vm, draw)
         print "%s_cpu.info percent of cpu time used by virtual machine" % vm
-	draw = "STACK"
+        draw = "STACK"
 
 
 def clean_vm_name(vm_name):
@@ -57,9 +57,12 @@ def clean_vm_name(vm_name):
 def detect_kvm():
     ''' Check if kvm is installed
     '''
-    kvm = Popen("which kvm", shell=True, stdout=PIPE)
-    kvm.communicate()
-    return not bool(kvm.returncode)
+    if os.path.isfile("/usr/libexec/qemu-kvm"):
+        return True
+    else:
+        kvm = Popen("which kvm", shell=True, stdout=PIPE)
+        kvm.communicate()
+        return not bool(kvm.returncode)
 
 def find_vm_names(pids):
     '''Find and clean vm names from pids
@@ -70,12 +73,12 @@ def find_vm_names(pids):
         cmdline = open("/proc/%s/cmdline" % pid, "r")
         result[pid] = clean_vm_name(re.sub(r"^.*-name\x00([a-zA-Z0-9.-_-]*)\x00\-.*$",r"\1", cmdline.readline()))
     return result
-    
+
 def list_pids():
     ''' Find the pid of kvm processes
     @return a list of pids from running kvm
     '''
-    pid = Popen("pidof qemu-system-x86_64", shell=True, stdout=PIPE)
+    pid = Popen("pidof qemu-kvm qemu-system-x86_64 kvm", shell=True, stdout=PIPE)
     return pid.communicate()[0].split()
 
 def fetch(vms):
@@ -85,7 +88,7 @@ def fetch(vms):
     for ( pid, name ) in vms.iteritems():
         ( user, system ) = open("/proc/%s/stat" % pid, 'r').readline().split(' ')[13:15]
         print '%s_cpu.value %d' % ( name, int(user) + int(system) )
-        
+
 if __name__ == "__main__":
     if len(sys.argv) > 1:
         if sys.argv[1] in ['autoconf', 'detect']:
@@ -99,4 +102,3 @@ if __name__ == "__main__":
             fetch(find_vm_names(list_pids()))
     else:
         fetch(find_vm_names(list_pids()))
-
--- a/plugins/virtualization/kvm_io
+++ b/plugins/virtualization/kvm_io
@@ -37,7 +37,7 @@ graph_args --base 1024
         print "%s_read.info I/O used by virtual machine %s" % (vm, vm)
         print "%s_write.label %s" % (vm, vm)
         print "%s_write.type COUNTER" % vm
-        print "%s_write.min 0" % vm        
+        print "%s_write.min 0" % vm
         print "%s_write.draw LINE1" % vm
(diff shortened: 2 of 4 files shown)


sumpfralle:
@duritong: are you still interested in this pull request? What do you think about the issues above?

duritong:
So I reworked the whole stuff and I hope problems to be solved by now

plugins/virtualization/kvm_cpu
```diff
@@ -57,9 +57,12 @@ def clean_vm_name(vm_name):
...
     '''
-    kvm = Popen("which kvm", shell=True, stdout=PIPE)
-    kvm.communicate()
-    return not bool(kvm.returncode)
+    if os.path.isfile("/usr/libexec/qemu-kvm"):
+        return True
+    else:
+        kvm = Popen("which kvm", shell=True, stdout=PIPE)
+        kvm.communicate()
+        return not bool(kvm.returncode)
```
sumpfralle: I would propose to phrase this clearer (instead of relying on the coincidence of the implicit interger/boolean conversion):
`return kvm.returncode == 0`

plugins/virtualization/kvm_cpu
```diff
@@ -70,12 +73,12 @@ def find_vm_names(pids):
...
         result[pid] = clean_vm_name(re.sub(r"^.*-name\x00([a-zA-Z0-9.-_-]*)\x00\-.*$",r"\1", cmdline.readline()))
     return result
-    
+
 def list_pids():
     ''' Find the pid of kvm processes
     @return a list of pids from running kvm
     '''
-    pid = Popen("pidof qemu-system-x86_64", shell=True, stdout=PIPE)
+    pid = Popen("pidof qemu-kvm qemu-system-x86_64 kvm", shell=True, stdout=PIPE)
```
sumpfralle: Does this list of names lack "qemu-system-i386"?
Other architectures are probably not available due to kvm being bound to intel/amd - right?

plugins/virtualization/kvm_io
```diff
@@ -68,15 +68,18 @@ def fetch(vms):
...
     '''
-    kvm = Popen("which kvm", shell=True, stdout=PIPE)
-    kvm.communicate()
-    return not bool(kvm.returncode)
+    if os.path.isfile("/usr/libexec/qemu-kvm"):
+        return True
+    else:
+        kvm = Popen("which kvm", shell=True, stdout=PIPE)
+        kvm.communicate()
+        return not bool(kvm.returncode)
```
sumpfralle: same as above

plugins/virtualization/kvm_io
```diff
@@ -87,14 +90,14 @@ def find_vm_names(pids):
...
         result[pid] = clean_vm_name(re.sub(r"^.*-name\x00([a-zA-Z0-9.-_-]*)\x00\-.*$",r"\1", cmdline.readline()))
     return result
-    
+
 def list_pids():
     ''' Find the pid of kvm processes
     @return a list of pids from running kvm
     '''
-    pid = Popen("pidof qemu-system-x86_64", shell=True, stdout=PIPE)
+    pid = Popen("pidof qemu-kvm qemu-system-x86_64 kvm", shell=True, stdout=PIPE)
```
sumpfralle: same as above

plugins/virtualization/kvm_mem
```diff
@@ -61,19 +61,22 @@ def fetch(vms):
...
     '''
-    kvm = Popen("which kvm", shell=True, stdout=PIPE)
-    kvm.communicate()
-    return not bool(kvm.returncode)
+    if os.path.isfile("/usr/libexec/qemu-kvm"):
+        return True
+    else:
+        kvm = Popen("which kvm", shell=True, stdout=PIPE)
+        kvm.communicate()
+        return not bool(kvm.returncode)
```
sumpfralle: same as above

plugins/virtualization/kvm_mem
```diff
@@ -84,14 +87,14 @@ def find_vm_names(pids):
...
         result[pid] = clean_vm_name(re.sub(r"^.*-name\x00([a-zA-Z0-9.-_-]*)\x00\-.*$",r"\1", cmdline.readline()))
     return result
-    
+
 def list_pids():
     ''' Find the pid of kvm processes
     @return a list of pids from running kvm
     '''
-    pid = Popen("pidof qemu-system-x86_64", shell=True, stdout=PIPE)
+    pid = Popen("pidof qemu-kvm qemu-system-x86_64 kvm", shell=True, stdout=PIPE)
```
sumpfralle: same as above

plugins/virtualization/kvm_cpu
duritong: I didn't change the original behavior. Just added one that fits more for RHEL/CentOS.
In general this part is from my point of view problematic in 2 cases:

1. which lookups are expensive
2. we're not only looking at kvm.

I can change it in general, but as I said the intend of my PR was not to change the original behavior. But I can come up with general improvements, if people would like to see that.

plugins/virtualization/kvm_cpu
duritong: As mentioned above I did not change the original behavior. So I just kept `qemu-system-x86_64` as it was the current behavior. I can add though other architectures as well.

duritong:
Regarding the RH specific plugin: I have only rhel based kvm boxes and there were some significant differences, I wasn't able to test on debian based kvm boxes. So either someone needs to test it or we should keep it as seperate plugins.

I agree though, that a unified plugin would be better, but given I don't have access to debian based kvm boxes I can't really verify my changes.

sumpfralle:
I would be happy, if you could change the bits necessary for your environment and also fix/improve more details along the way (in separate commits). Most of the plugins do not have a fixed maintainer and cannot be easily tested by the repository maintainers - thus it is usually highly welcome to combine fixes with general cleanups.

Regarding Debian/Red Hat: I have access to Debian systems with kvm and I am fluent in python. Thus I would be happy to test and fix (if necessary) the plugins, as soon as you included the Red Hat specific parts.

sumpfralle:
@duritong: what do you think?

duritong:
I'll try to work on it, december was a bit a busy month for me.

sumpfralle:
@duritong: just a friendly (and patient) reminder ...

sumpfralle:
@duritong: ping?

sumpfralle:
Ping?

Neustradamus:
@duritong: What is the situation of your PR?

PR #534, opened 2014-09-29
Avoid bashism

asgh:
This was changed in #430 but it's not legal for `/bin/sh`

### DIFF
--- a/plugins/gpu/nvidia_gpu_
+++ b/plugins/gpu/nvidia_gpu_
@@ -198,7 +198,7 @@ case $name in
 			totalMemGpu=`echo "$totalMemGpus" | sed -n $(( $nGpusCounter + 1 ))p`
 			usedMemGpu=`echo "$usedMemGpus" | sed -n $(( $nGpusCounter + 1 ))p`
 			percentMemUsed=$(( $usedMemGpu * 100 / $totalMemGpu ))
-			valueGpus="${valueGpus}${percentMemUsed}"$'\n'
+			valueGpus="${valueGpus}${percentMemUsed}"$(printf '\n')
 			: $(( nGpusCounter = $nGpusCounter + 1 ))
 		done
 		;;


leeclemens:
Seriously, why do we keep breaking this plugin?

asgh:
@leeclemens Care to elaborate? What's broken? If it's this change, it was tested by a few people and it works.

leeclemens:
The fix regarding the newline I added works - with /bin/bash. The shebang was updated to /bin/bash, but your commit to use printf was also committed. These seem to have been incompatible changes, as it seems to have broken using /bin/bash (with the printf).

What shell are you using where printf works? And have you tested with one and more than one GPU?

Referring to https://github.com/munin-monitoring/contrib/blob/master/plugins/gpu/nvidia_gpu_#L201 nvidia_gpu_mem

leeclemens:
There was only a day between the shebang and your printf change being merged, did you test with #!/bin/bash prior to your commit 2b3c17b97758b5fd29a6f123f3902e0b0b45aa1b ?

leeclemens:
Cross ref - https://github.com/munin-monitoring/contrib/pull/569

asgh:
Changing to `/bin/bash` would also fix this, but it's preferred not to use bash for scripts, only for an actual shell, so I prefer the change that makes `/bin/sh` work.

What error are you getting on redhat?

asgh:
And yes, it makes no sense to merge a commit that avoids a bashism and simultaneously change to bash. It's one or the other.

leeclemens:
I agree, I'd like to avoid /bin/bash specific changes as well. However, I have not seen a version that works with two GPUs and /bin/sh. Can you propose one? I'm still not sure what shell you're using or how many GPUs you have tested this plugin with - would you please answer those previously asked questions?

The error is related to the previous newline fix I pushed - I'm not sure what else you are looking for.

I provided a solution that works (plus the shebang being updated) - for some reason it had been reverted by @ssm 's merges with no details (what shell? how many GPUs) being provided.

asgh:
I'm using `dash` actually. (It's a more secure variant of `sh` that Debian uses.) Only one GPU.

When I applied your fix, using `sh` it failed, switching to `printf` instead and it worked and I did not check any further than that.

asgh:
Can you try it, using `/bin/sh`, with `$(printf '\n\n')` instead of a single `\n' ?

leeclemens:
On Ubuntu, with /bin/dash, Using printf '\n\n' does not solve the original issue. Tried using "valueGpus="${valueGpus}${percentMemUsed}"$(printf '\n\n')"

iirc, this update is only relevant when two or more GPUs exist. I'm not sure how your testing can be relevant. The issue with with internal variables, not printing values.

Using Ubuntu (/bin/dash), I can confirm with two GPUs this 'fix' does not work. However, using /bin/bash and the commit which was previously merged does still work.

asgh:
OK, my testing wasn't relevant. Don't be annoyed about it, let's just fix it. I did not spend a lot of time on this, I just saw the bashism, looked up what to change and that's it.

When I run:

    echo $'\n' | od -b -A n|sed 2d

And

    printf '\n\n' | od -b -A n|sed 2d

In `bash` I get identical results (and the 2nd one is identical with `sh`), so that should be the fix. If not, can you maybe run with `set -x` both ways so we can see what's different?

leeclemens:
I'm with you (just was admittedly annoyed by fixing something and having it unfixed - I hope you understand). I did a bunch of testing similar to what you have here and can say the solution I came up with (requiring /bin/bash) worked - I hope I missed something.

[user@host ~]# echo $'\n' | od -b -A n|sed 2d
 012 012
[user@host ~]# eprintf '\n\n' | od -b -A n|sed 2d
 012 012

However, I would rather have something that requires bash and works than something that does not work. My initial attempts all used posix/sh solutions. On the other hand, if you are not testing with two+ GPUs, you are not testing the issue.

I believe it is a variable replacement or argument interpretation issue vs  something you can echo two values of and compare them. I realize I am not explaining the problem as best as possible.

I ran `set -x` and then munin-run, but the only change was this line "+ munin-run nvidia_gpu_mem". I'm  not sure what you intended regarding `set -x` - can you provide more instruction?

asgh:
`set -x` outputs every line, what it runs and what the result was. It should help find what's different between the command. But you say that it's identical from `sh` and `bash`? That doesn't make sense! In that case why is one working, and one not?

What does it do/say when it doesn't work? Check just the simple output of `munin-run` both ways, without the `set -x` - what's the difference?

kenyon:
@asgh It appears that you don't understand the original purpose of the bashism. That line is intended to generate a string that has percentMemUsed for each GPU on a separate line. The replacement code you gave with printf does not do that because dash does not have such a feature yet, as far as I can tell. See http://austingroupbugs.net/view.php?id=249 for more information.

In my opinion, this plugin is complex enough that it probably should not have been written in any shell language in the first place. It should have been perl, python, ruby, etc., some language with proper string handling features. But since it was originally written for bash, and explicitly uses bash's features, just leave it at bash. There is no reason to spend effort on porting this to plain sh, in this case. Is anybody going to run this plugin on a system that lacks bash?

leeclemens:
Thank you @kenyon. I feel I lacked a detailed enough grasp of bash intrinsics to explain what I discovered via previous research and trial-and-error to come up with my 'solution' for the multiple-GPU issue (and solution) I encountered.

PR #540, opened 2014-10-06
Fix failure to monitor bean counters for servers with long names.

qris:
`vzlist -a -H --no-trim -o hostname` doesn't truncate hostnames, but
`vzlist -a -H --no-trim -o hostname,numflock` does. So the data values being
output by the plugin were truncated, and some didn't match any of the counter
names output by "openvz config", so no data was stored for those counters.

I'm also adding an underscore between the counter name and the server name,
for readability.


ssm:
The bugfix should not rewrite any working labels. That will remove historic data from the graphs.

sumpfralle:
This pull-request contains a breaking change of datasource names - thus it is not suitable for inclusion.
I created a separate patch containing only the proposed bugfix instead (b2a5b603e1523ea0f900d037376885e8535f1481).
Thank you for your submission!

PR #546, opened 2014-10-30
Cleaned up directory structure of mail related plugins

stoile:
Al mail plugins I could identify are now in the mail directory.


kenyon:
Seems like a good change to me.

sumpfralle:
Replaced with #906 (due to loads of changes since the initial proposal)

PR #562, opened 2014-12-25
Added zfsonlinux_stats plugin

elfurbe:
I wanted to get all these ZFS stats on my ZFSonLinux install, so I modified the FreeBSD plugin to read from the stats files generated by the SPL module instead. I also added some checks to make sure I wasn't dividing by zero in a number of places. From an output perspective, this produces the same values as the zfs_stats plugin.


elfurbe:
It looks like when you merged the branch for #563 it actually removed this plugin cause I didn't branch my commmits properly. I'm not entirely sure how to resolve it, but I thought I'd bring it to your attention. If there's something helpful I can do, let me know.

PR #566, opened 2015-01-02
modified data_url for firmware >= 6.20

laserb:
With the latest firmware update to 6.20 on the fritzbox 7390, the data fetching was broken. The url with the data could not be found anymore.
This corrects the url.


ssm:
This would break the plugin for anyone staying with the old firmware.

Change to reflect current firmware?
Add an environment variable for the URL base which defaults to the "current" one?

rrzefox:
Another idea could be to just try both. Try to fetch the new URL first, if that fails, retry with the old one.

ssm:
Then again, this is "contrib", and we do not automatically install or upgrade these plugins anywhere.  :)

PR #567, opened 2015-01-08
Fix hit rate calculation for varnish4 plugin

ioppermann:
The hit rate caluclation is correct as long as no purges are performed. If purges are performed, they are counted in client_req, but not counted in cache_(hit|miss|hitpass). This would lead to wrong values for the hit rate because client_req is bigger than the sum of cache_(hit|miss|hitpass).

Additionally there are some minor bugfixes.

### DIFF
--- a/plugins/varnish4/README.md
+++ b/plugins/varnish4/README.md
@@ -8,7 +8,7 @@ Installation
 ------------
 
 - Copy `varnish4_` to the Munin plugins directory (`/usr/share/munin/plugins/`)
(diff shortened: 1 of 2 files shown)


driskell:
This looks good! Thanks - I'm using Varnish4 plugin too.

Be great if you could rebase as requested by @ssm so they can merge it. That or I'd be willing to take over with my own PR if you want to close this one?

ioppermann:
I did the rebase of this pull request.

PR #568, opened 2015-01-13
Git munin

syswombat:


syswombat:
# Modicication data/reason: 12.01.2015 / removed all Max UP and Max Down and created new file

PR #569, opened 2015-01-14
Revert "Avoid bashism"

leeclemens:
The "Avoid bashism" commit does not work with /bin/bash (including RedHat, etc where /bin/sh is symlinked to /bin/bash). A subsequent commit to fix the shebang to /bin/bash solved the issue (for Debian/Ubuntu at least, theoretically everywhere).


leeclemens:
Can this be merged, given @kenyon 's comment in PR https://github.com/munin-monitoring/contrib/pull/534?

PR #570, opened 2015-01-14
Added nscd plugin

weeheavy:
A plugin to monitor nscd databases.


steveschnepp:
Merged, thx.

As it seems it doesn't need the bashisms, i'd welcome a patch to use `/bin/sh` instead :smiling_imp:

PR #573, opened 2015-01-21
Fix regexp in case iptables returns huge numbers. It may not have spaces...

HeisSpiter:
... at the begin of the line.

Make use of state file so that the plugin doesn't show an increasing jauge, but a current network usage


steveschnepp:
Anyway, I won't delay merge for that for too long, as it is contrib & the plugin works.

Just open another PR to fix those.

PR #574, opened 2015-01-22
Use MUNIN_STATEFILE env variable to support multiple masters instead of ...

HeisSpiter:
...own made.

Suggested by Steve Schnepp


steveschnepp:
:+1:

PR #590, opened 2015-02-22
vnstat_month: suppress warnings in perl 5.18+

dgilman:
This broke in an update to Debian Jessie.


dgilman:
@ssm does my travis.yml dependency need to be alphabetized?  does it go under XML:: or Device::?

PR #598, opened 2015-03-09
Update jabber/network plugins

dotdoom:
* add mnesia support to ejabberd
* use erl_call instead of hacky erl -c
* fix ping plugin logic


dotdoom:
Ping

steveschnepp:
better come over IRC for pinging us ;)

PR #599, opened 2015-03-09
Add TITLE_PREFIX to redis_

driskell:
This allows setup of multiple redis instances on a single server instance and allows the graphs to be distinguishable.

Currently if you setup multiple by using instance1_redis_connections and instance2_redis_connections both set of graphs have the same name.


ssm:
There are a lot of strange whitespace changes in this commit.  Could you make one with just changes to the code, please?

driskell:
Whoops sorry will sort that.

Might be worth some separate commits to sort the spacing my editor seems too smart and keeps fixing them! (Atom.io)

driskell:
@ssm All whitespace changes are now removed - apologies for that. This is the bare minimum change now.

ssm:
Thanks :)

PR #600, opened 2015-03-09
Improve vsftpd plugin

driskell:
- Allow log file path to be configured via plugin-conf.d env.logfile
- Add counters for successful and failed renames
- Switch to use DERIVE and min 0 to fix problems during rotation and show request counts per period
- Improve performance for large log files by using awk

### DIFF
--- a/plugins/ftp/vsftpd
+++ b/plugins/ftp/vsftpd
@@ -1,58 +1,104 @@
-LOGFILE=/var/log/vsftpd.log
+#!/bin/bash
+
+logfile=${logfile:-/var/log/vsftpd.log}
 
 if [ "$1" = "autoconf" ]; then
-    if [ -f "${LOGFILE}" ]; then
-        echo yes
-	exit 0
-    else
-        echo no
-	exit 1
-    fi
+	if [ -f "${logfile}" ]; then
+		echo yes
+		exit 0
+	else
+		echo no
+		exit 1
+	fi
 fi
 
 if [ "$1" = "config" ]; then
-        echo 'graph_title FTP Server'
-        echo 'graph_args --base 1000 -l 0'
-        echo 'graph_vlabel Requests'
-	echo 'graph_category FTP'
-        echo 'ftp_c.label connections'
-        echo 'ftp_sl.label successful_logins'
-	echo 'ftp_fl.label failed_logins'
-	echo 'ftp_su.label successful_uploads'
-	echo 'ftp_fu.label failed_uploads'
-	echo 'ftp_sd.label successful_downloads'
-	echo 'ftp_fd.label failed_downloads'
-	echo 'ftp_sde.label successful_deletes'
-	echo 'ftp_fde.label failed_deletes'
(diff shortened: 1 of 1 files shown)


plugins/ftp/vsftpd
```diff
@@ -34,18 +62,24 @@ ftp_su=U
...
+	ftp_fl=$(grep -c "FAIL LOGIN" "${logfile}")
+	ftp_su=$(grep -c "OK UPLOAD" "${logfile}")
+	ftp_fu=$(grep -c "FAIL UPLOAD" "${logfile}")
+	ftp_sd=$(grep -c "OK DOWNLOAD" "${logfile}")
+	ftp_fd=$(grep -c "FAIL DOWNLOAD" "${logfile}")
+	ftp_sr=$(grep -c "OK RENAME" "${logfile}")
+	ftp_fr=$(grep -c "FAIL RENAME" "${logfile}")
+	ftp_sde=$(grep -c "OK DELETE" "${logfile}")
+	ftp_fde=$(grep -c "FAIL DELETE" "${logfile}")
+fi
```
ssm: I think "0" is a better value for known events which did not happen.

To use less resources (and initialize everything to "0", you could also call "awk | sort" once, instead of many instances of "grep".  Example:

    awk '
    BEGIN {
    	count["ftp_c"] = 0
    	count["ftp_sl"] = 0
    	count["ftp_fl"] = 0
    	count["ftp_su"] = 0
    	count["ftp_fu"] = 0
    	count["ftp_sd"] = 0
    	count["ftp_fd"] = 0
    	count["ftp_sde"] = 0
    	count["ftp_fde"] = 0
    }
    /CONNECT/       { count["ftp_c"]++   }
    /OK LOGIN/      { count["ftp_sl"]++  }
    /FAIL LOGIN/    { count["ftp_fl"]++  }
    /OK UPLOAD/     { count["ftp_su"]++  }
    /FAIL UPLOAD/   { count["ftp_fu"]++  }
    /OK DOWNLOAD/   { count["ftp_sd"]++  }
    /FAIL DOWNLOAD/ { count["ftp_fd"]++  }
    /OK DELETE/     { count["ftp_sde"]++ }
    /FAIL DELETE/   { count["ftp_fde"]++ }
    
    END {
    	for (sort label in count)
    		printf "%s.label %d\n", label, count[label]
    }
    ' $logfile | sort

driskell: Does grep -c not return "0"?

That awkward is definitely better I'll throw that in. I'd initially just modified existing and kept roughly the same method - but you are right to take this further while someone's attention is on it!

ssm: Jason Woods <notifications@github.com> writes:

> Does grep -c not return "0"?

Well spotted.

echo foo | grep -c bar
0

-- 
Stig Sandbeck Mathisen

driskell:
Switched to awk :+1:

driskell:
Also added rename counts.

PR #602, opened 2015-03-11
General purpose backup monitoring plugin

AronNovak:
If you have a directory where daily, weekly, monthly you generate
backups of files, SQL databases, etc, you might want to be informed
if the backup generation process is gone or the behavior changes
greatly. This monitoring plugin watches the number of young files
in a certain directory. Relies on find command to do the job.


steveschnepp:
Merging anyway, since : 
1. It is contrib
2. The plugin itself is interesting and working.

PR #607, opened 2015-04-15
Add graph_scale no to php_acp_ for Percents graph

leeclemens:


leeclemens:
I've been running this since before submitted, seems to work fine :)

PR #614, opened 2015-05-04
fix - remove error message, fetch correct count source

mittyorz:
* redirect error message of `ls` to /dev/null
  * ```ls: cannot access /sys/devices/virtual/block/*/md/mismatch_cnt: No such file or directory```
* fetch mismatch_cnt to get correct count

* https://www.kernel.org/doc/Documentation/md.txt
  * sync_completed: the number of sectors that have been completed
  * mismatch_cnt: the number of errors that are found

### DIFF
--- a/plugins/disk/raid-mismatch-count
+++ b/plugins/disk/raid-mismatch-count
@@ -24,7 +24,7 @@
 # #%# capabilities=autoconf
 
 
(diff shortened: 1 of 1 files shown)


mittyorz:
# before

* munin-run
```
md0.value 3952256 / 7809024
md1.value delayed
```

* graph became like blow
before

# after

* munin-run
```
md0.value 0
md1.value 0
```
raid_mismatch_count-day

PR #615, opened 2015-05-15
Re-submit zfsonlinux_stats_

elfurbe:
Re-submitting the zfsonlinux_stats_ plugin as it got butchered by my shoddy commits when merged. This looks like a duplicate of #562 but that merge was actually undone by #563. I'm still bad at git, years in.

I'll include the original description here for clarity:
I wanted to get all these ZFS stats on my ZFSonLinux install, so I modified the FreeBSD plugin to read from the stats files generated by the SPL module instead. I also added some checks to make sure I wasn't dividing by zero in a number of places. From an output perspective, this produces the same values as the zfs_stats plugin.

Thanks!

### DIFF
--- a/plugins/zfs/zfsonlinux_stats_
+++ b/plugins/zfs/zfsonlinux_stats_
@@ -62,7 +62,7 @@ do
             export $VARNAME=$data
             ;;
     esac
-done < /proc/spl/kstat/zfs/arcstats
+done < $ARCSTATS
 
 #
 # Pull all values from zfetchstats
@@ -84,7 +84,7 @@ do
             export $VARNAME=$data
             ;;
     esac
-done < /proc/spl/kstat/zfs/zfetchstats
+done < $ZFETCHSTATS
 
 #
 # Calculation macros
--- a/plugins/zfs/zfsonlinux_stats_
+++ b/plugins/zfs/zfsonlinux_stats_
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 # ZFS statistics for ZFSonLinux
 # Author: Adam Michel (elfurbe@furbism.com)
 #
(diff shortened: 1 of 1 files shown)


plugins/zfs/zfsonlinux_stats_
```diff
@@ -0,0 +1,330 @@
...
+#
+#%# family=auto
+
+FUNCTION=$(basename $0 | cut -d_ -f3)
+MEMMAX=`cat /proc/meminfo | grep MemTotal | awk '{print $2}'`
+BC='/usr/bin/bc -q'
+ARCSTATS="/proc/spl/kstat/zfs/arcstats"
+ZFETCHSTATS="/proc/spl/kstat/zfs/zfetchstats"
+
+ARC_HITS=`cat $ARCSTATS | grep ^hits | awk '{print $3}'`
```
ssm: There are some inefficiencies here, which should be addressed.

First, the pattern "cat $ARCSTATS | grep ^bar | awk {print $2}" can be done as a single "awk '$1 == "bar" {print $3}' $ARCSTATS" per line.

The plugin would also improve if you find more ways to avoid forking. Perhaps the arcstats should be parsed in its entirety with shell (cat $ARCSTATS | while read name something value; do ... ; done), alternatively awk, or perl, without calling out to external utilities?

Thanks for the pull request.  A good ZFS plugin would be welcome.

elfurbe:
That was a pretty ugly section. This is cleaner from a forking perspective, at least. Still a little hacky, but at least it only takes two file reads now.

elfurbe:
And I forgot to switch back to my declared variables for the stat file locations. I think this is actually done now.

ssm:
Looks much better now :)

https://travis-ci.org/munin-monitoring/contrib/builds/62782028#L1077 notes that there is "bash" specific language in the plugin, but that the interpreter is /bin/sh. (it also says which lines should be changed, and to what, to work on all systems)

elfurbe:
Yeah, it thinks [[ and [ are equivalent which is not the case. I'm just going to switch the interpreter to bash, this is a ZoL-specific plugin and I don't think it's unreasonable to expect bash on the systems that officially supports.

PR #619, opened 2015-05-23
Add modem-nvg510

smith153:
Tracks most stats reported from NVG510 DSL modems

### DIFF
--- a/plugins/network/modem-nvg510
+++ b/plugins/network/modem-nvg510
@@ -1,0 +1,178 @@
+#! /usr/bin/env perl
+
+
+=head1 NAME
+
+modem-nvg510 - Plugin to monitor Motorola/Arris NVG510 DSL modem stats
+
+=head1 CONFIGURATION
+
+	[modem-nvg510]
+	env.url http://192.168.1.254/cgi-bin/dslstatistics.ha
+
+=head1 AUTHOR
+
+Samuel Smith <esaym (snail) cpan.org>
+
+=cut
+
+use strict;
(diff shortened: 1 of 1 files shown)


plugins/network/modem-nvg510
```diff
@@ -0,0 +1,178 @@
...
+	lof_down => 24,
+	lof_up => 25,
+	fec_down => 26,
+	fec_up => 27,
+	crc_down => 28,
+	crc_up => 29,
+};
+
+if(defined $ARGV[0] and $ARGV[0] eq 'autoconf'){
+	print "yes\n";
```
ssm: It should not print "yes" for autoconf unless it has verified it is able to connect to the web server in question,

plugins/network/modem-nvg510
```diff
@@ -0,0 +1,178 @@
...
+	los_up => 23,
+	lof_down => 24,
+	lof_up => 25,
+	fec_down => 26,
+	fec_up => 27,
+	crc_down => 28,
+	crc_up => 29,
+};
+
+if(defined $ARGV[0] and $ARGV[0] eq 'autoconf'){
```
ssm: It should have a magic marker (#%# family=auto) in the comments somewhere for munin-node-configure to call it with "autoconf"

See example at https://github.com/munin-monitoring/munin/blob/devel/plugins/node.d/apache#L99

smith153:
Added correct autoconf support. Thanks for the tip.

PR #625, opened 2015-06-01
Adding 1sec system plugins C for linux

steveschnepp:
They are written in C, as low overhead is the main design point.

### DIFF
--- a/plugins/system/1sec/if1sec-c.c
+++ b/plugins/system/1sec/if1sec-c.c
@@ -1,0 +1,273 @@
+/*
+ * if1sec C plugin
+ */
+#include <stdlib.h>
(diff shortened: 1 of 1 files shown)


Skaronator:
Any info's how to compile it properly?

steveschnepp:
`make ifsec` in the directory should work.

Skaronator:
Hi, thanks for the reply. I've already tried that but I'm getting **Segmentation fault** error when trying to run the plugin via munin-run.

Enabled the plugin with:
```
[if1sec-c]
user root
```
Restarted Munin
```
# munin-run if1sec-c
```
Return **Segmentation fault**

PR #626, opened 2015-06-08
[monit] Not using memory is a warning condition

szepeviktor:
When a daemon fails `-nan` is its memory usage.


szepeviktor:
Thank you.

PR #631, opened 2015-07-12
rewrite memcached family plugins for unix socket usage

western:
rewrite memcached family plugins: bytes_, connections_, hits_, items_, requests_, traffic_
for unix socket usage


ssm:
Thank you.  Having socket support is very nice :)

PR #641, opened 2015-08-24
Keepconf plugins

rfrail3:
Adding keepconf munin plugins
keepconf_time-day
keepconf_status-day

### DIFF
--- a/plugins/keepconf/keepconf_status
+++ b/plugins/keepconf/keepconf_status
@@ -22,7 +22,7 @@ env.files /var/tmp/keepconf-mon.csv
 Ricardo F. <rfrail3@yahoo.es>
 
 =head1 VERSION
-v.1.1.1
+v.1.2.0
 
 =head1 LICENSE
 
@@ -37,13 +37,20 @@ GPLv2
 
 . $MUNIN_LIBDIR/plugins/plugin.sh
 
+LIST=$files
 
 if [ "$1" = "autoconf" ]; then
+
+	for j in ${LIST}; do
+		if [ ! -f ${j} ]; then
+			echo no
+			exit 1
+		fi
+	done
+
 	echo yes
 	exit 0
 fi
-
-LIST=$files
 
 if [ "$1" = "config" ]; then
 
--- a/plugins/keepconf/keepconf_status
+++ b/plugins/keepconf/keepconf_status
@@ -1,0 +1,84 @@
+#!/bin/sh
+# -*- sh -*-
+
+: << =cut
+
+=head1 NAME
+
+keepconf_report -Indicate the status of keepconf execution
+
+=head1 APPLICABLE SYSTEMS
+Linux box with keepconf installed.
+
+=head1 CONFIGURATION
+
+[keepconf_*]
+env.files /var/tmp/keepconf-mon.csv
+# Multifile its available
+#env.files /var/tmp/keepconf-client1.csv /var/tmp/keepconf-client2.csv
+
+=head1 AUTHOR
+
+Ricardo F. <rfrail3@yahoo.es>
+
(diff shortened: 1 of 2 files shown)


plugins/keepconf/keepconf_status
```diff
@@ -0,0 +1,84 @@
...
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+
+. $MUNIN_LIBDIR/plugins/plugin.sh
+
+
+if [ "$1" = "autoconf" ]; then
+	echo yes
```
ssm: "autoconf" must not answer "yes" unconditionally. It must check for the presence of what it is graphing.

plugins/keepconf/keepconf_status
```diff
@@ -0,0 +1,91 @@
...
+
+. $MUNIN_LIBDIR/plugins/plugin.sh
+
+LIST=$files
+
+if [ "$1" = "autoconf" ]; then
+
+	for j in ${LIST}; do
+		if [ ! -f ${j} ]; then
+			echo no
```
sumpfralle: maybe you want to add the reason for "no" here - e.g. "no (file '$j' is missing)"

plugins/keepconf/keepconf_status
```diff
@@ -0,0 +1,91 @@
...
+	exit 0
+fi
+
+if [ "$1" = "config" ]; then
+
+	echo 'graph_title Keepconf report'
+	echo 'graph_args --base 1000 -l 0 '
+	echo 'graph_vlabel Hosts'
+	echo 'graph_scale no'
+	echo 'graph_category keepconf'
```
sumpfralle: please pick one of the [commonly used categories](http://munin-monitoring.org/wiki/graph_category_list), if one of them matches somehow

plugins/keepconf/keepconf_status
```diff
@@ -0,0 +1,91 @@
...
+	echo 'graph_info Indicate the status of a keepconf execution.'
+
+	for i in ${LIST}; do
+
+                HNAME=`echo ${i} | md5sum | cut -d - -f1 | sed 's/ //g'`
```
sumpfralle: please check you usage of spaces/tabs (not just in this line)

plugins/keepconf/keepconf_status
```diff
@@ -0,0 +1,91 @@
...
+
+		NAME=`cat ${i}  | grep ^\"CFG-FILE\" | awk -F, '{print $2}' | tr -d '"'`
```
sumpfralle: what do you think about this?
`NAME=$(grep '^"CFG-FILE"' "$i" | awk -F, '{print $2}' | tr -d '"')`
I guess, it could be slightly more readable.

Or maybe even use a shell function:
`get_config_value() { grep "^\"$2\"" "$1" | awk -F, '{print $2}' | tr -d '"'); }`
It could be called like this:
`get_config_value "$i" "CFG-FILE"`

plugins/keepconf/keepconf_time
```diff
@@ -0,0 +1,99 @@
...
+=head1 LICENSE
+
+GPLv2
+
+=head1 VERSION
+v.1.2.0
+
+=head1 MAGICK MARKERS
+
+ #%# family=auto
```
sumpfralle: Since both plugins share most of their code, it could be a good approach to combine both.
In this case you would add the "suggest" capability returning "status" and "time".

sumpfralle:
@rfrail3: what do you think?

rfrail3:
Hello sumpfralle, Due the time passed, I prefer to close the pull-request, it will have a different approach in a few time. Thanks

PR #644, opened 2015-09-01
add postfix-policyd-spf-python plugin

lynix:
Very simple but useful plugin to monitor postfix-policyd-spf classification results. Written by me some time ago, but still works :)

### DIFF
--- a/plugins/mail/policyd-spf-python
+++ b/plugins/mail/policyd-spf-python
@@ -1,0 +1,86 @@
+#! /bin/bash
+#
+#   Munin plugin to monitor postfix-policyd-spf-python results
+#   Contributed by Alexander Koch <lynix47@gmail.com> 
+# 
+# Parameters understood:
+#	config		(required)
+#	autoconf 	(optional - used by munin-config)
+# 
+# Config variables:
+#       logfile      - Where to find the postfix log (mail.log)
+#
+# Add the following line to a file in /etc/munin/plugin-conf.d:
+# 	env.logfile /var/log/your/mail.log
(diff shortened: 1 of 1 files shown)


ssm:
Could you pick a license for it, please, and add it to the plugin?

Either a copyleft or permissive opensource license would be fine, but it should be explicitly stated which license to use in the plugin.

lynix:
Sure, I'd like to go with the MIT license (although I don't care much, so if you've got a better suggestion I'm open to it). Shall I paste the entire license in the script or just state the name of the license?

I'm a bit irritated because I've seen many plugins without any license information (partially even without author information).

ssm:
Alexander Koch <notifications@github.com> writes:

> Sure, I'd like to go with the MIT license (although I don't care much,
> so if you've got a better suggestion I'm open to it). Shall I paste
> the entire license in the script or just state the name of the
> license?

For most plugins in the "munin" repo, we just state the name of the
license in the plugin.  (MIT/Expat, MIT/X11, GPL-v12345, Apache-2.0,
etc…)

Some plugins have the entire license text.  I'm fine with both
alternatives.

> I'm a bit irritated

Sorry about that. :/

> because I've seen many plugins without any license information
> (partially even without author information).

Yes, there are many of those, unfortunately.  Tracking down previous
authors, in case that becomes necessary, will take a lot of time.  The
alternative may be to drop them altogether.

-- 
Stig Sandbeck Mathisen

lynix:
Thanks for the info :)  I've added the license commit.

PR #645, opened 2015-09-02
Plugin monitoring CPU on Cisco SB Devices

jmdevince:
Monitors CPU % Utilization on Cisco Small Business Devices (Does not use the typical Cisco iOS OIDs)


ssm:
thanks

PR #648, opened 2015-09-20
plugins/virtualization/vserver/vserver_jiffies: several enhancements (but zsh)

akorn:
* Switch to zsh (used to be bash)
 * Fix to remove dots from vserver names (replace them with underscores; dots confuse munin)
 * Drop support for old (pre 2.6.19) kernels
 * Replace cat | grep | cut pipelines with a single sed each
 * Add env.stripdomain (a domain name to strip from the end of vserver names when generating labels; be sure to include the leading dot)

### DIFF
--- a/plugins/virtualization/vserver/vserver_jiffies
+++ b/plugins/virtualization/vserver/vserver_jiffies
@@ -1,6 +1,18 @@
-#!/bin/bash
+#!/bin/zsh
 #
 # Created by Jan Rękorajski <baggins@pld-linux.org> based on vserver_cpu_ plugin.
+#
+# Obtained from https://github.com/munin-monitoring/contrib.git
+#
+# Changes by Andras Korn, 2015:
+#
+#  * Convert to zsh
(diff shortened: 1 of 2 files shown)


ssm:
Since this is the first zsh plugin, the test suite also needs updating to test it before we can merge this.

akorn:
I _might_ be willing to do that because I may write additional plugins in zsh later; do you have some pointers? I don't even know how where to begin.

ssm:
Two pointers:

1: figure out how to syntax check a zsh script

2: add an "if interpreter is zsh" section in https://github.com/munin-monitoring/contrib/blob/master/t/test.t (copy from one of the other sections with just one test)

akorn:
OK, thanks; in case anyone else wants to do it, "zsh -n" works for syntax checking.
I certainly won't get around to this today, alas.

akorn:
Looks like the CI test is failing because the CI system doesn't have `zsh` installed.

PR #649, opened 2015-09-22
Updated to v0.2

stimpy23:
Updated to match the outputs of current ice- and shoutcast versions...


kenyon:
Next time, make whitespace changes in separate commits, so we can see what actually changed. Since you're the original author, I assume you know what you're doing here.

stimpy23:
I'll keep that in mind. Thank you!

PR #650, opened 2015-09-24
pf_tables : Munin plugin to monitor pf tables.

lduchosal:
Inout: bandwidth usage for table
Addresses: number of entries in table


kenyon:
You might want to specify the license for arp_bsd_.

PR #651, opened 2015-10-05
p: fixing openweather_: replaced fgrep, tr, cut with expr

obma:
Use expr to extract values and names from the OpenWeather API xml instead of the, when having a city with spaces in its name not correctly working fgrep | tr | cut construct.

### DIFF
--- a/plugins/weather/openweather_
+++ b/plugins/weather/openweather_
@@ -19,6 +19,7 @@
 # env.apikey XYZ
 
 query_string=$(printf '%s' "${0#*_}" | tr '_' '=')
+plugin_name=$( basename $0 )
 
 OWAPI=$( curl -s "http://api.openweathermap.org/data/2.5/weather?mode=xml&${query_string}")
 # API returns temp in K, we have to convert it in C
@@ -30,7 +31,7 @@ CITY=$( expr "$OWAPI" : '.*\<city.*name=\"\(.*\)\"><coord.*' )
 if [ "$1" = "config" ];
 then
 	cat <<- EOF
-		multigraph $0
+		multigraph $plugin_name
 		graph_title Temperature in ${CITY}
 		graph_vlabel Celsius
 		graph_category weather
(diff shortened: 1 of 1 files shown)


plugins/docker/docker_cpu
```diff
@@ -0,0 +1,116 @@
...
+      print "yes\n";
+      exit 0;
+   }
+   else{
+      print "no (Docker has not been found)\n";
+      exit 0;
+   }
+}
+
+$docker =~ s/\s+$//;
```
sumpfralle: sorry, I am not fluent in perl: what is the purpose of this line?

plugins/docker/docker_cpu
```diff
@@ -0,0 +1,116 @@
...
+
+=head1 APPLICABLE SYSTEMS
+
+Should work on any Linux system that has docker support.
+
+=head1 CONFIGURATION
+
+Root privilege required to execute docker command.
+
+1. Create a new file named "docker" inside the folder /etc/munin/plugin-conf.d/
```
sumpfralle: This configuration file location is distribution specific.
It should be sufficient to just mention the required configuration.

plugins/docker/docker_cpu
```diff
@@ -0,0 +1,116 @@
...
+         $graphlimit = $$_{'ncpu'};
+      }
+   } 
+   $graphlimit = $graphlimit * 100;
+   print "graph_title Docker container CPU usage\n";
+   print "graph_args --base 1000 -r --lower-limit 0 --upper-limit $graphlimit\n";
+   print "graph_vlabel %\n";
+   print "graph_scale no\n";
+   print "graph_period second\n";
+   print "graph_category Docker\n";
```
sumpfralle: please use the category "virtualization" instead (or pick something else from the list of [recommended categories](http://munin-monitoring.org/wiki/graph_category_list))

sumpfralle:
Sorry - I misunderstood your pull request (I thought it is a new plugin). Please rebase it.

sumpfralle:
@obma: what do you think?

obma:
Is it okay now? Or do I have to do anything - sorry, I'm not using git too often and am still quite a novice.

sumpfralle:
Yes, technically (for git) it is fine now.

I like that you managed to get rid of the temporary file. This is much better now.

PR #656, opened 2015-10-17
modify comment

bassbone:
/ur/ -> /usr/

### DIFF
--- a/plugins/system/cpu-usage-by-process
+++ b/plugins/system/cpu-usage-by-process
@@ -65,7 +65,7 @@ fi
 
 for proc in $procs ; do {
 
-    ps axo 'pcpu,comm' | grep "$proc" |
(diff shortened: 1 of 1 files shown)


ssm:
I'm not sure about the change to "top", since it is often spectacularly non-portable between systems and versions, while "ps" is easier.

bassbone:
I think "ps" is not the instantaneous CPU utilization, so I want to change to "top".

ssm:
They give the same data, except that "top" emits the "args" field, where the ps output specifies "comm".

Compare:

    diff -u \
      <(top -b -n 1 | awk '{print $9, $12}'|sort) \
      <(ps axo 'pcpu,args' | awk '{print $1, $2}'|sort)

Given that "ps" is more portable across operating system, I prefer that.

bassbone:
FYI
http://www.linuxquestions.org/questions/linux-server-73/top-and-ps-940393/

PR #658, opened 2015-10-22
Add Docker plugins

scanterog:
Hi,

I've written this plugin to monitor the CPU usage and memory usage on docker containers. Currently, we are using it on a production environment at Sugar Labs infra. I want to contribute this to munin.

Regards,


plugins/docker/docker_memory
```diff
@@ -0,0 +1,81 @@
...
+1. Create a new file named "docker" inside the folder /etc/munin/plugin-conf.d/
+2. Docker file content:
+
+[docker_memory]
+user root
+
+=head1 MAGIC MARKERS
+
+  #%# family=auto
+  #%# capabilities=autoconf
```
ssm: This marker tells munin-plugin-configure that the plugin can be run with "autoconf" command line argument.

Please see http://munin-monitoring.org/wiki/PluginConcise#autoconf
.

scanterog: Thanks Stig. I've recently added the autoconf support. If there is anything else to fix just let me know. Regards.

plugins/docker/docker_cpu
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/perl -w
```
ssm: Instead of using commandline flags to enable warnings, please do:

    use strict;
    use warnings;

PR #659, opened 2015-10-26
shebang: using env for portability

lr1980:
used this one-liner:

for i in `egrep -Ri '^#! */' . | grep -v '/\.' | grep -v '/usr/bin/env'
| cut -d: -f1 | uniq`; do perl -pi -e 's/^#!
*.*\/(.*)$/#!\/usr\/bin\/env \1/g' $i; done


kenyon:
I'm not sure this really helps with portability. I thought the package build scripts rewrite the shebangs anyway.

lr1980:
example:
https://github.com/munin-monitoring/contrib/blob/master/plugins/zfs/zfs-filesystem-graph
````
#!/usr/local/bin/bash
```
-> not work with linux, but "env bash"

kenyon:
Ah, I didn't realize this was the contrib repository. This massive change is even less necessary in the contrib repository. These plugins are intended to be carefully vetted and possibly customized by the user.

PR #664, opened 2015-10-29
new network plugin added: olsrd

sumpfralle:
Collect basic information about the neighbours of an OLSR node:
 * link quality
 * neighbour link quality
 * number of nodes reachable behind each neighbour
 * ping times of direct neighbours

OLSR is a routing protocol (layer 3).

### DIFF
--- a/plugins/network/olsrd
+++ b/plugins/network/olsrd
@@ -61,6 +61,10 @@ else
     python "$0" "$@"
 fi
 exit $?
+
+# For shell: ignore everything starting from here until the last line of this file.
+# This is necessary for syntax checkers that try to complain about invalid shell syntax below.
+true <<EOF
 """
 
 
@@ -338,3 +342,7 @@ if __name__ == "__main__":
             remote = get_clean_fieldname(link["remote"])
             print("multigraph olsr_neighbour_ping.host_{remote}".format(remote=remote))
             print("neighbour_{remote}.value {value:.4f}".format(remote=remote, value=ping_time))
+
+# final marker for shell / python hybrid script (see "Interpreter Selection")
+EOF = True
+EOF
--- a/plugins/network/olsrd
+++ b/plugins/network/olsrd
@@ -1,0 +1,340 @@
+#!/bin/sh
+# weird shebang? See below: "interpreter selection"
+#
+# Collect basic information about the neighbours of an OLSR node:
+#   * link quality
+#   * neighbour link quality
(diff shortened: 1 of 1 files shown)


plugins/network/olsrd
```diff
@@ -0,0 +1,348 @@
+#!/bin/sh
+# weird shebang? See below: "interpreter selection"
```
ssm: Not entirely sure about this.  It would probably be better to pick one interpreter, and document how to use it with others.

sumpfralle: I understand your concerns - I was also unsure about this detail.

Usually I would not mind to just pick python3 - as it is well distributed and the most common python interpreter.
There are two reasons for the approach that I proposed:
1. The most common environment for an olsr daemon is an embedded router. These devices usually work with [http://openwrt.org](openwrt) running on hardware with 4 or 8 MB storage. Most people will not install python (minimal version >3 MB) but micropython instead (~200kB) on their routers. Thus I am confident, that the most likely use-case of this plugin runs with micropython only.
2. This specific plugin (in combination with micropython) sadly requires to be run with a non-default parameter for micropython due to its memory usage (>128kB heap size). To make it all really complicated: this parameter cannot be added to the shebang since these are not effective with openwrt's default shell (ash). Thus the current hybrid shell/python execution is the only way to run this plugin in openwrt (as far as I know).

If the above sounds foolishly complicated, I would kindly ask you to read my reasoning above one more time. I have a fairly good understanding of the openwrt environment and I tried very hard to avoid the discussed hybrid shell/python design. But I failed to find a better solution.
I welcome any suggestions for improvements or alternatives.

I see the following three approaches for this issue:
1. Keep the current shell/python hybrid code.
2. Remove the shell portions and set the shebang to pyhon3. Add documentation for running the plugin on openwrt (this will be a bit complicated).
3. Implement the plugin in a different language. This failed for me with shell since it required way too many forks/execs (runtime: ca. 60s on modest router hardware). Lua could be an alternative since it is quite common on openwrt, but I am not fluent in this language.

Since I want to use this plugin on openwrt (for a freifunk community firmware), I would be keen to just source it from this repository without changes. Thus I would prefer option (1).

Option (2) would require me to implement a slightly more complicated process for tracking this upstream source.

Option (3) needs an implementor.

I am fine with all options.
What do you think?

steveschnepp:
Let's integrate as is. Something not perfect is better than nothing at all.

PR #665, opened 2015-10-29
new network plugin added: ath9k

sumpfralle:
Collect information related to ath9k wireless events and states.
 * rate control statistics ("rc_stats")
 * events (dropped, transmitted, beacon loss, ...)
 * traffic (packets, bytes)

All data is collected for each separate station (in case of multiple
connected peers). Combined graphs are provided as a summary.


sumpfralle:
please ignore - there is already an updated pull request

PR #666, opened 2015-10-30
Icecast stats

sumpfralle:
This icecast plugin captures a set of data similar to the "icecast2_" and "icecast2_all" plugin. Sadly both are not configurable (manual changes are required for target host and access credentials). Additionally they use the rather old-fashioned (and password restricted) xml status output.

This new plugin "icecast2_stats" (written from scratch) contains the following notable differences:

* runs with Python 3
* supports a real "autoconf" and "suggest" interface
* is configurable (target host and port) via environment settings
* uses the json status output instead of the xml data
* contains no hard-coded stream filename patterns

### DIFF
--- a/plugins/icecast/icecast2_stats_
+++ b/plugins/icecast/icecast2_stats_
@@ -1,0 +1,182 @@
+#!/usr/bin/python3
+#
+# This plugin shows the statistics of every source currently connected to the Icecast2 server.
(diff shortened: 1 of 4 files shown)


sumpfralle:
sorry for the noise - some extra commits slipped into this pull request

please ignore

PR #669, opened 2015-10-30
tests: test compiling with python3 if given as shebang interpreter

sumpfralle:
Currently plugins fail to compile if they use Python 3 syntax features,
even though they use a correct shebang.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: perl
 install:
   - sudo apt-get update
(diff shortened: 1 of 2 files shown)


ssm:
Looks like python3 is missing from the travis environment, too.

Adding it to  the .travis.yml file would do the trick

sumpfralle:
@ssm: thank you for this hint!
I was already searching for the proper place to add this requirement ...

Issue #660, opened 2015-10-27
Not working with Varnish 4.1


michbsd:
Unfortunately this plugin does not work with varnish4 - as it seems varnish4.1 has introduced new file permissions and a new user context

 ❯ munin-run  varnish4_memory_usage
Can't open log - retrying for 5 seconds

valurhrafn:
If you add the group varnish to the munin plugin conf you get some stats and graphs.
```
[varnish4_*]
group varnish
env.varnishstat varnishstat
```
The graphs for Memory usage and Misbehavior are not being generated though.
I tried debugging, and removing items that were not in the xml anymore, but still no graphs.

ssm:
Varnish 4.1 requires anyone accessing the varnish shared memory to be a member of the "varnish" group.

This should work fine with older varnish installs as well so I think it would be a sensible addition to the plugin documentation.

valurhrafn:
The problem with Memory usage and Misbehaviour seems to be because they are using a counter, the type that gets reported back to munin for those fields are g and c instead of COUNTER, DERIVE or GAUGE

valurhrafn:
I have got all the statistics working for Varnish 4.1 on my installation. I am going to clean up the code changes I made and then issue a pull request.

michbsd:
Awesome! thanks

michbsd:
fixed, thanks!

PR #670, opened 2015-11-09
Varnish 4.1 update

valurhrafn:
Updated README.md with the config changes needed for 4.1
Updated varnish4_ so it is displays graphs that were not generated in 4.1 because the type was unknown to munin. 
This will resolve issue #660


mmh:
I can confirm that this works with varnish 4.1.
Thanks :+1:

gnuheidix:
Works great! Thanks :)

PR #671, opened 2015-11-10
proc_mem_by_user:

drzraf:
- use of bash
- don't corrupt rrd files when "root" (or any unclean value) is requested

### DIFF
--- a/plugins/processes/proc_mem_by_user
+++ b/plugins/processes/proc_mem_by_user
@@ -1,5 +1,6 @@
 #!/bin/sh
 #
+# (c) 2015, Raphaël Droz <raphael.droz@gmail.com>
 # (c) 2014, Gilles Fauvie <gfauvie@opendbteam.com>
 # Based on the 'du_multidirs' plugin, written by Christian Kujau <lists@nerdbynature.de>
 #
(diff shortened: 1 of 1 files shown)


drzraf:
ping

drzraf:
ping

drzraf:
ping

sumpfralle:
Hi drzraf,

thank you for fixing the "root" issue!

Regarding your proposed change from "sh" to "bash":
The munin plugins are targeting a wide range of platforms including systems that do not contain bash by default (e.g. openwrt, OpenBSD, AIX). Since your change does not require specific bash features, the loss of support for different shell variants does not seem to be compensated by any significant benefit.
Thus please revert this part of your change.

Thank you!
Lars

drzraf:
posix SHell does not support:
* here document
* numeric tests expression ``(( ))``
* multiple fd redirection ``&>``

sumpfralle:
Yes, indeed [ticket #921](http://munin-monitoring.org/ticket/921) should really be fixed.

Regarding sh/bash: sorry for the confusion, I was referring the generation of sh-like interpreters that is similar to the set of features offered by dash. This should work well with all `/bin/sh` interpreters provided by modern operating systems.
`shellcheck -s dash foo` should help you with syntax checks.

drzraf:
Rerolled to comply with "dash-feature-level".
I realized that (at least in recent release), /usr/share/munin/plugins/plugin.sh 's ``clean_fieldname`` provides the root => __root substitution.
About [root issue #921](http://munin-monitoring.org/ticket/921]):
* does anyone still care about Trac?
* is the bug fixed in 3.0?

sumpfralle:
Regarding the `clean_fieldname` munging of "root": good to know!
Trac: I assume it was superseded by github's ticket system.

PR #673, opened 2015-11-10
monitor some values of the PrestaShop web application

drzraf:
(note: Debian-only since it uses /etc/mysql/debian.cnf)

### DIFF
--- a/plugins/prestashop/prestashop_
+++ b/plugins/prestashop/prestashop_
@@ -1,0 +1,109 @@
+#!/bin/bash
+
+: << =cut
+=head1 NAME
+
+Munin Plugin to grab PrestaShop statistics
+
+=head1 CONFIGURATION
+
+# Sample:
+[prestashop]
+	user root
+	env.database prestashop
+	env.dbprefix ps_
+
+=head1 AUTHOR
+
+# (c) 2015 Raphaël Droz <raphael.droz+floss@gmail.com>
+# General Public Licence v3 or later
+
+=head1 LICENSE
+
+GPLv3
+
+=head1 MAGIC MARKERS
+
+#%# family=auto
+#%# capabilities=autoconf suggest
+
+=cut
+
+graph_type=${0##*_}
+
+
+if [[ $1 = autoconf ]]; then
+    let ok=1
+    [[ -z $database ]] && echo "no (env.database not defined)" && ok=0
+    [[ -z $dbprefix ]] && echo "no (env.dbprefix not defined)" && ok=0
+    [[ -z $graph_type ]] && echo "no plugins linked with a bad name (unknown graph type '$graph_type')" && ok=0
+    [[ ! -r /etc/mysql/debian.cnf ]] && echo "no /etc/mysql/debian.cnf isn't readable" && ok=0
+    ! type -P mysql &>/dev/null && echo "no (one of mysql not found)" && ok=0
+    
+    (( $ok == 1 )) && echo yes
+    exit 0
+fi
+
+if [[ $1 = suggest ]]; then
+    for i in visits products orders ; do
+	echo $i
+    done
+    exit 0
+fi
+
+if [[ $1 = config ]]; then
+    if [[ $graph_type  == visits ]]; then
+	cat <<EOF
+graph_title PrestaShop connections
+graph_args --base 1000 -l 0
+graph_category php
+graph_info PrestaShop connections
+graph_period minute
+connection.label connections
+connection.info Connections
+connection.min 0
+connection.type DERIVE
+guest.label guests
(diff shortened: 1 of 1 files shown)


drzraf:
ping?

ssm:
Shellcheck has a few issues:

    In prestashop_ line 37:
        [[ -z $database ]] && echo "no (env.database not defined)" && ok=0
              ^-- SC2154: database is referenced but not assigned.
    
    
    In prestashop_ line 38:
        [[ -z $dbprefix ]] && echo "no (env.dbprefix not defined)" && ok=0
              ^-- SC2154: dbprefix is referenced but not assigned.
    
    
    In prestashop_ line 43:
        (( $ok == 1 )) && echo yes
           ^-- SC2004: $/${} is unnecessary on arithmetic variables.

drzraf:
Neither of this checks seem to ask for code change:
> SC2154: dbprefix is referenced but not assigned.  

Indeed, I don't see any sane default value for these variable and the values come for environment. So the lacks of assignation is expected.

> SC2004: $/${} is unnecessary on arithmetic variables.  

unnecessary but completely supported (I find it less confusing to explicitly put `$`)

IMO, what needs review is the graph-type choices.
Some may prefer one type over another and I'm still unsure about what's the best overall.

plugins/prestashop/prestashop_
```diff
@@ -0,0 +1,109 @@
...
+=cut
+
+graph_type=${0##*_}
+
+
+if [[ $1 = autoconf ]]; then
+    let ok=1
+    [[ -z $database ]] && echo "no (env.database not defined)" && ok=0
+    [[ -z $dbprefix ]] && echo "no (env.dbprefix not defined)" && ok=0
+    [[ -z $graph_type ]] && echo "no plugins linked with a bad name (unknown graph type '$graph_type')" && ok=0
```
sumpfralle: The descriptive text for "no" feels hard to read and does not seem to reflect the test causing it. Maybe just pick `no (invalid link or missing graph type)`?

plugins/prestashop/prestashop_
```diff
@@ -0,0 +1,109 @@
...
+    done
+    exit 0
+fi
+
+if [[ $1 = config ]]; then
+    if [[ $graph_type  == visits ]]; then
+	cat <<EOF
+graph_title PrestaShop connections
+graph_args --base 1000 -l 0
+graph_category php
```
sumpfralle: Please pick a category from the [list of commonly used ones](http://munin-monitoring.org/wiki/graph_category_list). For me "other" seems to be the most suitable one.

sumpfralle:
@drzraf: the plugin is already in a quite good shape. Let's finish it?

What was your comment regarding the "graph-type" about?

In case of doubt, I would also be willing to merge your new plugin as it is.

drzraf:
I have some doubts about DERIVE vs COUNTER.
Ideally I would like to rely upon DERIVE but I've myself some trouble to understand graph's data after some time have passed: numbers go down to pretty insignificant levels and it's not always to understand, for example, if that "0.15" was the number of new orders by 6h or so. COUNTER has its own issues (especially in case of burst of activity).

sumpfralle:
I think, DERIVE should be the right one here, since you cannot predict, at which values the different IDs (guest, connections) will wrap.
Probably you need to adjust `graph_period` further (also for orders)? Since your events (guests, sales, ...) are integers, you should pick a period that usually covers one or more events. "hour" is probably a good choice for most shops.
Regarding bursts: I think, you cannot avoid them, if the numbers are too low for an evenly distributed flow of events.

sumpfralle:
@drzraf: what do you think?

drzraf:
I'm pretty sure you're right, but I remember having hard time to test graphes (re)rendering when tweaking the graph_period config value. And sadly I'm too busy right now to dig into this again before some more weeks?
(But feel free to do what you think is the best for most users, I'm confident you'd do the right choice)

sumpfralle:
@drzraf: what would you advise to do with your changeset?
(I do not use this plugin - thus I cannot test it)

drzraf:
I've been using DERIVE since that time and it's quite ok.
Anyway results are confusing because of "aggregation" (we would rather want consistent results by days/weeks).
In the end it's not really "production-ready", only "better-than-nothing".

sumpfralle:
> In the end it's not really "production-ready", only "better-than-nothing".

Thank you for your honest evaluation!

I leave it up to you to decide, if it is useful enough to be part of this collection.

sumpfralle:
What do you think?

drzraf:
I'm not using it anymore so no maintenance could be expected from me. If no one takes over, then just close it.

PR #674, opened 2015-11-10
change.org API: monitor signatures count for one (or various) petition(s) you own

drzraf:


drzraf:
ping

drzraf:
ping

ssm:
Shellcheck has a few issues:

    In change_org line 45:
    	    (( $during_autoconf )) && echo "no ($i isn't a valid petition ID)"
                   ^-- SC2004: $/${} is unnecessary on arithmetic variables.
    
    
    In change_org line 51:
    	    (( $during_autoconf )) && echo "no ($i isn't given a title)"
                   ^-- SC2004: $/${} is unnecessary on arithmetic variables.
    
    
    In change_org line 63:
        [[ -z $APIkey || ! $APIkey =~ ^[a-fA-F0-9]{64}$ ]] && echo "no (env.APIkey not defined or bad format)" && ok=0
              ^-- SC2154: APIkey is referenced but not assigned.
    
    
    In change_org line 68:
        (( $ok == 1 )) && echo yes
           ^-- SC2004: $/${} is unnecessary on arithmetic variables.
    
    
    In change_org line 81:
    graph_info $change.org signature count
               ^-- SC2154: change is referenced but not assigned.
    
    
    In change_org line 86:
        while (( $i < ${#petIDs[*]} )); do
                 ^-- SC2004: $/${} is unnecessary on arithmetic variables.
    
    
    In change_org line 101:
    while (( $i < ${#petIDs[*]} )); do
             ^-- SC2004: $/${} is unnecessary on arithmetic variables.
    
    
    In change_org line 104:
        echo signcount_${id}.value $signcount
                       ^-- SC2086: Double quote to prevent globbing and word splitting.
                                   ^-- SC2086: Double quote to prevent globbing and word splitting.

drzraf:
>     In change_org line 81:
>     graph_info $change.org signature count
>                ^-- SC2154: change is referenced but not assigned.

fixed.


>     In change_org line 104:
>         echo signcount_${id}.value $signcount
>                        ^-- SC2086: Double quote to prevent globbing and word splitting.

fixed.

drzraf:
ping

sumpfralle:
@drzraf: can the title of a partition be retrieved via the API? This would simplify the usage of the second ("simple") plugin a lot.

sumpfralle:
Regarding your approach of using two different plugins for the same functionality: could you please either pick one of the two plugins or merge both interfaces into a single plugin?
(e.g. use symlink configuration and fall back to env settings or the other way around)

Two plugins instead of one would require an extra maintenance effort for this plugin repository (in the long term) that is probably not counterbalanced by the benefit of providing two different plugin interface for the same functionality.

drzraf:
If you monitor multiple petition, then "change_org" is better since:
* You only don't have the overhead of running multiple plugins
* all can be configured in one plugin-conf file and graphed together
If only one petition is to be monitored the other plugin is much more simple.

But the interesting question is "why is it complicated when it comes to multiple petition".
From reading the "complex" version, the answer clearly is:
> "Because there is no way (that I'm aware of) to pass structured (= array) data inside plugin env"
... thus bashisms / associative arrays magic are used in other to do it well.


NB: the title of a petition can be retrieved via the API. I'm just afraid about special/UTF-8 character as a label value.

plugins/change.org/change_org
```diff
@@ -0,0 +1,98 @@
...
+        pass
+    req = urllib.request.Request("https://api.change.org/v1/petitions/{}?api_key={}&fields=title".format(i, environ.get('APIkey')),
+                                      data=None, 
+                                      headers={ 'User-Agent': 'curl/7.38.0' })
+    response = urllib.request.urlopen(req).read().decode('utf-8')
+    petition_titles[i] = json.loads(response)
+    write_cache = True
+
+if environ.get('MUNIN_PLUGSTATE') and write_cache:
+    with fdopen(os.open(petition_cache_names, os.O_WRONLY | os.O_CREAT, 0o600), 'w') as outfile:
```
sumpfralle: Wouldn't `with os.open(petition_cache_names, "w", mode=0o700) as outfile` suffice?

plugins/change.org/change_org
```diff
@@ -0,0 +1,98 @@
...
+        print('''signcount_{pid}.label {title}
+signcount_{pid}.info Total signatures for {title}
+signcount_{pid}.draw AREA
+signcount_{pid}.min 0'''.format(pid=i, title=petition_titles[i]["title"]))
+    exit(0)
+
+for i in environ.get("petitions").split():
+    req = urllib.request.Request("https://api.change.org/v1/petitions/{}?api_key={}&fields=signature_count".format(i, environ.get('APIkey')),
+                                      data=None, 
+                                      headers={ 'User-Agent': 'curl/7.38.0' })
```
sumpfralle: same as above: default user agent?

plugins/change.org/change_org
```diff
@@ -0,0 +1,98 @@
...
+if environ.get('MUNIN_PLUGSTATE') and write_cache:
+    with fdopen(os.open(petition_cache_names, os.O_WRONLY | os.O_CREAT, 0o600), 'w') as outfile:
+        json.dump(petition_titles, outfile)
+    
+
+if len(argv) > 1 and argv[1] == 'config':
+    print('''graph_title change.org signature count
+graph_args --base 1000 -l 0
+graph_vlabel Signatures
+graph_category application
```
sumpfralle: Please pick a category from the [list of known categories](http://munin-monitoring.org/wiki/graph_category_list), if possible. I guess, "other" is fine.

plugins/change.org/change_org
```diff
@@ -0,0 +1,98 @@
...
+graph_info change.org signature count
+graph_period minute
+''')
+
+    for i in petition_titles:
+        print('''signcount_{pid}.label {title}
+signcount_{pid}.info Total signatures for {title}
+signcount_{pid}.draw AREA
```
sumpfralle: Maybe "LINE" is more suitable for multiple petitions? But you probably know, which looks best ...

plugins/change.org/change_org
```diff
@@ -0,0 +1,98 @@
...
+=head1 NAME
+
+Munin Plugin to grab signature count for change.org petition given its ID
+You need a valid API key for this petition
+
+=head2 CONFIGURATION
+
+# Sample:
+[change_org]
+	user nobody
```
sumpfralle: User and group selection depends on the distribution. Since "nobody/nogroup" should be the default setting, I would recommend to just remove these lines from the config example.

sumpfralle:
@drzraf: what do you think?

plugins/change.org/change_org
```diff
@@ -0,0 +1,98 @@
...
+    petition_cache_names = environ['MUNIN_PLUGSTATE'] + '/' + 'petition_names'
+    try:
+        with open(petition_cache_names, 'r') as f:
+            petition_titles = json.load(f)
+    except FileNotFoundError:
+        pass
+
+for i in environ.get('petitions').split():
+    if i in petition_titles:
+        pass
```
drzraf: thx

plugins/change.org/change_org
```diff
@@ -0,0 +1,98 @@
...
+    req = urllib.request.Request("https://api.change.org/v1/petitions/{}?api_key={}&fields=title".format(i, environ.get('APIkey')),
+                                      data=None, 
+                                      headers={ 'User-Agent': 'curl/7.38.0' })
```
drzraf: Yes, this is needed in order to avoid a 403 (at least default's python UA value triggers a 403 from change.org)

plugins/change.org/change_org
drzraf: that was about umask. but got it created in mode=0600 without fdopen()

plugins/change.org/change_org
drzraf: indeed

sumpfralle:
That looks good!

One last thing: please pick a more specific plugin name, in order to lease some room in the namespace for future plugins. Maybe "changeorg_signature_count"?

drzraf:
done

sumpfralle:
Just after a little bit more than a year :(
Thank you for your patience!

PR #677, opened 2015-11-26
fix zpool_iostat plugin

MartinNowak:
- use /bin/sh instead of /bin/bash
- add autoconf capabilities

### DIFF
--- a/plugins/zfs/zpool_iostat
+++ b/plugins/zfs/zpool_iostat
@@ -1,4 +1,7 @@
-#!/bin/bash
+#!/bin/sh
+
+#%# family=auto
+#%# capabilities=autoconf
 
 if [ "$1" = "autoconf" ]; then
                 echo yes


plugins/zfs/zpool_iostat
```diff
@@ -1,4 +1,7 @@
-#!/bin/bash
+#!/bin/sh
+
+#%# family=auto
+#%# capabilities=autoconf
```
ssm: To add "autoconf" to capabilities, the autoconf itself should be able to say both "yes" and "no", after checking for the presence of whatever's being monitored.  For now, it just says "yes", which would be wrong on any system not using zfs.

plugins/zfs/zpool_iostat
```diff
@@ -1,4 +1,7 @@
-#!/bin/bash
+#!/bin/sh
```
ssm: :+1:

plugins/zfs/zpool_iostat
ssm: I'll rewrite the required autoconf function, unless someone else beats me to it. :)

MartinNowak:
Neither using zfs nor munin any more.

PR #679, opened 2015-12-15
Remove old params

ki0:
This params in new hhvm version doesnt work any more.


ki0:
Please, could check it out as soon as you can?¿ I think its easy fix.
Thanks

PR #684, opened 2016-01-06
ceph_capacity and ceph_osd: fix autoconf

adrianimboden:
ceph_capacity and ceph_osd always recommend themselves to be installed when using
>munin-node-configure --suggest
this patch adds a test for the ceph command to avoid that.

I tested it successfully on Debian 7, 8 and Ubuntu 14.04

### DIFF
--- a/plugins/ceph/ceph_capacity
+++ b/plugins/ceph/ceph_capacity
@@ -27,6 +27,8 @@ GPLv2
 
 =cut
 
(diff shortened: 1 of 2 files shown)


adrianimboden:
Hmm, "command -v ceph" is not posix-standard. If I find another way to do it (I for myself don't need another way in my setups), I will post another pull request...

PR #685, opened 2016-01-08
improvements to some PHP/Apache plugins

drzraf:
* New plugin: apache_cache_disk_count
* apache_average_time_last_n_requests: ability to create multiple instances for distincts log files
* added slow requests for php fpm


leeclemens:
I believe changing the graph_category deserves mention, was that intentional?

drzraf:
It was although I didn't explored side-effect. I already have php / php-fpm plugins storing results below "php".
I thought that php-fpm is too much specific when visualizing graphs.
But if php-fpm was commonly accepted name as a complete autonomous category, I can change that hunk.

leeclemens:
Since this is only graphing php-fpm's status and the category was already php-fpm, I think it's appropriately specific. If there is a case for changing the category, it at least deserves mention (probably a commit, but at least a comment) that you did so. imo, the category should not be changed without a reason to do so.

drzraf:
granted: remove the change about graph category of php-fpm-process

drzraf:
ping

leeclemens:
IMHO, it doesn't make sense to graph 'slow requests' on the same axis as 'processes'. Besides them being unrelated, slow requests would probably be better as a DERIVE type, while processes should remain a GAUGE type.

ssm:
Please use one pull request per logical change, or per new plugin, that makes it much easier to renew.

drzraf:
PR #707 #708 #709
For php_fpm_process it'll come too including leeclemens's point about multigraph.

Issue #692, opened 2016-02-06
mtr100_ uses (%) in title


leeclemens:
The mtr100_ plugin hardcodes (%) as the source, for the `graph_title`. It seems to have always been this way, as mtr's output staring with HOST has always been excluded.

I believe it would be useful to display the source (although I haven't figured out a simple way to get the source IP for the route....as opposed to simply using mtr's output).

Was there a reason for this to be done this way initially? I'm going to work to update it, but wanted to raise the question as well.

leeclemens:
Fixed in #693

leeclemens:
Rewrote the command to use the same arguments as autoconf uses (also fixed unforeseen compat issue with Ubuntu).

sumpfralle:
@leeclemens: your pull request was closed some months ago without merging.
Do you still consider this bug report to be open? Would you mind preparing another pull request taking the discussion of #693 into account?

leeclemens:
@sumpfralle I was asked a question and I answered. The pull was closed before I could respond - so I'm not sure I would call it a 'discussion'.

sumpfralle:
I cannot read ssm's mind regarding the closing of the merge request, but I have to admit that I needed a good amount of guessing (even after your additional explanation in the pull request) to understand your goal. Reopening an issue is also an option, if you feel misunderstood.

My theory:
1. You misunderstood the percent sign in the title: it is supposed to differentiate this plugin from another (not yet existing) plugin visualizing absolute ping times (ms) instead of ping time distribution (%). Thus the original title is technically perfectly fine.
2. This misunderstanding lead you to the conclusion that it would be worthwhile to determine the local source IP of the traffic to the target
3. Thus you found a reasonably well working way to achieve this. But this approach is hard to understand / easy to misunderstand if your goal is not clear.

Additionally picking the wording "SAME target" caused another bit of confusion, since this very exact spelling is used by iptables for a specific SNAT target implementation (see `man iptables-extensions`). This pushed me completely off-track :)

Sincerely I did not even get your point when I asked for the state of this bug report yesterday.
Confusion everywhere - certainly no need for hard feelings.
Have fun!

leeclemens:
I'll admit I wasn't very excited for the interaction with the related PR. To be honest, I had forgotten this related Issue was still open after the PR was closed.

[...]

PR #693, opened 2016-02-06
Add source HOST to graph_title

leeclemens:
Whitespace cleanup as well, during edit to this file

### DIFF
--- a/plugins/network/mtr100_
+++ b/plugins/network/mtr100_
@@ -5,7 +5,7 @@
 #
 # Version: 1.0
 # Author: tobias.geiger@vido.info
-# Please email me bugs/suggestions 
+# Please email me bugs/suggestions
 #
 # Version: 1.1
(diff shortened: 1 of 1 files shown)


leeclemens:
Fixes #692 692

ssm:
What, exactly, is the problem you are trying to solve?

"mtr" does emit a truncated hostname after a bit, after doing a traceroute to localhost, but I'm quite sure it is not the right tool for this job.

"uname -n" emits the local node name, with or without domain according to local policy. 
Otherwise, a "hostname --fqdn" works rather well with GNU userspace, but on Solaris machines it will, or it used to, rename the server to "--fqdn".  Use with caution. :)

leeclemens:
The problem I solved was including the source in the graph's title. Running this plugin from one munin-node against multiple targets made interpreting/sorting the graphs difficult (the title is useful for this purpose).

PR #694, opened 2016-02-07
wordpress-multisite

null-git:
A Munin plugin to monitor posts, comments and pingbacks from every multisite instance.
It uses multigraphs and also shows the combined number of posts, comments, pingbacks, instances and users.

Most database requests came from the wordpress-mu-plugin which was written by Andre Darafarin and Chris Bair

### DIFF
--- a/plugins/other/wordpress-multisite
+++ b/plugins/other/wordpress-multisite
@@ -2,6 +2,7 @@
 # wordpress-multisite plugin
 # 
 # Author Jonas Palm
+# E-Mail jonaspalm . posteo. de
 # Version 0.1
 # Date 2016-02-07
 # 
@@ -20,13 +21,14 @@ The plugin need access to the wordpress database
 =head2 Config file
 Create the config file plugin-conf.d/wordpress with the following values:
 
-=over 5
+=over 6
 =item * [wordpress*]
 =item * env.DB_USER <database user (mandatory)>
 =item * env.DB_PASSWORD <database password (mandatory)>
 =item * env.DB_NAME <database name (defaults to wordpress>
 =item * env.DB_PREFIX <database prefix (defaults to wp_)>
 =item * env.DB_HOST <database host (defaults to localhost)>
+=item * env.DB_PORT <database port (defaults to 3306)>
 
 =back
 
--- a/plugins/other/wordpress-multisite
+++ b/plugins/other/wordpress-multisite
@@ -1,40 +1,51 @@
 #!/usr/bin/env sh
 # wordpress-multisite plugin
-# 
+#
+# Version 0.3
+# Date 2017-01-12
+# Some last minor fixes
+#
 # Version 0.2
 # Date 2016-10-24
 # Code improvements
-# 
+#
 # Version 0.1
 # Date 2016-02-07
 # Initial release
 #
 : <<=cut
 =head1 NAME
+
 Wordpress-Multisite Munin Plugin
 
 A Munin plugin to monitor posts, comments and pingbacks from every multisite instance.
 It uses multigraphs and also shows the combined number of posts, comments, pingbacks, instances and users.
 
-Most database requests came from the wordpress-mu-plugin which was written by Andre Darafarin and Chris Bair
+Most database queries came from the wordpress-mu-plugin which was written by Andre Darafarin and Chris Bair
 
 =head1 CONFIGURATION
+
 The plugin need access to the wordpress database
 
 =head2 Config file
+
 Create the config file plugin-conf.d/wordpress with the following values:
 
-=over 4
-=item * [wordpress*]
-=item * env.mysqlopts <ie -uroot -prootpass>
-=item * env.mysqlconnection <defaults to -hlocalhost>
-=item * env.database <ie -Dwordpress>
-=item * env.dbprefix <defaults to wp_>
+  [wordpress*]
+  env.mysqlopts		# i.e. -uroot -prootpass
+  env.mysqlconnection	# defaults to -hlocalhost
+  env.database		# i.e. -Dwordpress
+  env.dbprefix		# defaults to wp_
 
-=back
+=head1 VERSION
 
-=head1 VERSION																														   
-Version 0.2 (2016-02-07)
+Version 0.3 (2017-01-12)
+
+=head2 Some minor fixes:
+
+  * fixed perldoc
+  * fixed some syntax warnings from shellcheck
+  * replaced expr by $(( ))
 
 =head1 AUTHOR
 
@@ -43,7 +54,6 @@ Jonas Palm <jonaspalm . posteo . de>
 =head1 LICENSE
 
 GPLv3 or higher
-
 =cut
 
 # fill vars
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for this new plugin!

I have some questions and suggestions:
* please specify a license
* would it be possible to change "bash" to "sh"?
    * you would need to replace "=="  with "=" or "-eq"
    * `(( FOO += 1 ))` can be replaced with `FOO=$((FOO + 1))`
* would you mind using "$(...)" instead of the old-fashioned backticks?
* maybe you want to consider using a shell function for your repetitive "mysql ... count" statement
* do you think it could be worthwhile to change your configuration (DBHOST, ...) to the style of the core mysql plugin (env.mysqlopts, env.mysqluser, env.mysqlconnection)? This could allow password-less access (e.g. via the "debian-sys-maint" user on debian-based systems) while still allowing your style of "-H${DBHOST}" arguments.

Anyway: it is your code - everything except for the missing license is optional ...

null-git:
Thank you for your suggestions. I've added the license and reworked the code.
I couldn't test this plugin with a pure sh but I'm fairly confident that I have removed all the bashisms.
I've also changed the names of the configuration variables but the plugin still can't access the variable from the [mysql*] section because of its [wordpress_*] section name?

sumpfralle:
we are almost there :)

I have a few last nitpicks for you:
* the configuration section is not rendered for me, when I run `perldoc wordpress-multisite` - maybe just remove the "item" formatting and simply indent these lines
* the "VERSION" date is not the same as the one at the top
* please change the shebang from `/usr/bin/env sh` to `/bin/sh` - the "env" approach should only be useful for non-shell interpreters IMO
* if you want to run a syntax check of the script you can use `shellcheck -s dash SCRIPT`
** this will report a good amount of places where it would be advisable to double-quote variables (the only exceptions are probably DB_OPTIONS and DB_CONNECTION)
* extra bonus: add [capabilities](http://munin-monitoring.org/wiki/PluginConcise#Magicmarkers) (especially: autoconf)

sumpfralle:
@null-git: what do you think regarding my nitpicks?

Anyway: since the new plugin is already in a good state, I would also be willing to merge it as it is.

null-git:
Hey. I've fixed the perldoc issue, fixed the version and got rid of most shellcheck warnings.

evn is also appropriate for shells. I.e. FreeBSD has /usr/local/bin/bash, Arch Linux has /usr/bin/bash and Debian Jessie has /bin/bash. While I don't have an example right now for sh, other projects like Minix ( https://github.com/Stichting-MINIX-Research-Foundation/minix/blob/master/build.sh ) also do this.

plugins/other/wordpress-multisite
```diff
@@ -0,0 +1,131 @@
...
+	case $1 in
+		comments) QUERY="SELECT COUNT(*) FROM ${DB_PREFIX}${2}comments WHERE comment_approved = '1' AND comment_type = '';" ;;
+		ids) QUERY="SELECT blog_id FROM ${DB_PREFIX}blogs;" ;;
+		pingbacks) QUERY="SELECT COUNT(*) FROM ${DB_PREFIX}${2}comments WHERE comment_approved = '1' AND comment_type = 'pingback';" ;;
+		posts) QUERY="SELECT COUNT(*) FROM ${DB_PREFIX}${2}posts WHERE post_status = 'publish' AND post_password = '' AND post_type = 'post';" ;;
+		title) QUERY="SELECT option_value FROM ${DB_PREFIX}${2}options WHERE option_name = 'siteurl';" ;;
+		users) QUERY="SELECT COUNT(*) FROM ${DB_PREFIX}users;"
+	esac
+
+	$MYSQL_CMD $DB_CONNECTION $DB_OPTIONS $DB_NAME --column-names=0 -s --execute="$QUERY"
```
sumpfralle: MYSQL_CMD and DB_NAME should be quoted in order to avoid whitespace issues.

plugins/other/wordpress-multisite
```diff
@@ -0,0 +1,131 @@
...
+	echo "graph_info Some Statistics of Wordpress"
+	echo "instances.label Instances"
+	echo "users.label Users"
+	echo "posts.label Posts"
+	echo "comments.label Comments"
+	echo "pingbacks.label Pingbacks"
+else
+	for n in $(wp_get ids); do
+		i=
+		test "$n" -gt "1" && i=${n}_
```
sumpfralle: I have problems grasping the relation between `i` and `n`. Could you please add a comment here?

plugins/other/wordpress-multisite
```diff
@@ -0,0 +1,131 @@
...
+	echo "pingbacks.value $PINGBACKS"
+	echo "instances.value $CNT"
+	echo "users.value $(wp_get users)"
+fi
+
+# single blogs
+for n in $(wp_get ids); do
+	i=
+	test "$n" -gt "1" && i=${n}_
+	test "$n" -le "9" && n=0${n}
```
sumpfralle: I understand that you are using this manipulation of `n` in order to influence the sort order of the following multigraphs.
In this case I recommend to pick a longer numerical width and pick an easier understable variable name (e.g. `sort_name=$(printf "%06d" "$n")`). Future readers of this code will surely appreciate this detail.

plugins/other/wordpress-multisite
```diff
@@ -0,0 +1,131 @@
...
+
+# fill vars
+DB_OPTIONS=${mysqlopts}
+DB_CONNECTION=${mysqlconnection:--hlocalhost}
+DB_NAME=${database}
+DB_PREFIX=${dbprefix:-wp_}
+
+MYSQL_CMD=$(which mysql)
+wp_get() {
+	case $1 in
```
sumpfralle: Please add `local QUERY` (or at least expand the `case` with a default branch) in order to avoid hard-to-debug situations.

plugins/other/wordpress-multisite
null-git: done

plugins/other/wordpress-multisite
null-git: done

plugins/other/wordpress-multisite
null-git: Used the printf approach but with configurable number length.
Also changed the variable name and moved the $i part to wp_get() and commented it.

plugins/other/wordpress-multisite
null-git: See next comment

null-git:
Never did anything in perl so I just changed it to /bin/sh for now.

Yeah. /bin is a link to /usr/bin in Arch Linux but it's only there so things like this don't break. But there is no reason why sh shouldn't be anywhere else than /bin so I think the env approach is justified even for sh scripts :)

sumpfralle:
Regarding the location of /bin/sh: yes, probably this is really the future. Somehow it still feels wrong, that /usr/bin/env should be the only universal path, we can rely on. It was surely not meant like that :)

Anyway: thank you for your time and this shiny new plugin!

steveschnepp:
Indeed, I usually don't really like the `/usr/bin/env` in published scripts as it depends on yet another tool to work.

The usual way is to put a simple path, either in `/bin/` for shells & POSIX stuff (such as `awk`), and `/usr/bin/` for anything else (such as `perl`, `python` or `node`).

Let users just change the path if they have a weird setup 😉

PR #710, opened 2016-04-04
added slow requests for php fpm (going multigraph)

drzraf:


drzraf:
@leeclemens (PR #685)

PR #711, opened 2016-04-05
Added variables for logs and command in postfix_stats

CristianDeluxe:
Makes this plugin compatible with other OS (like Ubuntu)


sumpfralle:
Your changes look fine!

If you are in the mood to polish this plugin a bit more, I would suggest the following details:
* use double quotes around "$TMP" or (much better) store the result of pflogsum in a variable instead of writing to a file (and pipe the variable into `grep` via `echo`)
* use `$(...)` instead of backticks for better readability and simpler quoting
* maybe introduce a function for the repetitive combination of grep and awk
* add an autoconf action (which is missing right now) that checks for the presence of pflogsum
Of course, your changes are already fine and can be merged without the above related improvements ...

Could you please get rid of the three merge commits at the end of your commit list? Maybe just discard them, rebase on master (instead of merging it) and forcefully push again?

PR #714, opened 2016-04-16
fixed naming for multiple GPUs in amd_gpu_; adjusted lower limit for …

inbaz:
…temperature to 20


sumpfralle:
Thank you for your improvement proposal!

I am not sure about the sed statement: would it work with a CPU count > 9?

sumpfralle:
@inbaz: what do you think?

inbaz:
Uh, sorry for the late response, yes sure, it needs to be [0-9]+ to support more than 9 GPUs.

sumpfralle:
Thank you!

PR #718, opened 2016-05-14
Added munin_events plugin

szepeviktor:
munin_events - Plugin to monitor munin update log

### DIFF
--- a/plugins/munin/munin_events
+++ b/plugins/munin/munin_events
@@ -39,7 +39,7 @@ INFO, WARNING, ERROR, FATAL.
 
 =head1 VERSION
 
-  1.2.20161017
+  1.2.20160514
 
 =head1 AUTHOR
 
@@ -70,9 +70,9 @@ logtail_bin=${logtail_bin:-/usr/sbin/logtail2}
 do_value() {
     FIELD="$1"
     EVENT_LABEL="$2"
-    EVENT_COUNT="$("$logtail_bin" -t "$muninupdate" 2> /dev/null | grep -c "^[0-9/: ]\{19\} \[${EVENT_LABEL}\]")"
 
-    if ! [ -z "$(echo "$EVENT_COUNT" | sed 's|[0-9]\+||')" ]; then
+    EVENT_COUNT="$("$logtail_bin" -t "$muninupdate" 2> /dev/null | grep -c "^[0-9/: ]\{19\} \[${EVENT_LABEL}\]")"
+    if echo "$EVENT_COUNT" | grep -q "[^0-9]"; then
         echo "Cannot determine event count" 1>&2
         exit 10
     fi
--- a/plugins/munin/munin_events
+++ b/plugins/munin/munin_events
@@ -1,0 +1,142 @@
+#!/bin/bash
+# -*- sh -*-
+: <<=cut
+
+=head1 NAME
+
+munin_events - Plugin to monitor munin updates
+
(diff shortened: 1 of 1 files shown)


sumpfralle:
This new plugin looks great!

Do you think, you could switch from "bash" to "sh"? I think, you do no use any bash-specific features.

Otherwise I would merge it as it is.

szepeviktor:
shellcheck tells me:
```
In munin_events line 70:
    local FIELD="$1"
    ^-- SC2039: In POSIX sh, 'local' is undefined.


In munin_events line 71:
    local EVENT_LABEL="$2"
    ^-- SC2039: In POSIX sh, 'local' is undefined.


In munin_events line 72:
    local EVENT_COUNT
    ^-- SC2039: In POSIX sh, 'local' is undefined.


In munin_events line 75:
    if ! [ -z "${EVENT_COUNT//[0-9]/}" ]; then
               ^-- SC2039: In POSIX sh, string replacement is undefined.


In munin_events line 90:
    "$logtail_bin" "$muninupdate" &> /dev/null
                                  ^-- SC2039: In POSIX sh, &> is undefined.
```

I've never met non-Bash shells.

szepeviktor:
We could skid local-s, but what to do with `${EVENT_COUNT//[0-9]/}`?

szepeviktor:
It is quiet sad to live without Bash...

sumpfralle:
uh - sorry: I did phrase exactly, what I wanted.
Indeed I meant "dash", when I said "sh" - sorry for that.
In your case everything was fine before, I think - just the shebang referred to bash instead of sh.
I am sorry to have wasted your time.

Regarding non-bash shells:
* debian's default (non-interactive) shell was changed from bash to dash some years ago
* embedded environments (e.g. openwrt) run with busybox, which contains ash (quite close to dash) - you do not have enough space on most routers to install bash ...

szepeviktor:
OK-OK
Now it is POSIX compatible.

There is a test failure during `cpanm --notest File::Tail::Multi`

Please merge it.

szepeviktor:
Hello!
I do not know what to do.

Using /bin/sh requires me to be POSIX compatible.
Using /bin/bash may not start  on embedded environments.

Please advise.

sumpfralle:
Hello Viktor,

please use "shellcheck -s dash" for your tests. I guess that your original patch just works.

Lars

szepeviktor:
Should I revert to that with /bin/sh ?
FYI That is POSIX, and my scrpit breaks that.

Would you like me to do that?
Please specify exactly what to do.

-- 
+36204242498
Ezen a készüléken sok az elütés. 
Elnézést!

sumpfralle:
AFAIK /bin/sh points to shell implementations derived from the bourne shell (ash, dash, bash, busybox's sh). Thus the "dash" suite for shellcheck should represent the lowest common denominator of this range of implementations quite well.

Thus please do the following:
* go back to 82edf596896ddcec0e0d75af87790614cfa1f5d0
* replace /bin/bash with /bin/sh
* remove the string replacement in line 75 - probably `echo "$EVENT_COUNT" | grep -q "[^0-9]"` is a suitable alternative
* replace &> in line 90

szepeviktor:
shellcheck is happy about it!

sumpfralle:
It looks good - thank you!
We are discussing the broken test setup on the mailinglist ...

PR #724, opened 2016-06-14
Fix for pymongo > 3.0 and connection support of MongoClient only

nichok:
https://api.mongodb.com/python/current/changelog.html#mongoclient-changes

### DIFF
--- a/plugins/mongodb/mongo_collection_
+++ b/plugins/mongodb/mongo_collection_
@@ -1,4 +1,4 @@
-#!/usr/bin/env /usr/local/bin/python2
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 # vim: set sts=4 sw=4 encoding=utf-8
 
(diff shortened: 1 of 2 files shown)


benpro:
Nice. But should use mongodb:// URI to uses authentication.

nichok:
if you need to do auth with a mongodb URI just config settings_host as:

settings_host = 'mongodb://username:password@host/database

(should work... and keep settings_user empty if you do not want to do db.authenticate)

benpro:
This will not work.
```
conn = pymongo.MongoClient(host,port)
```
The ```,port``` part will not handle it.

nichok:
Well, I changed the settings, though if you repeated the same port in both settings it would work (not nice).

Try it now :smile:

sumpfralle:
@nichok / @benpro: this request is ready for merge, right?

nichok:
yes!

sumpfralle:
@benpro did not object, thus I consider all remaining problems to be solved.
Thank you!

PR #725, opened 2016-06-17
backuppc: Split age graph into last backup / last full backup

plepe:
With this change there are two separate graphs:
* one for the age of the last backup (either incr/full)
* one for the age of the last full backup

If you update your backuppc plugin to the new version, the lines showing the last full backup age will be removed from the "last backup" graph, the new "last full backup" graph will be empty and start a-new.


sumpfralle:
finally mergd ...
Thank you!

PR #726, opened 2016-06-20
Variation of lxc_ plugins which resolve the hostname on Proxmox Hosts

deveth0:
The original lxc_ plugins work fine for all kinds of LXC hosts. But as the amount of containers grows, it becomes quite annoying that the plugins only list the id of the containers.
Those 4 modified versions only work for Proxmox hosts and use the config-files to determine the hostname of the containers.

### DIFF
--- a/plugins/lxc/lxc_pct_cpu
+++ b/plugins/lxc/lxc_pct_cpu
@@ -68,7 +68,7 @@ if [ "$1" = "config" ]; then
  for guest_name in $guest_names;
  do 
   guest="$(clean_fieldname $guest_name)"
-  name=`cat /etc/pve/nodes/proxmox/lxc/$guest_name.conf | grep hostname | awk '{print $2}'`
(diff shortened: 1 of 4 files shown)


sumpfralle:
@deveth0: thank you for your pull request!

Right now you applied all your changes by copying existing plugins and changing a few lines of code.
As much as I would appreciate your improvements with regards to the usage of these plugins on proxmox hosts, this approach is not sustainable for this plugin repository in the long term.

Please incorporate your proxmox-specific changes into the previously existing plugins.
For example the lxc_pct_cpu plugin just needs a check for the existence of a proxmox-specific directory and overwrite the visible hostname in this case.

sumpfralle:
@deveth0: what do you think?

sumpfralle:
@deveth0: Ping?

sumpfralle:
What do you think?

sumpfralle:
Closing this proposal for now.

The inclusion of this change in the existing `lxc_` plugin is still welcome (instead of a code copy).

PR #728, opened 2016-07-12
[plugins/upnpc_] Add upnpc-based router monitoring plugin

shtrom:

### DIFF
--- a/plugins/network/upnpc_
+++ b/plugins/network/upnpc_
@@ -7,6 +7,13 @@
 
 upnpc_ - Plugin to monitor routers via UPnP
 
+This plugin uses the upnpc utility (package miniupnpc in Debian), to monitor an
+router using UPnP. It can monitor the following aspects, and plot them as separate graphs:
+* uptime: how long the link has been up;
+* bitrate: the up and downlink bitrate (e.g., sync speed for DSL);
+* traffic: the actual up and downstream traffic rate;
+* pkts: the number of packets coming in and out.
+
 =head1 APPLICABLE SYSTEMS
 
 Linux systems with upnpc installed.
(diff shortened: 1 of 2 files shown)


sumpfralle:
Hi Olivier,

your plugin looks perfect!

Before merging I took a look at the current plugin directory and I found this one: [plugins/network/upnp_router_](https://github.com/munin-monitoring/contrib/blob/master/plugins/network/upnp_router_)

How would you compare your plugin against the old one?
Maybe we could add a comment to both plugins and find distinguished names describing their differences?
Relocating your plugin to "network/" could also be good due to the similarity of both plugins.

What do you think?

shtrom:
Hey sumpfralle,

On Sun, Oct 16, 2016 at 02:52:52PM -0700, sumpfralle wrote:
> Before merging I took a look at the current plugin directory and I found this one: [plugins/network/upnp_router_](https://github.com/munin-monitoring/contrib/blob/master/plugins/network/upnp_router_)
> How would you compare your plugin against the old one?
> Maybe we could add a comment to both plugins and find distinguished names describing their differences?

Yeah, I discovered it after the fact, too. However, my plugin is a
superset of that one, in that it can also plot the up/downlink bitrates,
as well as the uptime of the connection, in addition to the immediate
traffic measurement.  It also doesn't require PHP (:

The other plugin could *probably* be removed in favour of this one, but
I wouldn't want to break other people's setup.

> Relocating your plugin to "network/" could also be good due to the similarity of both plugins.
> What do you think?

Moving to network/ makes sense (this was my first plugin, so I was a bit
haphazzard in working out where to put it).

-- 

sumpfralle:
Hi Oliver,

adding another upnp plugin with additional features and a simpler interpreter sounds good!

Could you please add the list of plotted details to the documentation of your plugin?
I will do the same for "upnp_router_". This will make it easier for users to pick the right one.

Regarding the names of both plugins: do you think, that "upnpc_" properly reflects the content of your plugin? Otherwise maybe you want to rename it.
Could you move your plugin to the network directory and update your branch, please?

Cheers,
Lars

PR #731, opened 2016-07-16
If a pool is named something like "m" or "ra" or any string that matc…

acranox:
…hes the types of zpool, the grep will match multiple lines.

Adding a space after the zlabel will allow for more exact matching.

### DIFF
--- a/plugins/zfs/zpool_iostat
+++ b/plugins/zfs/zpool_iostat
@@ -40,9 +40,9 @@ echo $zlist | tr ' ' '\n' | while read iz; do
 	*) name=`echo $iz | gawk '{ gsub("[^a-zA-Z0-9_]", "_", $1); print $1 }'` ;;
 	esac
 	echo -n $name'_read.value '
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for this fix!

In order to preserve readibility: could you please check if `awk '{ if ($1 == "'"$zlabel"'") print $7; }' "$ztmp"` would be a suitable replacement for the current combination of grep and the first awk call in these lines? I think it would be easier to understand.

Otherwise I would just merge your original proposal.

acranox:
That works for me.
I may have just messed up my branch though while trying to include that change.

sumpfralle:
Hi acranox,
I am not really used to github's visualization of history, but I am afraid, that you merged master into your branch. Instead you should have rebased your branch on master. Both approaches result in the same code, but different history.
Maybe you want to create a fresh branch, cherry-pick your commits and push the new branch (forcefully) into this one?

acranox:
Thanks.  I pushed an update that should have resolved whatever mess I made last time.

sumpfralle:
Thank you!

PR #734, opened 2016-07-29
added graph for netfilter ipset

Harvie:


sumpfralle:
Thank you for contributing a new plugin!

Please consider the following changes:
* add a POD style documentation (see [processes/multimemory](https://github.com/munin-monitoring/contrib/blob/master/plugins/processes/multimemory) for a good example)
* specify a licence
* add "exit" at the end of the "config" branch
* consider changing "bash" to "sh" (it looks like you do not need bash-specific features)
* remove the "sudo" call and mention the required `user root` config line instead
    * this should allow you to simplify your "autoconf" check ("which" will work in root's context)
* maybe you want to add double-quotes around variables (e.g. in `ipset list $list`)

Did you check, if we may assume that the lists returned by `ipset list` are always [safe](http://munin-monitoring.org/wiki/notes_on_datasource_names)?

sumpfralle:
@Harvie: what do you think?

Harvie:
i don't have time to do this right now? can you merge now, so we can do cosmetics later?

dipohl:
removing sudo is not a _cosmetic_ (!)
It would help your users if you then add a perldoc section "Configuration" with a template for files in plugin-conf.d

concerning license: If you don't address one yourself, it will be published under GPL 2.0, which is the default for Munin contributuion.

sumpfralle:
> I've fixed problems you mentioned

Good!

But please take a look at the handling of `config` now: previously it was intermingled between the first `config` block and the final combined `config`/`fetch` block. Now the branch in the last block is never reached for `config`.

Another two style nitpicks (they are acceptable - but maybe you want to fix them):
* The coding style `a && b || c` may look similar to `if ... then ... else ...`, but it is not the same. E.g. it differs, whenever the last line of code in `b` fails. Thus code linters (e.g. `shellcheck`) will complain about this for a good reason. `if/then/else` is a flawless and readable alternative without performance penalty (everything is done by the shell).
* The tests `[ ... -o/-a ... ]` are not well defined and may show surprising effects, since all substitutions happens at once. This could be problematic, if undefined variables are supposed to be caught by the first test expression. `shellcheck` would also complain about this and suggest to use `[ ... ] && [ ... ]` or `[ ... ] || [ ... ]` instead.

Harvie:
@sumpfralle You are true. I've blindly added exit, but didn't realized that exit at the end of file is common to both exit and fetch. I removed it, so it makes sense again.

> differs, whenever the last line of code in `b` fails

might happen in complex code, but i never had single `echo` failed on me...

> The tests `[ ... -o/-a ... ]` are not well defined and may show surprising effects

never had that problem either

sumpfralle:
Thank you!

> never had that problem either

But nevertheless the wild beasts are out there hiding in the dark :)

I will probably fix these issues up, in order to reduce the number of plugins with failing code quality checks ...

sumpfralle:
> Next time i'll try to get it right at first try...

There is nothing to worry about :)

Thank you for your persistence!

PR #735, opened 2016-08-01
add wildcard version of rabbitmq messages and consumer plugins

kouk:
Based on the existing rabbitmq plugins here are two new wildcard plugins which run for each virtual host present.

Additional options:
- `ignore`: a regex to ignore certain queues

### DIFF
--- a/plugins/rabbitmq/rabbitmq.inc
+++ b/plugins/rabbitmq/rabbitmq.inc
@@ -1,0 +1,33 @@
+CMD=$(which rabbitmqctl)
+
+if [ "$1" = "autoconf" ]; then
+    test -x "$CMD" && echo yes || echo "no ('$CMD' not executable)"
+	exit 0
+fi
+
+if [ "$1" = "suggest" ]; then
+    for h in $(HOME=$HOME sudo $CMD list_vhosts -q); do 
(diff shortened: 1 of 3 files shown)


sumpfralle:
Hi kouk,

thank you for adding more features to the rabbitmq plugins!

I see a few issues with your proposal:
* if you just want to extend the previous plugin by looping over multiple hosts, then you should incorporate this into the original plugin. E.g. using symlink configuration should be a good approach.
* Specify a user in the plugin configuration instead of using sudo.
* Please do not use a shared include file, but keep each plugin self-sufficient even if you need to copy a few lines. Munin's plugin style values simplicity higher than code deduplication.
* If you are feeling really adventurous, then you want to start merging the rabbitmq plugins into a single file controlled via symlink configuration. All current plugins differ only by a few lines - thus it should be possible to create a merged plugin with all current features while still keeping it simple.

What do you think?

sumpfralle:
@kouk: any thoughts?

sumpfralle:
@kouk: what do you think? Do you want to create a multigraph plugin?

sumpfralle:
Please open a new pull request, if you want to propose a multigraph plugin instead.

PR #736, opened 2016-08-02
fix colored border and highlights

swus13:
add style from default css for borders and colored highlights


templates/munstrap/static/css/style-munstrap.css
```diff
@@ -35,10 +35,34 @@ ul.groupview, ul.groupview ul {
 	list-style-type: none;
 }
 .munin-icon {
-   background: url(../img/logo-munin.png) left top; 
+   background: url(../img/logo-munin.png) left top;
```
TafThorne: A previous.  I prefer layout changes in a seperate patch posted after the functional ones. It is entirely up to the actual maintainers of this project to say what they prefer though.

sumpfralle:
@swus13: do you want to integrate @TafThorne comment's into your pull request?

However: it seems like no one raised concerns, thus I would merge your changes.

It is a pity that there are so few people involved with layout and web design in munin. I am sorry for the lack of response regarding your improvements.

swus13:
sorry, appending white spaces removed the IDE by itself, which is not my intention to change that.
I hope this is no problem now and can remain so in the commit.

sumpfralle:
yes, these few added formatting changes are no problem - thank you!

PR #737, opened 2016-08-03
lxd plugins

felix-engelmann:
lxd is the abstraction of lxc containers and provides multiple additional services. Through a REST interface and pylxd it provides a lot of metrics about running containers which can be graphed. To run it, the munin user must either be in the lxd group or the plugin must run as root in the `munin-node` config

    [lxd_*]
    user root

Creating a new plugin for another value, such as network traffic, can easily be copied from these ones.

### DIFF
--- a/plugins/lxd/lxd_disk
+++ b/plugins/lxd/lxd_disk
@@ -1,0 +1,26 @@
+#!/usr/bin/python3
+
+import sys
+from pylxd import api
+
+c=api.API()
+
+if len(sys.argv) == 2:
+    if sys.argv[1]=="autoconf":
+        print("yes")
+        sys.exit(0)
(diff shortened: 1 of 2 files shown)


plugins/lxd/lxd_disk
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/python3
+
+import sys
+from pylxd import api
+
+c=api.API()
+
+if len(sys.argv) == 2:
+    if sys.argv[1]=="autoconf":
```
ssm: There should be a check before autoconf answers "yes", and it should default to "no".  A good check is the API responding, or a required command is present in the path.

ssm:
Very nice, thanks.

PR #738, opened 2016-08-03
Ubuntu 14.04 support, new graph option

mittyorz:
- add Ubuntu 14.04 support for lxc_proc
- add new graph draw type option for lxc_ram
  - example: https://github.com/mittyorz/munin-plugins/wiki/LXC#envareastack

### DIFF
--- a/plugins/lxc/lxc_ram
+++ b/plugins/lxc/lxc_ram
@@ -12,9 +12,17 @@ lxc_ram - Plugin to monitor LXC memory usage.
   [lxc_*]
     user root
 
+  [lxc_ram]
+    env.stack_usage true
+
 =head1 INTERPRETATION
 
 This plugin needs root privilege.
+
+If env.stack_usage is set to true, all memory usages of containers will be
+ drawn as stacked area charts.
+This option changes graph order, all of 'Mem usage' comes first and then others.
+(default: empty)
 
 =head1 AUTHOR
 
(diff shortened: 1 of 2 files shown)


plugins/lxc/lxc_ram
```diff
@@ -62,13 +70,24 @@ if [ "$1" = "config" ]; then
  echo 'graph_vlabel byte'
  echo 'graph_category lxc'
 
+ if [ "$areastack" = "true" ]; then
+  for guest_name in $guest_names; do
+    guest="$(clean_fieldname $guest_name)"
+
+    echo 'mem_usage_'$guest'.label '$guest_name': Mem usage'
```
sumpfralle: You may want to use double quotes here in order to improve readability.

plugins/lxc/lxc_ram
```diff
@@ -62,13 +70,24 @@ if [ "$1" = "config" ]; then
  echo 'graph_vlabel byte'
  echo 'graph_category lxc'
 
+ if [ "$areastack" = "true" ]; then
```
sumpfralle: Splitting the two conditional branches for areastack (line 73 and line 87) makes it quite hard to understand the logic of the code.
Maybe you want to change the code in a way to always output the mem_usage first and then follow with the other datasources. Each datasource will need a separate guest-loop in this case, but it will be easy to understand and maintain.

mittyorz:
@sumpfralle Thanks a lot for your review, and please excuse my late reply.

> Splitting the two conditional branches for areastack (line 73 and line 87) makes it quite hard to understand the logic of the code.
I fixed it to simple if-else branches.

> You may want to use double quotes here in order to improve readability.
Yes, and I'll fix all of them by next another PR because other plugins also have same code.

plugins/lxc/lxc_ram
```diff
@@ -62,13 +70,18 @@ if [ "$1" = "config" ]; then
...
  echo 'graph_category memory'
 
+ if [ "$areastack" = "true" ]; then
+  for guest_name in $guest_names; do
+    guest="$(clean_fieldname $guest_name)"
 
- for guest_name in $guest_names;
-  do
-   guest="$(clean_fieldname $guest_name)"
+    echo 'mem_usage_'$guest'.label '$guest_name': Mem usage'
```
sumpfralle: The quoting of this line (and the ones following) is a bit hard to read (and I think it is misleading).
I assume that you just want to do the following?
```
echo "mem_usage_$guest.label $guest_name: Mem usage"
```
Single quotes: no variable expansion inside.
Double quotes: variable expansion happens inside.

sumpfralle:
What do you think about my comment regarding the quoting?
Should I just merge it as it is?

sumpfralle:
@mittyorz: do you want to improve the readability of the quoting?
Otherwise I will merge it as it is.

PR #739, opened 2016-08-04
plugins: network/udp-statistics: added more fields

RubenKelevra:


RubenKelevra:
Any progress?

sumpfralle:
Thank you for your changes!
Sorry for the delay ...

Your change removes the old "errors" attribute. Could you please comment this in the commit message if this is in line with your intention?

Btw: are you the original author of this plugin? We would like to clean up code authorship and licensing of the plugins ...

RubenKelevra:
The error attribute wasn't deleted, it just moved. The attributes are now sorted accordingly to the output of the screen readed tool just to make it a bit clearer to read the output.

Nope I'm not the original author, feel free to add me somewhere if necessary, else my commit is added to the project anyway so that's fine for me.

sumpfralle:
Sorry for my confusing wording: I meant to say that your change will hide old values/statistics of the "errors" attribute from the current users of this plugin. This inevitably happens when an attribute is renamed. User could potentially work around this issue, if they rename their rrd files manually.
I agree with you, that your suggested change of fieldnames is desirable for this plugin in the long term.

For me the only problematic point is the title of the plugin ("add more fields"). It should also reflect the renaming of the "errors" attribute.

RubenKelevra:
Actually the old field name "errors" was just wrong, there are several types of errors and the old version just counted one while displaying this like "all errors".

I guess hiding old, wrongly labeled values is a better solution than everything else.

Feel free to add a more descriptive point at the changelog for the version which contains this commit. But I'm fine with short commit messages, and don't want to invest the time to fix the commit history here with all the git stuff I need for this.

sumpfralle:
Yes, as I said - I agree with your rationale of renaming the fieldname. I appreciate this change.

Regarding the commit message:
We try to maintain this repository of contributed plugins  in a way that follows the principle of least surprise for their users. Changes that entail the loss of data (here: historcal statistics) should at least mention this fact in the commit message. The users of these plugins will appreciate this care.

The following steps are sufficient for rewriting the history for this case:
 ```
git commit --amend -m "[udp-statistics] rename 'errors' field, add more fields"
git push -f
```
Thank you!

RubenKelevra:
@sumpfralle thanks for the commands :)

sumpfralle:
thank you!

RubenKelevra:
@sumpfralle I just looked at the repo, just to find out that @dipohl uploaded my change to a new location, without any history - so he's basicly the only author.

If my work is threated that way, I really don't see a point contributing again - what was actually the reason to come back here.

sumpfralle:
Hello Ruben,

Thank you for your contribution - and indeed there is no chance, it could get lost!

Here is your commit: 50c13da584f1ab77c1727ff029734e5a6af8b0df

@dipohl recently worked on improving the directory structure for the plugins (e.g. in order to make the plugin gallery more useful). During this operation she moved plugins to more appropriate locations. Such operations do not claim any kind of ownership.

Tracking the history of a file that was renamed or moved usually requires the additional parameter `--follow` for each git command.
E.g. see the output of `git log --follow -p -- plugins/udp/udp-statistics`.

I doubt that there are more than a dozen files in the munin project or the plugin repository, that *never* got renamed or moved over their lifetime. Thankfully - after the era of CVS ended - version control software can deal with these events nicely and without forgetting the original contributors like you.

Cheers,
Lars

RubenKelevra:
@sumpfralle 
I didn't used git, so github isn't showing anything in this case - which was quite confusing.

Thanks for the clearup. :)

Best regards

Ruben

PR #741, opened 2016-08-08
Add job stats graph

yoe:
job stats (running/queued/submitted) for gridengine

### DIFF
--- a/plugins/sge/sge_job_stats
+++ b/plugins/sge/sge_job_stats
@@ -5,6 +5,29 @@
 #
 # Tested with Son of Gridengine.
 
+: <<=cut
+
+=head1 NAME
+
+sge_job_stats - Munin plugin to monitor jobs submitted and executed on a Grid
+Engine installation
+
+=head1 APPLICABLE SYSTEMS
+
+Any system with a qstat command that behaves like the one found in Son of
+GridEngine.
+
+=head1 CONFIGURATION
+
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for this new plugin!

I would suggest to change some minor details before merging it:
* consider to switch the shebang to /bin/sh - it looks like you do not need any bash-specific features
* please add a POD style documentation (see the [processes/multimemory](https://github.com/munin-monitoring/contrib/blob/master/plugins/processes/multimemory) plugin for a good example)
    * adding the "autoconf" and "suggest" capabilities would be perfect, if applicable (see [wiki](http://munin-monitoring.org/wiki/PluginConcise#autoconf))
* please mention the implicite requirement (due to the sourcing from /opt/)
    * in case of a variable installation directory (for different distributions) it could be helpful to allow a path override via the environment (e.g. `SGE_SHELL_LIB_PATH=${SGE_SHELL_LIB_PATH:-/opt/sge/default/common/settings.sh}`)

sumpfralle:
Thank you for the shebang and the configurable location.

Please also add a POD-style documentation header. Otherwise the plugin will not show up properly in the [plugin gallery](http://gallery.munin-monitoring.org/).
We are already a bit overwhelmed with the task of cleaning up all existing contributed plugins - thus we need to require the documentation header for all new plugins before merging. Thank you!

sumpfralle:
Thank you!

PR #742, opened 2016-08-08
SenderBase Plugin

asciiphil:
This adds a plugin for the SenderBase email server reputation service.

### DIFF
--- a/plugins/senderbase/senderbase
+++ b/plugins/senderbase/senderbase
@@ -19,6 +19,12 @@ variable to the appropriate value.
 
 Phil! Gold <phil_g@pobox.com>
 
+=head1 LICENSE
+
+This plugin is made available under a CC0 waiver:
+
+  https://creativecommons.org/publicdomain/zero/1.0/
+
 =head1 MAGIC MARKERS
 
  #%# family=auto
--- a/plugins/senderbase/senderbase
+++ b/plugins/senderbase/senderbase
@@ -42,20 +42,36 @@ This plugin is made available under a CC0 waiver:
 
 =cut
 
-real_ip_address=${ip_address:-$(hostname -I | awk '{print $1}')}
-ip_reversed=$(echo $real_ip_address | sed -re 's/^([0-9]+)\.([0-9]+)\.([0-9]+)\.([0-9]+)$/\4.\3.\2.\1/')
-query_host=${ip_reversed}.rf.senderbase.org
+if which dig >/dev/null; then
+    query_program=dig
+    do_query () {
+        hostname=$1
+        dig +short $hostname TXT | sed 's/"//g' | head -1
+    }
+elif which host >/dev/null; then
+    query_program=host
+    do_query () {
+        hostname=$1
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for contributing this plugin!

I would propose the following small changes:
* specify a license
* add a link to the some information about the "senderbase" concept

Please verify the output of "hostname -I": here on my system (Debian Jessie) it returns multiple IPs separated by spaces on a single line. You are expecting single IPs on multiple lines instead.
Maybe the following snippet works better? It would even provide a reproducible IP selection ("hostname" does not specify the order of results).
`hostname -I | sed 's/[[:space:]]/\n/g' | grep . | sort | head -1`

Maybe you want to replace "dig" with "host" in order to avoid the additional dependency? I would suggest the following:
`host -t TXT "${query_host}.rf.senderbase.org" | cut -f 2 -d '"'`
You could move the TXT query to a separate function in order to avoid code duplication.

I would also recommend double quotes around most variables.

Anyway: it is your plugin - style it in the way you like ...

asciiphil:
I think I've addressed most of your comments.

I totally misread the `hostname -I` documentation previously; I now take the first address in the list.  I'd rather just leave it at that (rather than sorting them) and let people use the `ip_address` environment variable if they need disambiguation.

`dig` is a lot easier to script than `host` (especially when you need to handle nonexistent records), but I think I've got `host` working.  I considered adding `nslookup` support but didn't feel like the effort was worth it.  :-/

I'd previously left out quotes around variables that shouldn't have whitespace in them, but it's probably more defensive to put them in, so I did.

(It looks like the CI checks failed, but not because of my code.)

sumpfralle:
Yes, the CI is currently missing an unrelated dependency. We can ignore this for the moment.

I like your changes and would be happy to merge them.

Just a last question occured to me:
As far as I understand your script, it would not be possible to monitor more than one mail server on every munin-node host (only via manual symlinks and multiple plugin configurations).
Do you think, that users will want to follow only a single host? Could maybe host selection via plugin symlink name suffixes could be suitable for this plugin (e.g. "senderbase_127_0_0_1")? Or allowing multiple addresses to be given via the environment variable?
Anyway: of course, it can stay like it is, if you prefer it that way.

asciiphil:
Hm.  My personal preference is to have the monitoring on the mail server, so all of the graphs relating to that system are on the same webpage.  I could see a scenario where someone would want to monitor multiple servers (a pool, perhaps) from one, central host.

Lemme see about adding support for multiple hosts, probably via the `ip_address` environment variable.

sumpfralle:
I agree with your approach o monitoring on host offering the service. Here in my case this breaks down to two different public IPv4 addresses. This requirement is probably rare, but will grow with mixed IPv4/IPv6 deployments.

Anyway: just let me know if I should merge your plugin in its current state. You can add more (hopefully non-breaking) changes later. The decision is up to you.

asciiphil:
Go ahead and merge it.  The idea I've got for adding multiple hosts won't change the way the plugin currently works with single hosts.

PR #743, opened 2016-08-11
fixed lxd_disk for stopped containers

felix-engelmann:
stopped containers have no disk space info and the plugin crashed. Now
it only shows disk usage for running containers

### DIFF
--- a/plugins/lxd/lxd_disk
+++ b/plugins/lxd/lxd_disk
@@ -10,7 +10,7 @@ try:
 except:
     HAS_LIB=False
     errors.append("no pylxd module")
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for fixing this issue!

While reading the changed lines I was tempted to squash the "if" conditional and the "for" loop into a simple "for" loop. But then I understood, that this is exactly your point: the disc info of a stopped container is not iterable :(
Could you please add a comment regarding the non-iterable behaviour in this specific case? Otherwise people could be tempted to "simplify" your fix :)

PR #744, opened 2016-08-15
Add systemd plugin counting units in each state

shtrom:


shtrom:
There seems to be an error in the Travis config, the checks failing here don't seem to be due to changes in this PR.

avian2:
Thanks for making this plugin! 

Please consider also including the change in this branch: https://github.com/avian2/munin-contrib/tree/add-systemd It fixes the following warning in `munin-update.log` caused by empty extinfo lines:

    [WARNING] 4 lines had errors while 8 lines were correct (33.33%) in data from 'fetch systemd'

shtrom:
Can you send a PR against my branch? (:

avian2:
By the way, you still have a pull request opened against your branch. https://github.com/shtrom/munin-contrib/pull/1

shtrom:
Ah, thanks. I didn't get a notification about it :thinking: Merged now.

sumpfralle:
Hi Oliver,

your new plugin currently half-way supports "suggest".
I would recommend to either remove references to "suggest" (capability, "mode") or add the missing pieces:
* append an underscore to the name of the plugin
* enable the currently inactive "mode" assignment

Additionally you should take a look at the indentation between line 58 and 65.
You may also want to add some more double quotes around variables and replace backticks with "$(...)".

sumpfralle:
We are almost there :)

* please remove the remaining pieces of the "mode" selection - they are confusing for future readers/committers
* please use an explicit test expression (`-n`) in line 107 (`[ "$extinfo" ]`) for better readability
* remove the dependency for `mtemp`. You could simply store the result in a variable and `echo` it into the `grep` call. Otherwise you would need to `trap` any errors or unexpected situations in order to clean up the temporary file.
* rename the plugin to something like `systemd_units` or `systemd_unit_states` - whatever seems like the most suitable one for you. Maybe someone else will want to add more plugins into the "systemd" namespace later ...
* extra bonus: take a look at the outpuf of `shellcheck -s dash SCRIPT` and add some double-quotes in order to avoid accidental tokenizing, globbing and other surprises. Additionally it will save time for anyone reviewing and improving this plugin later, since he/shell will not need to check for potential whitespace issues again and again.

shtrom:
Done. The two remaining warnings from `shellcheck` are voluntary to preserve the desired behaviour (nice tool, btw, I didn't know about it), and I couldn't avoid the temporary file, as using `echo` otherwise loses the newlines, making `grep` calls much less useful (:

avian2:
@shtrom Take a look at https://github.com/shtrom/munin-contrib/pull/2. It should fix remaining issues mentioned by @sumpfralle.

sumpfralle:
perfect!

PR #745, opened 2016-08-23
Skip over comment lines in dhcpd.conf

acranox:
to avoid finding random subnets in the comments.
Without this comment lines will end up getting captured, and statistics erroneously generated for them.


sumpfralle:
Thank you for taking the time to fix this issue!
avian2 created a pull request (#752) some days after you, which also fixed another issue.
That pull request was merged a few days ago - thus your fix is merged now, as well.

PR #749, opened 2016-09-03
[deborphan] New plugin counting orphaned packages in all sections

shtrom:
Signed-off-by: Olivier Mehani shtrom@ssji.net

### DIFF
--- a/plugins/system/deborphan
+++ b/plugins/system/deborphan
@@ -1,0 +1,64 @@
+#!/bin/bash
+#
+# Plugin to monitor the number of orphaned packages on the
+# system (using deborphan). Might work on section distib, who knows...
+#
+# Based on the debsecan plugin by Nicolas Bouthors. 2016-09-02
+#
+# Olivie Mehani <shtrom@ssji.net>
+#
+# Licence : GPLv2
+#
+#%# family=auto
+#%# capabilities=autoconf
+
+# Auto enable if we have deborphan only
+if [ "$1" = "autoconf" ] ; then
+  if [ -x /usr/bin/deborphan ]; then
+    echo yes
+  else
+    echo no
+  fi
+  exit 0
+fi
+
+# Fail if we don't have deborphan
+if [ ! -x /usr/bin/deborphan ]; then
+  exit 1
+fi
+
+OUT=`mktemp -t deborphan.XXXXXX`
+deborphan --show-section --guess-all | sed 's/\//_/' | sort > ${OUT}
+
+CATEGORIES="$(sed 's/ .*//' ${OUT} | uniq)"
+
+if [ "$1" = "config" ]; then
+  cat <<EOF
+graph_title Orphaned packages
+graph_args -l 0 --base 1000
+graph_vlabel number of packages
+graph_category system
+graph_period second
+graph_info This graph show the number of orphaned packages on your system. Use deborphan to see details.
+EOF
+  for cat in ${CATEGORIES}; do
+    cat <<EOF
+${cat}.label ${cat/_/\/}
+${cat}.type GAUGE
+${cat}.draw AREASTACK
+${cat}.min 0
+${cat}.info The number of orphaned packages in ${cat/_/\/}
+EOF
+  done
+
+else
+  for cat in ${CATEGORIES}; do
+    TMP=`mktemp -t deborphan.XXXXXX`
+    sed -n "s/${cat} *//p" ${OUT} > ${TMP}
+    echo "${cat}.value `cat ${TMP} | wc -l`"
+    echo "${cat}.extinfo `echo $(cat ${TMP})`"
+    rm ${TMP}
+  done
+fi
+
+rm -f ${OUT}
--- a/plugins/system/deborphan
+++ b/plugins/system/deborphan
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 : << =cut
 
--- a/plugins/system/deborphan
+++ b/plugins/system/deborphan
@@ -1,36 +1,57 @@
 #!/bin/bash
-#
-# Plugin to monitor the number of orphaned packages on the
-# system (using deborphan). Might work on section distib, who knows...
-#
-# Based on the debsecan plugin by Nicolas Bouthors. 2016-09-02
-#
-# Olivie Mehani <shtrom@ssji.net>
-#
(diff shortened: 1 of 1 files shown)


shtrom:
Travis failed with a `cpanm` error not linked to this PR:

    $ cpanm --notest File::Tail::Multi
    ! Finding File::Tail::Multi on cpanmetadb failed.
    ! Finding File::Tail::Multi () on mirror http://www.cpan.org failed.
    ! Couldn't find module or a distribution File::Tail::Multi
    The command "cpanm --notest File::Tail::Multi" failed and exited with 1 during .

plugins/system/deborphan
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+#
+# Plugin to monitor the number of orphaned packages on the
```
sumpfralle: Please turn the documentation into POD format.

plugins/system/deborphan
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
sumpfralle: /bin/sh should work, as well.

plugins/system/deborphan
```diff
@@ -0,0 +1,64 @@
...
+# Olivie Mehani <shtrom@ssji.net>
+#
+# Licence : GPLv2
+#
+#%# family=auto
+#%# capabilities=autoconf
+
+# Auto enable if we have deborphan only
+if [ "$1" = "autoconf" ] ; then
+  if [ -x /usr/bin/deborphan ]; then
```
sumpfralle: Please use `which deborphan >/dev/null` for a distribution/packaging-independent test.

plugins/system/deborphan
```diff
@@ -0,0 +1,64 @@
...
+    echo yes
+  else
+    echo no
```
sumpfralle: Add the reason for failure in brackets.

plugins/system/deborphan
```diff
@@ -0,0 +1,64 @@
...
+  fi
+  exit 0
+fi
+
+# Fail if we don't have deborphan
+if [ ! -x /usr/bin/deborphan ]; then
+  exit 1
+fi
+
+OUT=`mktemp -t deborphan.XXXXXX`
```
sumpfralle: You should avoid temporary files (or carefully track their removal via `trap`). Storing the result in a variable should work well here.

plugins/system/deborphan
```diff
@@ -0,0 +1,64 @@
...
+deborphan --show-section --guess-all | sed 's/\//_/' | sort > ${OUT}
+
+CATEGORIES="$(sed 's/ .*//' ${OUT} | uniq)"
```
sumpfralle: `awk '{print $1}'` should be easier to read - at least for people who do not remember the output format of deborphan by heart (like me).

plugins/system/deborphan
```diff
@@ -0,0 +1,64 @@
...
+${cat}.draw AREASTACK
+${cat}.min 0
+${cat}.info The number of orphaned packages in ${cat/_/\/}
+EOF
+  done
+
+else
+  for cat in ${CATEGORIES}; do
+    TMP=`mktemp -t deborphan.XXXXXX`
+    sed -n "s/${cat} *//p" ${OUT} > ${TMP}
```
sumpfralle: This pattern could match unexpectedly (e.g. "main/text" would match "cat=main/tex").
Maybe use `awk '{ if ($1 == "'"$cat"'") print $1}'`?
Using a variable instead of a file would be nice here, as well.

plugins/system/deborphan
```diff
@@ -0,0 +1,64 @@
...
+    echo "${cat}.value `cat ${TMP} | wc -l`"
+    echo "${cat}.extinfo `echo $(cat ${TMP})`"
```
sumpfralle: Does the formatting of this extinfo look ok? Or should it be space separated?
Please use `$(...)` instead of backticks for improved readability. It can be nested.
`echo $(cat foo)` is the same as `cat foo`.

plugins/system/deborphan
```diff
@@ -0,0 +1,64 @@
...
+${cat}.min 0
+${cat}.info The number of orphaned packages in ${cat/_/\/}
+EOF
+  done
+
+else
+  for cat in ${CATEGORIES}; do
+    TMP=`mktemp -t deborphan.XXXXXX`
+    sed -n "s/${cat} *//p" ${OUT} > ${TMP}
+    echo "${cat}.value `cat ${TMP} | wc -l`"
```
sumpfralle: Here on my system (Debian testing), deborphan outputs categories including a slash. This is not allowed in fieldnames. Probably you want to include  `$MUNIN_LIBDIR/plugins/plugin.sh` and use `clean_fieldname`.

plugins/system/deborphan
```diff
@@ -0,0 +1,64 @@
...
+    echo yes
+  else
+    echo no
+  fi
+  exit 0
+fi
+
+# Fail if we don't have deborphan
+if [ ! -x /usr/bin/deborphan ]; then
+  exit 1
```
sumpfralle: Please emit a warning instead of failing silently. This helps users when trying "munin-run".

plugins/system/deborphan
shtrom: Not really, as I'm using some bashism to do some quick pattern replacement later on.

plugins/system/deborphan
shtrom: I fixed that with `trap` (:

plugins/system/deborphan
shtrom: It's already taken care of in the call to `deborphan`: there's a `sed` that converts all slashes to underscores.

shtrom:
Should be good for another look.

plugins/system/deborphan
```diff
@@ -0,0 +1,85 @@
...
+
+# Fail if we don't have deborphan
+if ! which deborphan >/dev/null; then
+  echo "deborphan is missing" >&2
+  exit 1
+fi
+
+OUT=$(mktemp -t munin-deborphan.XXXXXX)
+TMPFILES=$OUT
+trap 'rm -f $TMPFILES' EXIT
```
sumpfralle: I would still prefer to not use files at all (avoiding any potential whitespace issue for `rm`), but this should work, if you really want to insist on it :)

PR #750, opened 2016-09-03
[file_length_] A versatile plugin to count the lines in specified files

shtrom:

### DIFF
--- a/plugins/system/file_length_
+++ b/plugins/system/file_length_
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+: << =cut
+
+=head1 NAME
+
+file_length_ - Plugin to monitor the length of specified files
+
+Useful for things such as lists (white, black, user, ...).
+
+=head1 CONFIGURATION
+
+    [file_length_IDENTIFIER]
+    env.files FILEPATHGLOB1 FILEPATHGLOB2 ...
+    env.category DEFAULTS_TO_system
+    env.title OPTIONAL_TITLE
+
+=head1 AUTHOR
+
+Olivier Mehani (based on disk/log_sizes)
+
+=head1 LICENSE
+
+GPLv2
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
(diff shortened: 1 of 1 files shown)


plugins/system/file_length_
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
sumpfralle: /bin/h should be sufficient.

plugins/system/file_length_
```diff
@@ -0,0 +1,65 @@
...
+
+=cut
+
+#NAME=`echo $0 | sed 's/.*_//'`
+NAME=${0#*_}
+TITLE=${title:-File lengths for $NAME}
+CATEGORY=${category:-system}
+
+FILES=${files:-/var/log/messages}
+FILES=$(echo $(ls $FILES))
```
sumpfralle: `echo $FILES` should be sufficient for globbing (if this your point). Please add a comment regarding the purpose of this line.

plugins/system/file_length_
```diff
@@ -0,0 +1,65 @@
...
+	cat <<EOF
+graph_title ${TITLE}
+graph_args ${graph_args} --base 1000
+graph_category ${CATEGORY}
+graph_info This graph shows the length of ${FILES}
+graph_vlabel length (lines)
+EOF
+
+	for F in $FILES; do
+		MF=`echo $F | sed 's/[-\/\.]/_/g'`
```
sumpfralle: Please use "clean_fieldname" from "$MUNIN_LIBDIR/plugins/plugin.sh" instead.

plugins/system/file_length_
shtrom: Nope, needed for `echo -n`, otherwise `shellcheck` complains (:

shtrom: Ah, actually, no, I know how to work around that.

shtrom:
Should be good for another look.

PR #751, opened 2016-09-26
Add Chilli plugin for sessions state

gmarsay:


plugins/chilli/chilli_
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
```
sumpfralle: Please change /bin/bash to /bin/sh, if possible. This will ease the usage of this plugin on platforms without a default installation of bash (embedded systems, some *BSD, AIX, ...).
The only required change below seems to be the rewriite of the tests:
* replace double square brackets with single ones
* replace `==` with `=`

sumpfralle:
@gmarsay: what do you think?

gmarsay:
@sumpfralle : I updated the script by following your recommendations !
I would like to keep the category "chilli", other plugins will be added in this category.

plugins/chilli/chilli_sessions_
```diff
@@ -0,0 +1,117 @@
...
+          exit 0
+        fi
+      fi
+    else
+      echo "no ($CHILLI_PATH_BIN not found)"
+      exit 0
+    fi
+    ;;
+  suggest)
+      INSTANCES_LIST=$(ls /var/run/chilli_*.sock)
```
sumpfralle: This statement will not work with sockets containing a space. Additionally it neglects "CHILLI_PATH_SOCK".

How about this one?
```
find "$CHILLI_PATH_SOCK/" -name "chilli_*.sock" | while read filename; do
    basename "$file" .sock | cut -d _ -f 2
done
```
This should also take care for `shellcheck` warnings regarding quoting.

plugins/chilli/chilli_sessions_
```diff
@@ -0,0 +1,117 @@
...
+
+      for file in $INSTANCES_LIST; do
+        basename "$file" .sock | cut -d _ -f 2
+      done
+
+      echo "total"
+      
+            exit 0
```
sumpfralle: This indentation looks weird.

plugins/chilli/chilli_sessions_
```diff
@@ -0,0 +1,117 @@
...
+    exit 0
+    ;;
+esac
+
+if [ "$INSTANCE" = "total" ]; then
+  STATE_PASS=$("$CHILLI_PATH_BIN" list | grep -wc "pass")
+  STATE_DNAT=$("$CHILLI_PATH_BIN" list | grep -wc "dnat")
+  STATE_NONE=$("$CHILLI_PATH_BIN" list | grep -wc "none")
+else
+  STATE_PASS=$("$CHILLI_PATH_BIN" -s $CHILLI_PATH_SOCK/chilli_"$INSTANCE".sock list | grep -wc "pass")
```
sumpfralle: This would fail, if `$CHILLI_PATH_SOCK` contained whitespace. Just quote the full socket patch instead of only the INSTANCE variable. IMO, this also enhances readability.

plugins/chilli/chilli_sessions_
```diff
@@ -0,0 +1,117 @@
+#!/bin/sh
+# -*- sh -*-
+
+: << =cut
+
+=head1 NAME
+
+chilli_sessions_ - Wildcard-plugin to monitor sessions state on Coova Chilli.
+
```
sumpfralle: Please add a short description (`=head1 DESCRIPTION`) of "chilli" (radius ...) and the scope of this plugin (number of active, blocked and unknown(?)) users/connections/devices(?) for each session (or is there a better word?).

gmarsay:
@sumpfralle : I updated the script by following your recommendations, it's ok ?

sumpfralle:
Yes, I like it. Thank you!

PR #764, opened 2016-10-26
Prevent data corruption if output fields of "tc -s" vary, stack children lines

nyetwurk:
Also a few cosmetic changes.


sumpfralle:
Hi nyetwurk,

your changes look good and I am ready to merge them in their current state.

But I would be glad, if you were willing to polish this plugin a bit more:
* change the documentation header into the POD format (e.g. see plugins/processes/multimemory) - this allows `munindoc tc_`
* add double-quotes around shell variables (usually `$DEVICE`) - see `shellcheck -s dash tc_` for details
* use `grep -E` instead of legacy `egrep`
* change `family` to `auto` (since the plugin supports "autoconf" and "suggest")
* combine the multiple `sed` statements in `mytc` into a single one (a semicolon separated list of sed statements) - this will make a difference for embedded devices
* remove the superfluous comment from the "suggest" branch
* add your graphs generated by this plugin as an example graph (see [instructions](http://munin-monitoring.org/wiki/PluginGallery#Rulesforplugincontributors))

Most of these tasks can only be done by real users of this plugin. This is why I ask you ...
Anyway: if you do not feel like, I will merge your above fixes in their current state.

dipohl:
Hi nyetwurk,

thanks for improving the plugin!

I want to add some words concerning the plugin documentation.

sumpfralle wrote:
>  change the documentation header into the POD format (e.g. see plugins/processes/multimemory) - this allows `munindoc tc_`

The documentation header in POD format is **needed** to show a doc page for the plugin in the Munin-Gallery. At the time the gallery will [only show hint on missing doc for this plugin](http://gallery.munin-monitoring.org/contrib/plugins/network/tc_.html)

See the following instructions about requirements for plugins to show up nice in the gallery: http://munin-monitoring.org/wiki/PluginGallery

In my opinion better templates than plugins/processes/multimemory are:

http://gallery.munin-monitoring.org/distro/plugins/node.d.linux/buddyinfo.html
http://gallery.munin-monitoring.org/distro/plugins/node.d/munin_update.html

nyetwurk:
I will work on it time permitting. Thank you for your guidance!

nyetwurk:
Ugh, apologies for the .pod thrashing, not sure what to do there still, especially for the APPLICABLE SYSTEM section, which should probably just be removed.

sumpfralle:
The code looks fine now.
Thank you for the example graphs!

Regarding the documentation:
"Applicable Systems" usually contains a human readable description of the context suitable for this plugin - here I would mention the required linux kernel version (as you did) and "traffic control" or "traffic shaping" - whatever seems suitable for you. Afterwards I will merge your commits.

Regarding the separate .pod file: I did not realize up to now, that this is a documented possibility (instead of embedding it in the header of the script file). "tc_" will be the first plugin in the contrib repository choosing this approach. Keep it like that if you prefer so.

dipohl:
Thanks for the plugins doc! Unfortunately the gallery build script doesn't look for .pod files (there were only few such cases at the time, when I wrote the script. So I've choose the simple approach and included those view .pod files into the plugins code files)

I didn't write that on the wiki page about the Plugin Gallery, so the misleading instruction was my fault (I will correct that when I move the page to the Munin Guide very soon).

Please include the contents from .pod file into the plugin code file and sorry for the inconvenience.

sumpfralle:
I think you forgot to add the documentation to the script :)

sumpfralle:
thank you!

dipohl:
Thank you also from my side for the nice POD! :-)

You can see the result here in [my instance of the Munin Plugin Gallery](http://munin-gallery.dipohl-kunden.de/contrib/svn/trunk/plugins/network/tc_.html)

Unfortunately the gallery of the Munin project seems to be broken / not uptodate at the time :-(
I will create a ticket for this.

PR #765, opened 2016-11-02
Postfix stats - Rewrited plugin

CristianDeluxe:
Hi, i've followed all your suggestions from previous pull request, unfortunately i can't remove merges and finally i've created another pull request 😐 

Previous pull:
https://github.com/munin-monitoring/contrib/pull/711

Plugin has been completely rewritten:

- No use of TMP files 👍 
- Much faster because "pflogsumm" command has been tuned
- More efficient "parsing" now it use grep and regexp, and is more "failsafe"
- Added autoconf
- Script more documented
- Removed duplicated code as far as possible
- Removed backticks

I hope everything is fine!


plugins/mail/postfix_stats
```diff
@@ -3,50 +3,102 @@
...
+# Autoconf Section
+#
+if [ "$1" = 'autoconf' ]; then
+	# Try to find pflogsumm with default name
+	PFLOG_EXIST=$(command -v pflogsumm.pl)
+
+	# Try to find pflogsumm without any extension
+	if [[ -z "${PFLOG_EXIST}" ]]
+		then
+		PFLOG_EXIST=$(command -v pflogsumm)
```
sumpfralle: You are using path name guessing for determing the location of "pflogsum". This is a good approach.
But you do not use the result of this guessing below in the actual "fetch" branch.

I would propose the following approach:
```
PFLOGSUM="${pflogsum:-}"
[ -z "$PFLOGSUM" ] && PFLOGSUM="$(which pflogsum pflogsum.pl | head -1)"
```

plugins/mail/postfix_stats
```diff
@@ -3,50 +3,102 @@
...
+	fi
+
+	if [[ -z "${PFLOG_EXIST}" ]]
```
sumpfralle: Style hint: finishing the `if` line with `; then` is quite common.

plugins/mail/postfix_stats
```diff
@@ -3,50 +3,102 @@
...
+		then
+		echo 'no';
```
sumpfralle: you can save one character here :)

Additionally you should add the reason for "no" in brackets afterwards. E.g.: `echo "no (failed to find pflogsum)"`

plugins/mail/postfix_stats
```diff
@@ -3,50 +3,102 @@
...
+	else
+		echo 'yes'
+	fi
+	exit 0;
```
sumpfralle: same as above

plugins/mail/postfix_stats
```diff
@@ -3,50 +3,102 @@
...
+if [ "$1" = 'config' ]; then
+	echo 'graph_title Postfix statistics'
+	echo 'graph_vlabel Postfix statistics'
+	echo 'graph_category mail'
+	echo 'graph_scale no'
+	echo 'graph_period minute'
+	echo 'graph_total Total'
+
+	# Generate config for each field
+	for i in "${FIELDS_ARR[@]}"
```
sumpfralle: if you moved away from `declare` above, then you can use `for i in $FIELDS_ARR; do`

plugins/mail/postfix_stats
```diff
@@ -3,50 +3,102 @@
...
+TMP_RAW=$("${PFLOGSUMM}" -d today --detail 0 --zero-fill "${SYS_LOG}" "${SYS_LOG2}")
+
+# Parse value from Raw result
+# 
+# Return digit if regex are parsed correctly
+#
+# Return -1 if any error occurs
+#
+parseValue() {
+	TMP_RETURN=$(echo "${TMP_RAW}" | grep -Ei '^[[:space:]]+[[:digit:]]+[[:space:]]+'"${1}"'.*$' | grep -oEi '[[:digit:]]+[[:space:]]+' | head -n 1 | sed 's: ::g')
```
sumpfralle: You should use `local` for variables in functions.
Style hint: wrapping the first `grep` expression in double-quotes should be more readable.

plugins/mail/postfix_stats
```diff
@@ -3,50 +3,102 @@
...
+	if [[ -z "${TMP_RETURN}" ]]
```
sumpfralle: bashism `[[ .. ]]`

plugins/mail/postfix_stats
```diff
@@ -3,50 +3,102 @@
...
+		then
+		echo -1
```
sumpfralle: probably you want to use `U` ("undefined") instead

plugins/mail/postfix_stats
```diff
@@ -3,50 +3,102 @@
 # Plugin to show Postfix statistics - needs pflogsumm
```
sumpfralle: If you like perfection, you may want to transform the current plugin documentation into the POD format (useful for `munindoc postfix_stats`). See the plugin "processes/multimemory" for an example.

CristianDeluxe:
Thanks for your suggestions and tips 👍  now the script is strictly "SH" and better than before.

I'm not a sh/bash programmer, usually i use it for do some works in my servers, but with this i've learned a lot of stuff!

Thanks again!
Chris

sumpfralle:
Hi Chris,
I am glad, that you enjoy the progress! :)

I have one last (really!) nitpick:
you use `eval` in the pflogsumm call - probably to work around shellcheck's complain about the missing quotes around SYS_LOG (due to the space separated filenames). `eval` works here, but since it can have a variety of side-effects, you should try to avoid it - your fellow readers will appreciate this.
Just use your original line, remove the quotes around SYS_LOG and add a `shellcheck disable=???` comment (see its manpage) in order to silence shellcheck's warning (since you know what you are doing and there is no better way).

CristianDeluxe:
Here we go again! 😄 

All the changes made! Thank you for being so conscientious!

When I have some free time I will continue working on other plugins that can be polished a little more

It has been a pleasure working with you!

regards

sumpfralle:
thank you for your kind words!
Polishing plugins is really welcome!

PR #766, opened 2016-11-04
w1_ plugin

webstoney:
Plugin to graph the output of 1-Wire DS1820 temperature sensors on the Raspberry PI

### DIFF
--- a/plugins/raspberry-pi/w1_
+++ b/plugins/raspberry-pi/w1_
@@ -52,7 +52,7 @@ esac
 
 if [ "$1" = "suggest" ]; then
     if [ -r /sys/bus/w1/devices ]; then
-        ls /sys/bus/w1/devices|head -n -1
+        ls /sys/bus/w1/devices|grep -v bus_master
     fi
     exit 0
 fi
@@ -71,7 +71,7 @@ if [ "$1" = "config" ]; then
         echo graph_title Temperature Sensor $sensor_id
         echo 'graph_args --base 1000 -l 0'
         echo 'graph_vlabel temperature (°C)'
-        echo 'graph_category temperature'
+        echo 'graph_category sensors'
         echo 'graph_info This graph shows 1-wire sensor temperature.'
         echo "w1.label $sensor_id"
         echo "w1.info Temperature at $sensor_id."
@@ -81,4 +81,4 @@ if [ "$1" = "config" ]; then
 fi
 
 
-sed  -n '/t=/ s/.*t=//p' /sys/bus/w1/devices/$sensor_id/w1_slave | perl -e '$l=<STDIN>;print "w1.value ", $l/1000 , "\n";'
+sed  -n '/t=/ s/.*t=//p' /sys/bus/w1/devices/$sensor_id/w1_slave | awk '{print "w1.value", $1/1000}'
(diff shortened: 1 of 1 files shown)


sumpfralle:
Your plugin looks great!

I have just a few questions / hints:
* please pick a graph category fom the current [category list](http://munin-monitoring.org/wiki/graph_category_list)
* please take a look at the output of `shellcheck -s dash w1_` - maybe you want to add some quotes at some places
* your "suggest" code (`ls /sys/bus/w1/devices|head -n -1`) is hard to understand for me. Is there a magic last file that you want to skip? Maybe use a specific `grep -v` filter?
* I think it would be convenient for users if multiple temperatures/devices would be shown in a single graph. Maybe you want to remove the "suggest" capability and just show all devices by default?

Did you already take a look at the [w1-therm plugin](https://github.com/munin-monitoring/contrib/blob/master/plugins/sensors/w1-therm)? You plugin looks like it is similar in scope but better documented and better styled. Maybe you want to replace/rewrite the other plugin instead of adding a new one?

webstoney:
Did the suggested changes. In my setup there are 8 sensors where some of them having almost the same temperature. I singled out the temperatures on purpose btw. 

I haven't seen w1-therm before

sumpfralle:
Did you already push your changes?

Since you prefer a single temperature graph (sounds reasonable), there is no functional difference between your plugin and w1-therm (but yours is more complete). Thus please replace the w1-therm plugin with your code and change your fieldname ("w1") to the fieldname of the "w1-therm" plugin.

Another small detail occoured to me: please replace your invocation of `perl` in order to run this plugin even on embedded platforms where perl is missing. Something like `$((foo / 1000))` (for integers) or `awk '{print $1/1000}'` should work.

sumpfralle:
What do you think?

sumpfralle:
@webstoney: did you already push the changes that you mentioned in your [last comment](766#issuecomment-258424418)?

sumpfralle:
@webstoney: any news?

sumpfralle:
This looks good - thank you!

One last thing:

> [..] please replace the w1-therm plugin with your code and change your fieldname ("w1") to the fieldname of the "w1-therm" plugin.

(you may want to rename your existing `rrd` data files, if you want to keep them)

What do you think?

Maybe you are already tired of my wishes?
In this case I could also also just merge your plugin and move it over to `w1-therm` on my own :)
(but I would prefer to leave these changes to you - for the revision history being more accurate)

webstoney:
I'm currently a bit busy, but otoh if someone wants to keep historical data I think it would be a bit of hassle for them to switch (2 different plugins with the same name though). But I'll look into it.

sumpfralle:
@webstoney: [..] if someone wants to keep historical data [..]

The other plugin used the fieldname `temp`. Thus if you change yours from `w1` to `temp`, then everything is fine.
Please be bold :)

sumpfralle:
Thank you!

sumpfralle:
@webstoney: I converted the plugin into a single plugin for all sensors (in a single graph). This allows to use the plugin without manual (or semi-automatic) configuration.

Could you please take a look, whether it works for you?

See #1011

PR #767, opened 2016-11-07
added responsive template, which shows month and year graphs on wide screens

hansmaulwurf23:
Copied official templates, changed ```munin-nodeview.tmpl``` and ```partial/head.tmpl``` accordingly.


sumpfralle:
Do you think that this could be integrated into the original template?
(Maintaining two copies of the same template would be a burden)

hansmaulwurf23:
Done :-)

sumpfralle:
Great!
If I understand your change correctly, you are hiding the fourth column or the third and fourth columns based on the media width.
Do you think, a comment explaining this would be helpful for your new css section? I am just an occasional css user - I cannot judge if your approach would be a trivial pattern easily understood by most css authors. What do you think?
Besides this, I think this can be merged.

hansmaulwurf23:
Yes, you understand correctly. Nevertheless, I added some comments to the css.

sumpfralle:
Thank you!

hansmaulwurf23:
One last thing: if anyone configures `graph_width` in the `munin.conf` the absolute values of the media queries will fail obviously (or look ugly). Since this setting can be used for the munin instance, a set of nodes only or even one specific graph, a calculated setting for the media queries is hard to achieve (though possible via inline css and the css3 calc()-function). 

A rather simple workaround could be making this responsiveness configurable (to add the month and year graphs only if the option is set). The docs (or a comment in the `munin.conf`) could state that the responsiveness option (when set) should only be used if munin is configured to use the default settings for `graph_width`.

sumpfralle:
Thank you for thinking further!

If I understand the problem correctly, then a smaller `graph_width` would lead to unused horizontal space (one or two missing graphs). A bigger `graph_width` would enforce a wider horizontal visualization (i.e. a horizontal scrollbar).
From my point of view both situations are acceptable. Thus I would think, that we do not need another configuration setting.

btw.: is this template the one used by munin by default? If yes, then a comment next to `graph_width` in munin.conf could be reasonable (needs another pull request for the other repository). Otherwise I would tend to leave it like it is.

What do you think?

hansmaulwurf23:
Yes the template is used in munin by default. Imho both scenarios are acceptable since it is an aesthetics issue ;-) (horizontal scrollbars and wasted horizontal space is basically what the template had before).

I cannot find the default values in the other repo for the munin core. https://github.com/munin-monitoring/munin/blob/master/etc/munin.conf.PL does not hold them. Where should this comment be added?

sumpfralle:
I found a reference to this option here:
https://github.com/munin-monitoring/munin/blob/master/doc/reference/munin.conf.rst

But personally I tend to think, that an imperfect usage of screen estate is an acceptable situation for people using modified graph sizes.
Of course, you may feel free to open a  pull request against the main munin repository. I do now know, how the maintainers of the main munin repository would think about this.

hansmaulwurf23:
I agree, most users probably use the default graph dimensions and for the ones with this settings changed it is really simple to clone the template and setup other media query thresholds accordingly.

PR #768, opened 2016-11-07
Plugins for monitoring EMC VNX5300 NFS Stats

Vshmuk:
These plugins monitor NFS statistics of EMC Unified Storage system. Probably they can also be compatible with other Isilon or Celerra systems. They use SSH to connect to Control Stations, then remotely execute /nas/sbin/server_stats and fetch and parse data from it. They support gathering data both from active/active and active/passive Datamover configurations, ignoring offline or standby Datamovers. If all Datamovers are offline or absent, the plugin returns error.
These plugins also automatically choose Primary Control Station from the list by calling /nasmcd/sbin/getreason and /nasmcd/sbin/t2slot.


Vshmuk:
Not a problem, I've deleted these lines. Anyway, I grepped with "donat" through the Repository, found some messages, so i decided that it's allowed here.

sumpfralle:
Hello Vshmuk,

thank your for pointing out the existence of donation requests in the plugin repository. I started a [mailinglist discussion](https://sourceforge.net/p/munin/mailman/munin-users/thread/20161127191104.753c0bc4%40erker.lan/#msg35517214) regarding the removal of these donation requests. Probably this inconsistency will be fixed soon.

Please take a look at your other "emc"-related pull request (#775). Most of my comments over there are relevant for this plugin, as well.

Since all three plugins of yours seem to share a good amount of common code, I would recommend to merge these plugins, if possible. Otherwise they will tend to become a maintenance burden - especially since these plugins cannot be tested without the specific proprietary ("navicli") software. What do you think?
Thank you!

Vshmuk:
Thank you for your comments, I will fix and push my code soon.
My knowledge of Github is not good enough, so how do we merge these two pull-requests? Do I have to prepare a new one, closing these two?

sumpfralle:
It would be nice for preserving the history of this conversation, if you continue using the existing ones instead of creating a new one. If you can merge all plugins into a single one (e.g. using the "suggest" method of mun plugins), you will want to close one of the pull requests and just continue the other one.
Just continue committing changes. Within your own pull requests (branches) you may also feel free to rebase (overwriting history) and force-push in order to cleanup the commit history.

PR #770, opened 2016-11-11
other/sshd_log: support logs from journald too

Niluge-KiWi:
other/sshd_log: support logs from journald too
To use journald, set env.logfile to special value 'journald': It will
read the sshd logs from journald --unit sshd.service.
To configure the sshd unit name, use the env.logfile special pattern:
'journald:<unit name>'. Example: 'journald:sshd_custom.service'.

Tested on archlinux which uses journald and debian 8 which uses /var/log/auth.log.

I'm not sure if it's useful to have an option to configure the journald unit name though.


plugins/other/sshd_log
```diff
@@ -28,18 +29,35 @@
...
 CATEGORY=${category:-system}
+JOURNALD_UNIT=${journald_unit:-sshd.service}
 
 
 if [ "$1" = "autoconf" ]; then
-        if [ -r "$LOG" ]; then
-                echo yes
-                exit 0
+        if [ $LOG == "journald" ]; then
+                if journalctl -F _SYSTEMD_UNIT | grep $JOURNALD_UNIT -q; then
```
sumpfralle: the manpage of my local grep wants the options before the pattern - please swap them

plugins/other/sshd_log
```diff
@@ -28,18 +29,35 @@
...
 LOG=${logfile:-/var/log/secure}
 CATEGORY=${category:-system}
+JOURNALD_UNIT=${journald_unit:-sshd.service}
 
 
 if [ "$1" = "autoconf" ]; then
-        if [ -r "$LOG" ]; then
-                echo yes
-                exit 0
+        if [ $LOG == "journald" ]; then
```
sumpfralle: the test "==" is a bashishm - please replace it with "="
`shellcheck -s dash sshd_log` will give you even more useful hints (e.g. regarding quoting)

plugins/other/sshd_log
```diff
@@ -28,18 +29,35 @@
...
+                        exit 1
+                fi
         else
-                echo no
-                exit 1
+                if [ -r "$LOG" ]; then
+                        echo yes
+                        exit 0
+                else
+                        echo no
```
sumpfralle: There is a convention to add some details regarding the failure in brackets, e.g.: "no ($logfile is missing)"

plugins/other/sshd_log
```diff
@@ -28,18 +29,35 @@
...
+                        exit 1
```
sumpfralle: "autoconf" should always exit with zero - even for "no"

plugins/other/sshd_log
```diff
@@ -28,18 +29,35 @@
 #env.logfile /var/log/messages
 #env.category users
 #
+# config example with journald
+#[sshd_log]
+#group systemd-journal
+#env.logfile journald
+#env.journald_unit sshd.service
```
sumpfralle: what do you think about merging these two settings (logfile + journald_unit) into a single one - e.g. `env.logfile journald:sshd.service`?

Niluge-KiWi: I already hesitated on reusing an env var with a special value: it seems hack-ish on principle, but does work great in practice.
What you propose is in fact a pseudo URL mechanism, we could go full URL, but it's probably overkill. 
BTW journald also has a remote service, on top of HTTP, this doesn't help us.

Let's give it a try !

plugins/other/sshd_log
Niluge-KiWi: done !

plugins/other/sshd_log
Niluge-KiWi: done !

plugins/other/sshd_log
Niluge-KiWi: done !

Niluge-KiWi:
> 1) can you think of a way to allow this plugin to run without manual configuration under common circumstances? E.g. looking for journcalctl if "env.logfile" is empty and trying some logfile names if journalctl is not available? Explicit configuration should still work, but clever autoconfiguration would be very helpful for most users.

Adding multiple log files by default would be easy. I just need a list of usual log files (debian is at /var/log/auth.log; /var/log/secure, the default, seems to be from redhat, others I don't know).

Auto-detection of journald is harder: on my debian 8 I do have journalctl and the sshd.service unit, it's even started. But the log files are still in /var/log/auth.log: the journald logs are empty. I haven't found where in sshd config it says to use rsyslog (debian) or journald (archlinux); it's probably a system-wide log config.
I don't want to dig that much deeper...

Niluge-KiWi:
> 2) It would be nice (but not mandatory), if you could turn the documentation of this plugin into the POD style format (e.g. see the munin_update plugin).

Done !

> 3) For a real shiny plugin within the plugin gallery it would be great, if you could add some example graphs (see Plugin-Gallery documentation).

Done !
I added 2 images although the documentation you linked talks about only one. Should I remove one ?

sumpfralle:
Your code and documentation changes look great!

The two images are surely fine (at least I always understood the documentation in this way).

Thank you for considering the autodetection (log/journald). It would be nice, if the world would be perfect :)

I have one last wish, that did not occur to me before:
Could you please remove the (already previously) configurable "category" selection? Variables make life harder for the Plugin Gallery parser (see [category parsing](http://munin-monitoring.org/wiki/PluginGallery#Categories)). The default value ("system") should be a good choice and needs no configuration, I think. Thank you!

Niluge-KiWi:
> Could you please remove the (already previously) configurable "category" selection? Variables make life harder for the Plugin Gallery parser (see category parsing). The default value ("system") should be a good choice and needs no configuration, I think. Thank you!

It seemed odd to me too, but I tried to be conservative and not break anything.
Let's break things !
Done.

Niluge-KiWi:
What about travis-ci btw ? File::Tail::Multi was recently removed from cpan ?

Niluge-KiWi:
After more testing I also found another way to get the sshd logs from journald: `journalctl _COMM=sshd`
* It does work on debian 8, although it has ~logrotate on different dates: I have just 2 days here, versus 6 days in /var/log/auth.log.
* It also works on archlinux, where I have much more history (5 months old !) than with `journalctl --unit sshd.service`. it becomes a performance issue...
I don't know how journald rotates log, and how it could keep more history on archlinux than with the service directly.
Maybe sshd was enabled without the service ?
Yes it is that:
https://wiki.archlinux.org/index.php/Secure_Shell#Daemon_management
I started sshd service only yesterday.
Previously it worked by socket activation, and I can't find a way to get the logs from there, except with `_COMM=sshd`.

I have to rethink my whole PR...

For the performance issue, the solution would be to limit by size or date (journalctl is good at that). But how much would make sense ? One week would emulate the default logrotate of debian.

sumpfralle:
Regarding your thoughts about different data sources: how are the current values related to time? Do they grow over one week an get reset during logrotate? Your graphs indicate this.
How about switching from GAUGE to DERIVE with "min=0" (see [datatypes](http://munin.readthedocs.io/en/latest/architecture/syntax.html#datatypes))? This should work well with logrotate-like behaviour, but could cause problems with "fixed timespan" (if this is what journald allows). Maybe journalctl supports something like "from the start of this day" (or can be tricked into).

sumpfralle:
Anyway I can also merge your current request, if you do not know if/when you want to approach further improvements?

sumpfralle:
travis: yes, a module got lost on cpan - we will need to work around this. Or maybe someone removes this dependency from [apache/apache_vhosts/apache_logparser](https://github.com/munin-monitoring/contrib/blob/master/plugins/apache/apache_vhosts/apache_logparser) ...

Niluge-KiWi:
I'll look into datatypes and the journalctl man page.
If you merge this now I'll probably have to break the options again to support more options for journalctl (I'm thinking of maybe giving full control of the arguments).

Niluge-KiWi:
> Regarding your thoughts about different data sources: how are the current values related to time? Do they grow over one week an get reset during logrotate? Your graphs indicate this.
That's the historical behavior of this plugin yes: it counts events from the auth.log file, and just returns them.
Then logrotates happens and the counter are reset.
DERIVE would make more sense indeed !

However, for journald emulating logrotate is tricky, it's easier to have a rolling time window ('last seven days': `-7days`) than to compute the logrotate date.
That being said, I have found another journalctl feature: they have cursors to get the new logs since last time.
It would work great with ABSOLUTE, but I need a way to store the cursor between the executions.
What is the recommended way for that ?

sumpfralle:
Storing cursors: there is an environment variable "MUNIN_PLUGSTATE". It refers to a directory to be used for storing states between plugin executions. By convention a plugin may create files with its name as a suffix.

Regarding states: please note that a good plugin should not get confused if it delivers data to more than one munin server (e.g. three consecutive calls in every five-minute period). This can make the usage of state files a bit challenging.

How about using the following timestamp: `$(date +%Y%m%d0000)`? This would be today's midnight.
Following a fictional logrotate period is not a worthwhile goal, I guess.

Niluge-KiWi:
According to http://guide.munin-monitoring.org/en/latest/plugin/env.html `MUNIN_PLUGSTATE` is when the plugin wants to share data with other plugins.
`MUNIN_STATEFILE` is for the plugin itself, and is guaranteed unique per plugin + master, which fixes the multi master issue you described.

So I could go on with this solution, hopefully it should have the same visible behavior as with the historical file-based mode. 
It would be more robust too: it would not miss the log entries that were saved during the last 5min window before logrotate, and after last polling.

Otherwise, I could follow your idea using `journalctl --since $(date +%Y%m%d0000)` (which could be simplified into `journalctl --since today`) but it would be equivalent to a daily logrotate, if we want a weekly rotate we could do `journalctl --since $(date -dlast-sunday +%Y%m%d0000)` (although I don't know if this work with older `date` versions).

I'll try one or the other.

sumpfralle:
Thank you for clearing up my understanding of the state files!

Just a note regarding `--today` combined with "DERIVE": IMO this is a perfect solution: the graphs will show the number of new events per second instead of the the currently shown growing number of accumulated events.
The same can probably be done for the non-journald case by just using the single latest logfile (with DERIVE).
Anyway: have fun!

Niluge-KiWi:
I'm OK with changing the datatype to DERIVE, let's call that v2.
For file-based it already uses the latest logfile.

I have a working version with cursors and datatype ABSOLUTE:
* pros
 * doesn't loose data on the last minutes before midnight
 * it's faster:we only read the new logs since last polling
* cons
 * the code is more complicated
 * on the first run it will read the whole journal, which can be slow (<1min)

The version with `--today` is trivial to implement.

I don't know which one is the best. I'll push the cursor based version on a new PR.

Niluge-KiWi:
v2 at #773

PR #771, opened 2016-11-12
Add source as first 'hop'

leeclemens:
The value of the first hop is undefined, but includes the source's hostname for comparison with other graphs generated for traceroutes to the same target.


sumpfralle:
This looks good to me.
Could you please add a comment for the awk statements? Currently it is hard to guess its purpose from the code.

sumpfralle:
fine!

PR #773, opened 2016-11-13
other/sshd_log: use DERIVE datatype and support logs from journald too (using cursors)

Niluge-KiWi:
Now shows the number of events per time unit, instead of a counter
always increasing until logrotate.

To use journald, set env.logfile to special value 'journald': It will
read the sshd logs from journalctl _COMM=sshd.
To configure the source of journald, set env.journalctlarg:
Example: "env.journalctlarg --unit=sshd.service"

Previous discussion: #770.
This version uses journald cursor to only parse new logs on each run.

### DIFF
--- a/plugins/ssh/sshd_log
+++ b/plugins/ssh/sshd_log
@@ -23,6 +23,9 @@ The following environment variables are used by this plugin:
                   journalctl to get the sshd logs.
                   default: _COMM=sshd
 
+ type - "GAUGE" or "DERIVE"
+         default: GAUGE
+
 If the "logfile" environment variable is set to "journald" the sshd
 logs are read from journald, filtering on program "sshd". The filtering
 may be changed using "journalctlargs".
@@ -48,6 +51,13 @@ Config example with journald on the sshd.service unit only:
       env.logfile journald
       env.journalctlargs --unit=sshd.service
 
+Config example with journald and type DERIVE:
+
+  [sshd_log]
+      group systemd-journal
+      env.logfile journald
+      env.type DERIVE
+
 =head1 MAGIC MARKERS
 
   #%# family=auto
@@ -71,6 +81,10 @@ Revision 1.0  2009/04/22 22:00:00  zlati
 
 LOG=${logfile:-/var/log/secure}
 JOURNALCTL_ARGS=${journalctlargs:-_COMM=sshd}
+TYPE=${type:-GAUGE}
+if [ "$LOG" = "journald" -a "$TYPE" = "DERIVE" ]; then
+        TYPE=ABSOLUTE
+fi
 
 
 if [ "$1" = "autoconf" ]; then
@@ -92,13 +106,6 @@ if [ "$1" = "autoconf" ]; then
 fi
 
 if [ "$1" = "config" ]; then
-
-        if [ "$LOG" = "journald" ]; then
-                TYPE=ABSOLUTE
-        else
-                TYPE=DERIVE
-        fi
-
         echo 'graph_title SSHD login stats from' "$LOG"
         echo 'graph_args --base 1000 -l 0'
         echo 'graph_vlabel logins'
@@ -139,7 +146,7 @@ if [ "$1" = "config" ]; then
         exit 0
 fi
 
-if [ "$LOG" = "journald" ]; then
+if [ "$LOG" = "journald" -a "$TYPE" = "ABSOLUTE" ]; then
         CURSOR_FILE="$MUNIN_STATEFILE"
         # read cursor
         # format: "journald-cursor <cursor>"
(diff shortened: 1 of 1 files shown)


Niluge-KiWi:
The png examples are wrong: they are from the old version with GAUGE instead of DERIVE/ABSOLUTE...

autoconf could be improved now that we read from _COMM and use cursors (no data lost even on debian where journald rotate happens on different dates than /var/log/auth.log).

sumpfralle:
Your changes look good. I have just a few comments:
* maybe you want to rename "env.journalctlarg" to "env.journalctlargs" and remove the double quotes from the lines 74 and 150? This would allow multiple arguments. A "disable" comment for shellcheck would work around the quoting warning.
* Does the code work even if the journald cursor file is missing?

Personally, I would still prefer DERIVE for both log sources, since currently it is not possible to change the log source (journald/file) without losing the history (or at least: losing the visual representation). But this design decision is obviously up to you.

Now I would wait for another fresh set of png graphs (maybe just the daily one) from you, before merging.

Niluge-KiWi:
> * maybe you want to rename "env.journalctlarg" to "env.journalctlargs" and remove the double quotes from the lines 74 and 150? This would allow multiple arguments. A "disable" comment for shellcheck would work around the quoting warning.

I thought about doing that, but then if the user wants to use space in a parameter it will not be able to. Whereas now it can only pass one parameter, but we support spaces.
Maybe you are right and multiple non-space parameters is better than just one with spaces.

> Does the code work even if the journald cursor file is missing?

Yes, it reads the whole journal (which may be big, but it's not that slow in my case)


> Personally, I would still prefer DERIVE for both log sources, since currently it is not possible to change the log source (journald/file) without losing the history (or at least: losing the visual representation). But this design decision is obviously up to you.
Good point, I hadn't thought about it.

After one week of usage, for low traffic servers it appears that the DERIVE/ABSOLUTE datatype is less appealing than the previous GAUGE (maybe it's because I was used to it, but now we only see one peak per login, the graph is discontinuous... not great.)

I'll try to add an option to choose between the two types.

To support GAUGE with journald I'll have to go back to non-incremental log read, ie no more cursor.

This also means no more ABSOLUTE, which is what you suggested too.

sumpfralle:
> I thought about doing that, but then if the user wants to use space in a parameter it will not be able to. Whereas now it can only pass one parameter, but we support spaces.
> Maybe you are right and multiple non-space parameters is better than just one with spaces.

Yes, whitespaces are problematic without quotes - I do not know a good approach for this problem. Since you are more familiar with the variety of useful arguments for journalctl, you will pick the right approach on your own ...

> After one week of usage, for low traffic servers it appears that the DERIVE/ABSOLUTE datatype is less appealing than the previous GAUGE (maybe it's because I was used to it, but now we only see one peak per login, the graph is discontinuous... not great.)

I am confused: shouldn't you see a steady trickle of new login attempts coming in? Personally I perceive regularly reset GAUGE graphs as rather hard to read and would prefer DERIVE.
Anyway: you will pick something suitable ...

> To support GAUGE with journald I'll have to go back to non-incremental log read, ie no more cursor.

I understood, that "--today" would solve this nicely? (just losing every last dataset before midnight)

> This also means no more ABSOLUTE, which is what you suggested too.

I fail to remember that :)

However: the plugin is much better than before. Feel free to follow your judgement regarding the remaining issues and questions. Have fun!

sumpfralle:
@Niluge-KiWi: ping?

Niluge-KiWi:
I may be able to work on it this weekend.

sumpfralle:
@Niluge-KiWi: any news?

Niluge-KiWi:
I have not worked on this since my last message here. I can't promise
anything; if anyone wants to continue working on this, feel free.

sumpfralle:
> I have not worked on this since my last message here. I can't promise anything; if anyone wants to continue working on this, feel free.

In my experience, people tend to follow the not-invented-here syndrom and will ignore existing pull request, even if they are almost done :(

Please take a look at my last comment, if/when you find the time. I think, *merge as it is* would also be an option, if you think so :)

Niluge-KiWi:
@sumpfralle here we go!

I went for your multi args proposition, and also added an option to choose between GAUGE and DERIVE types:
- for journald GAUGE emulates weekly logrotate: `--since=$(date -dlast-sunday +%Y-%m-%d)` instead of just one day with `--since=yesterday`: more data for low traffic nodes.
- choose to default to GAUGE type as that's what I use on my low traffic nodes.
Feel free to change the default if you think it's best to have DERIVE by default.

The screenshots reflects the default: `GAUGE` + `/var/log/auth.log`.

I believe it's thus ready to be merged, finally!

sumpfralle:
Great!
Thank you for persisting :)

btw: you mentioned screenshots. Did you maybe forgot to add example graphs?
(they would be a nice addition)

sumpfralle:
Now I committed your example graphs from #770. In case you want to upload new ones (without the zigzag?), please just add links here or open another pull request. Thank you!

PR #776, opened 2016-11-22
Added plugins for Oracle Database.

shakemid:
I wrote some plugins for Oracle Database. Although Oracle Enterprise Manager is used in many cases, I feel Munin useful to grasp trends well.
These plugins are written in Bash script and only need sqlplus command. I tested them with Oracle 12c (and 11g). 
Sample graph images are available in my site.
https://github.com/shakemid/munin-plugins-oracle
I added POD document, sample config, magic markers. Is it okay?
Please consider this pull request.


sumpfralle:
Hello shakemid,

thank you for submitting your plugins!

The plugins look well-written.

Maybe you want to consider some remarks:
* please pick a category from the [plugin category list](http://munin-monitoring.org/wiki/graph_category_list)
* verify if you could add some quotes (e.g. around "oracle_auth" in `sqlplus -s ${oracle_auth}`) - see `shellcheck ...` for more hints
* add a reason for "no" in "autoconf" - e.g. `echo "no (failed to find executable 'sqlplus')"`
* the definition of keys, labels, types, and drawing styles in "oracle_physicaliops" could be simplified by combining these into a single line for each field (e.g. "KEY TYPE DRAW LABEL...."), since the first three items do not contain whitespace. Maybe you would prefer this dense style.

All of your plugins share a good amount of common code.
Due to the number of plugins and the amount of duplicated code I would strongly suggest, that you try to merge these scripts into a single one - probably including a "suggest" mechanisms. Otherwise the sheer number of plugins will turn future changes of `sqlplus` calling conventions or similar changes into a maintenance nightmare. This is even more important for plugins like these, that require non-free software and thus can only be tested by a limited number of contributors/maintainers.

What do you think?

shakemid:
Hello sumpfralle,

Thank you very much for your review!

> please pick a category from the plugin category list"

"oracle" looks good to me.

>     verify if you could add some quotes (e.g. around "oracle_auth" in sqlplus -s ${oracle_auth}) - see shellcheck ... for more hints

I see. I will add quotes.

>     add a reason for "no" in "autoconf" - e.g. echo "no (failed to find executable 'sqlplus')"

All right. I will add a reason.

>     the definition of keys, labels, types, and drawing styles in "oracle_physicaliops" could be simplified by combining these into a single line for each field (e.g. "KEY TYPE DRAW LABEL...."), since the first three items do not contain whitespace. Maybe you would prefer this dense style.

Thank you for your advice. I imagined to change style as following. Does it look good?

```sh
...
# Graph settings
#   field type draw "label"
graph_settings=(
    'physical_read_total DERIVE LINE2 "physical read total IO requests"'
    'physical_read DERIVE LINE "physical read IO requests"'
    'physical_read_total_multi DERIVE LINE "physical read total multi block requests"'
    'physical_write_total DERIVE LINE2 "physical write total IO requests"'
    'physical_write DERIVE LINE "physical write IO requests"'
    'physical_write_total_multi DERIVE LINE "physical write total multi block requests"'
)
...

config() {
    echo graph_title Oracle Physical I/O Requests
    echo graph_category Oracle
    echo graph_args --base 1000 --lower-limit 0 --rigid
    echo graph_vlabel iops
    echo graph_info Oracle Physical I/O Requests

    local graph_setting t fields field type draw label
    for graph_setting in "${graph_settings[@]}"
    do
        eval t=( "${graph_setting}" )
        field="${t[0]}"
        fields+=( $field )
        type="${t[1]}"
        draw="${t[2]}"
        label="${t[3]}"

        echo ${field}.type  ${type}
        echo ${field}.draw  ${draw}
        echo ${field}.label ${label}
    done

    echo graph_order ${fields[@]}

    exit 0
}

...
```

> All of your plugins share a good amount of common code.
> Due to the number of plugins and the amount of duplicated code I would strongly suggest, that you try to merge these scripts into a single one - probably including a "suggest" mechanisms. Otherwise the sheer number of plugins will turn future changes of sqlplus calling conventions or similar changes into a maintenance nightmare. This is even more important for plugins like these, that require non-free software and thus can only be tested by a limited number of contributors/maintainers.

That makes sense. I think I can marge 9 plugins (oracle_cachehit, oracle_cursor, ...) using v$sysstat into single plugin like oracle_sysstat_* , and 3 plugins (oracle_event_wait_application, ...) using v$system_event into oracle_event_wait_*.
I will try it and make a pull request again!

sumpfralle:
>> please pick a category from the plugin category list"

> "oracle" looks good to me

The [graph category list](http://munin-monitoring.org/wiki/graph_category_list) recommends to use "db" for databases (e.g. oracle).

Regarding your one-line descriptions (e.g.: ` 'physical_read_total DERIVE LINE2 "physical read total IO requests"'`): I think, you can remove the double quotes around the label (at least I am not aware of the specific bash feature that would use it).
I usually prefer the following style:
```
graph_settings="physical_read_total DERIVE LINE2 physical read total IO requests
                physical_read DERIVE LINE physical read IO requests"

echo "$graph_settings" | while read field data_type draw_style label; do
    echo "$field.label $label"
    echo "$field.type $data_type"
    echo "$field.draw $draw_style"
```
It has the additional bonus of working with all shells (by not using bash-specific features).

Anyway: follow your taste :)

I am happy to hear that you can merge most of these plugins into fewer files!

Btw.: did you already take a look at the existing plugins for oracle in this repository? Maybe you can improve/replace some of them, instead of creating new ones. This will ease the long-term maintenance.

shakemid:
> The graph category list recommends to use "db" for databases (e.g. oracle).

Sure. No problem.

> Regarding your one-line descriptions (e.g.: 'physical_read_total DERIVE LINE2 "physical read total IO requests"'): I think, you can remove the double quotes around the label (at least I am not aware of the specific bash feature that would use it).
> I usually prefer the following style:

Ah, I see. But sometimes label contains letters like '()' required escape in
shell. Because I would like to use labels in SQL (e.g. SELECT value FROM
v$sysstat WHERE name = 'parse count (total)'; )

> It has the additional bonus of working with all shells (by not using bash-specific features).
> Anyway: follow your taste :)

I agree the bonus quite important. Almost always I try not to use
bash-specific features. In this case, I don't believe there is a
platform running Oracle Database without bash. And, I found arrays and process
substitution quite useful.
Thank you for your understanding ;-)

> Btw.: did you already take a look at the existing plugins for oracle in this repository? Maybe you can improve/replace some of them, instead of creating new ones. This will ease the long-term maintenance.

Yes, I have investigated the existing contrib plugins for oracle at first.
But, nothing could work in my environment because most of them are quite old,
some required perl+DBD::Oracle, ruby... That because I decided to
develop new plugins only require bash and sqlplus. 

oracle-pga-monitor, oracle-sga, oracle__connections,
oracle__database_hitratio, oracle__tablespace_usage in contrib seems able to be replaced
with my plugins. 
In my case, bash+sqlplus seems better because I had difficulty to install
DBD::Oracle to our database server for some reason. But, I think that depends
on case, there may be cases perl+DBD::Oracle to be better.
I wonder if I can replace perl and ruby plugins with shell plugins?

shakemid:
Hello sumpfralle,
I tried to merge 11 plugins into single plugin. But, I have not tested yet. I think I can merge all 18 plugins into one script. Although the script became quite large, I don't feel it so complicated.
How does it look? 
https://github.com/shakemid/munin-plugins-oracle/blob/for_pull_request_20161123/oracle_sysstat_
Kind regards,

sumpfralle:
Hi shakemid,

it is great, that you managed to merge all these plugins into one! I like that!

I have a few small comments:
* AFAIK you should not use the `/usr/bin/env` shebang for shells (just for other intepreter that are likely to be configured by the user)
* you used `$(cat <<EOF ...` for storing the different configs and data fields. Wouldn't a simple multiline quote be sufficient?
```
data_attrs[$key]="physical_read_total        DERIVE LINE2 'physical read total IO requests'
                  physical_read              DERIVE LINE  'physical read IO requests'"
```
* you used `eval` in some places. This function can have a multitude of side-effects. Thus it would be a great help for your fellow co-maintainers of that plugin, if you could either remove `eval` completely or at least comment, which effect exactly you are looking for. Otherwise people will likely never touch these lines, since they cannot foresee their purpose. Lines like `eval ${getfield_func[$module]}` should be trivially to be replaced with `${getfield_func[$module]}`, or?
* you used a similar set of `set` commands for sqlplus - maybe you want to join these into a single variable for code deduplication?
* documentation: please change the example configuration section from `[oracle_*]` to `[oracle_sysstat_]` (I like this name-change)

shakemid:
Hello sumpfralle,

Thank you for your review! I will merge these plugins.

> AFAIK you should not use the /usr/bin/env shebang for shells (just for other intepreter that are likely to be configured by the user)

All right, I will use /bin/bash. 

>     you used $(cat <<EOF ... for storing the different configs and data fields. Wouldn't a simple multiline quote be sufficient?

Yes, simple quote looks enough. 

>     you used eval in some places. This function can have a multitude of side-effects. Thus it would be a great help for your fellow co-maintainers of that plugin, if you could either remove eval completely or at least comment, which effect exactly you are looking for. Otherwise people will likely never touch these lines, since they cannot foresee their purpose. Lines like eval ${getfield_func[$module]} should be trivially to be replaced with ${getfield_func[$module]}, or?

Thanks. eval is not necessary. 
Actually I had tried to concatinate variables by using eval before. ( eg. eval $getfield_func_$module )
But, I found using hash to be smart in bash. 

>     you used a similar set of set commands for sqlplus - maybe you want to join these into a single variable for code deduplication?

That makes sence. I will cut sqlplus system variables off like following.

```
sqlplus_variables="\
     set pagesize 0
     set feed off
...
```

>     documentation: please change the example configuration section from [oracle_*] to [oracle_sysstat_] (I like this name-change)

I see. I will fix it.

sumpfralle:
Hello shakemid,

I like all of your changes! Feel free to update this pull request, so I can merge it.

And maybe you are in the mood for a last nitpick: you could remove the single whitespace, that you left at the end of all multiline variables (e.g. `graph_info Oracle Execute Count "`). But this is just a thing for the taste ...

shakemid:
Thank you! I will do tests for a few days in my production environment, and will commit again.

> And maybe you are in the mood for a last nitpick: you could remove the single whitespace,

I agree. Your advice is very helpful!
I felt sequence of single-quote and double-quote '" to be hard to see. Also, I feel to make a new line as following to be better for readability except for unnecessary last blank line.
```
multi_line_variable="\
     foo1 bar1 'value contains white spaces'
     foo2 bar2 'value contains white spaces'
"
```
And also, indents in multi-line variable are unnecessary but I feel it good for readability. How do you feel?

sumpfralle:
The readability of the adjacent quotes is indeed not perfect, but it should be acceptable from my point of view.
Regarding the multiline variables with the trailing newline: personally I would pick this approach (then you could also get rid of the escaping backslash at the start). But then you will need to filter/skip empty lines when processing these multiline variables. This would be the approach for my taste. Pick yours! :)
Indentation should be fine.

It is good to hear, that you will tests these plugins for some days!
While you are at it: maybe you want to add one or two graph images (just pick a random graph from these 18 sets of data) for the [gallery](http://munin-monitoring.org/wiki/PluginGallery)?

steveschnepp:
> In my case, bash+sqlplus seems better 
> I wonder if I can replace perl and ruby plugins with shell plugins?

Shell plugins are usually preferred due to their installation simplicity. Moving to more complete languages such as Perl or ruby (or python or ...), is sometimes useful for performance reasons. But it usually means that the plugin does too many things 😄 

Specially in this particular case where requiring `DBD::Oracle` only for munin plugins seems overkill since it is usually difficult to install.

So, you have my 👍 on that.

shakemid:
Hello sumpfralle,

> Regarding the multiline variables with the trailing newline: personally I would pick this approach (then you could also get rid of the escaping backslash at the start). 

I agree with your opinion. That looks better for future maintenance to add new variables:)

> While you are at it: maybe you want to add one or two graph images (just pick a random graph from these 18 sets of data) for the gallery?

Thank you for your suggestion! I will make a commit with one or two images in example-graphs.

shakemid:
Hello  steveschnepp.

Thank you for your agree. I thought so exactly. Actually developing with perl+DBD::Oracle sounds easier for me than bash+sqlplus. But, as you said, sometimes DBD::Oracle installation looks quite hard, and looks too much only for munin plugins.

shakemid:
I made a commit including single script and 2 images. Please check it.

shakemid:
Sorry, I forgot git add oracle_sysstat_ . Please refer bb2a275 .

sumpfralle:
This looks good now!
Just the naming of the graph images is not correct. You need to name them "oracle_sysstat_-day.png" and so on. Additionally "oracle_sysstat_-1.png" (up to "9") is allowed. The "suggest" suffix is sadly not supported by the munin plugin gallery renderer.

And now comes the formal stuff :)

Please discard your previous commit regarding multiple plugins and the "merge master" commit. `git rebase -i cec2136c57eb~` should be a good start. Afterwards run `git rebase master`. Forcefully push your changes when you are finished (rewriting the commit history of your own pull request branch is acceptable).
Take a look here for details: https://help.github.com/articles/about-git-rebase/

If you feel unsure about rebasing: just create a new pull request with a single perfect commit :)

shakemid:
Ah, thanks! I will try it.

sumpfralle:
Just as usual: create a new branch (`git checkout -b FOO`) in order to work in your private safe playground for experiments.
Have fun!

shakemid:
Thank you so much. Excuse me, I am not so familiar with git and github. Your advice is very helpful!
I tried rebase all changes into one commit  7c08e6c . Does it look good?

steveschnepp:
> I tried rebase all changes into one commit

Well, the _1 commit_ isn't that important, I actually even prefer to have many individual commits, as it tells the whole story of the code. I usually use it as an extra memory when I was wondering why I did something in a particular way.

But I agree that a PR should not be polluted by extra merging commits, hence a rebase.

The commit 7c08e6c looks okay to me, but I let @sumpfralle decide, as I didn't follow the whole debate as I didn't dwell into the code itself.

sumpfralle:
@steveschnepp: I fully agree.

@shakemid: It looks good!
After taking a final look, I noticed, that there still six occurrences of "eval" left. I think, you can simply remove all of them. Please add a comment regarding its purpose, if eval should be really necessary.
Then we are done, I guess ...

shakemid:
@steveschnepp All right! Thank you for your advice!

@sumpfralle Ah, sorry, I forgot to care evals. I will try to remove them.

shakemid:
Hi @sumpfralle 

I think I can remove evals in getvalue(), getfield() to call getvalue_* and getfield_*

Regarding evals to split line into field, type, draw, label in config() and so on, I tried to test several ways as following. (Excuse me for long post)

```bash
#!/bin/bash

# test_split.sh

line=" v1 v2  'v3   with   whitespace'  'v4 with * wildcard'"
line_csv="v1,v2,v3   with   whitespace,v4 with * wildcard"

echo ----------------
echo case 1: using array
t=($line)
declare -p t

## result:
# declare -a t='([0]="v1" [1]="v2" [2]="'\''v3" [3]="with" [4]="whitespace'\''" [5]="'\''v4" [6]="with" [7]="test_split.sh" [8]="wildcard'\''")'

# -> Looks bad. splitted at whitespace in variable, globbing *

echo ----------------
echo case 2: using array with eval
eval 't=('"$line"')'
declare -p t

## result:
# declare -a t='([0]="v1" [1]="v2" [2]="v3   with   whitespace" [3]="v4 with * wildcard")'

# -> Looks good. preserving whitespaces, no globbing

echo ----------------
echo case 3: using read
read -r v1 v2 v3 v4 <<< "$line"
declare -p v1 v2 v3 v4

## result:
# declare -- v1="v1"
# declare -- v2="v2"
# declare -- v3="'v3"
# declare -- v4="with   whitespace'  'v4 with * wildcard'"

# -> Looks bad. splitted at whitespace in variable

echo ----------------
echo case 4: using read -a
read -r -a t <<< "$line"
declare -p t

## result:
# declare -a t='([0]="v1" [1]="v2" [2]="'\''v3" [3]="with" [4]="whitespace'\''" [5]="'\''v4" [6]="with" [7]="*" [8]="wildcard'\''")'

# -> Looks bad. splitted at whitespace in variable

echo ----------------
echo case 5: using set
set -- $line
v1=$1
v2=$2
v3=$3
v4=$4
declare -p v1 v2 v3 v4

## result:
# declare -- v1="v1"
# declare -- v2="v2"
# declare -- v3="'v3"
# declare -- v4="with"

# -> Looks bad. splitted at whitespace in variable

echo ----------------
echo case 6: using read with IFS=,
IFS=, read -r v1 v2 v3 v4 <<< "$line_csv"
declare -p v1 v2 v3 v4

## result:
# declare -- v1="v1"
# declare -- v2="v2"
# declare -- v3="v3   with   whitespace"
# declare -- v4="v4 with * wildcard"

# -> Looks not bad. But need to consider to escape "," in variable
```

Actually, Oracle statistics parameters sometimes contain "*". I don't believe they contain whitespaces in a row, but I am not sure. I think it should be necessary preserving whitespaces and no globbing in this case.  I prefer the behavior like perl qw//. Using eval with some comments look good to me at this time.
Do you have any idea? 
Thanks.

shakemid:
> I think I can remove evals in getvalue(), getfield() to call getvalue_* and getfield_*

Sorry, there is a case args for getfield_func contain whitespace. eval looks necessary. 
```bash
getfield_func[$key]='getfield_eventwait2 "User I/O"'
...
eval "getfield_func[$module]"
...
```

sumpfralle:
Hi shakemid,

regarding "eval" for the whitespace/globbing lines - how about the following:
```
data_attrs[$key]="
    parse_count_total DERIVE LINE parse count (total)
    parse_count_hard DERIVE LINE parse count (hard)
"

while IFS= read -r field type draw label; do
        [ -z "$field" ] && continue
        fields+=( $field )
        echo "${field}.type  ${type}"
        echo "${field}.draw  ${draw}"
        echo "${field}.label ${label}"
done <<< "${data_attrs[$module]}"
```

In case the whitespace is confusing for you, maybe the following is better:
```
data_attrs[$key]="
    parse_count_total|DERIVE|LINE|parse count (total)
    parse_count_hard|DERIVE|LINE|parse count (hard)
"

while IFS="|" read -r field type draw label; do
        [ -z "$field" ] && continue
        fields+=( $field )
        echo "${field}.type  ${type}"
        echo "${field}.draw  ${draw}"
        echo "${field}.label ${label}"
done <<< "${data_attrs[$module]}"
```

The above approaches should be quite common and thus easy to read for most shell users.

Regarding the function calls with whitespace:
```
getfield_func[$key]='getfield_eventwait2 User I/O'
read func_name argument <<<"$getfield_func[$module]"
# function call
"$func_name" "$argument"
```
In your case it should not be important whether "argument" is empty or not. Thus you can use the above approach for functions without or with one argument.
What do you think?

shakemid:
Thank you for your suggestion! I would like to choose the first method. In this  case, the variable containing white-space is "label" only. Thus, using "read" looks enough and smart.

shakemid:
I have removed all evals. Please check it.

sumpfralle:
Now I am happy with the code - thank you for your patient work on it!

shakemid:
Thank you so much for approving this pull request! Also, my plugin became smart and tidy thanks to your advice. I am happy, too!

PR #778, opened 2016-12-02
munin2snmp - snmp agent for querying munin-node plugins over snmp

mestia:
Munin plugins are easy to create, plus there are many ready plugins for almost every use case.
This makes munin-node a valuable tool for monitoring also for other systems which speak mostly snmp.
This PR contains snmp agent for exporting munin plugins via snmp.
Not sure if it is useful for the community or if it fits to 3rd party tools of munin contrib repository.


tools/munin2snmp/munin2snmp.pl
```diff
@@ -0,0 +1,205 @@
...
+my %cache = ();         # Cache
+my @cache_oids = ();    # Keys, sorted
+my $cache_updated = 0;
+my $oidbase = ".1.3.6.1.4.1.123456.100.1.1";
+my $delimiter = ' = ';
+
+# initialize
+my %Munin;
+$Munin{PORT} = '4949';
+$Munin{HOST} = 'localhost';
```
steveschnepp: It would be nice to retrieve them from the env variables, but the defaults are quite sane.

tools/munin2snmp/munin2snmp.pl
```diff
@@ -0,0 +1,205 @@
...
+                last;
+            }
+        }
+    }
+    $SOCKET->print("quit\n");
+    $SOCKET->close();
+    return \@collect if @collect;
+}
+
+=head1 NAME
```
steveschnepp: Seems that the `README.pod` is already having this, and therefore it is redundant. 

It would be nice to choose either, and auto-generate the other.

tools/munin2snmp/munin2snmp.pl
mestia: munin2snmp is a separate program. it doesn't have access to munin env variables as it is not started from munin-node. I've added some code to read a config file or command arguments

sumpfralle:
@mestia: the content of README.md and the footer of the plugin is identical, right? I suggest, you could just remove the README in order to reduce our future maintenance burden?

@steveschnepp: do you see other obstacles besides the duplicate documentation?

mestia:
Hi, sorry for the delay. I've removed the readme file and pushed the changes.
Regards, 
Alex

sumpfralle:
From my point of view, everything should be fine now.

I will keep this pull request open for some more days before merging it. This will give @steveschnepp time to comment, if he feels like. I am not familiar with snmp ...

sumpfralle:
Thank you!

PR #780, opened 2016-12-03
add plugin for PUMA monitoring

alexanikanov:


plugins/puma/puma_
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+# Copyright (c) 2016 Alexey Anikanov (alexey.anikanov@gmail.com)
+# License GPLv2
+# This plugin monitors number of workers, total memory used and average memory per process for puma.
```
steveschnepp: It would be great to explain what PUMA is.

alexanikanov:
description added https://github.com/munin-monitoring/contrib/pull/780/files#diff-5b3b92f741ef3bdc309f49b7424fcea6R5

plugins/puma/puma_
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+# Copyright (c) 2016 Alexey Anikanov (alexey.anikanov@gmail.com)
+# License GPLv2
+# This plugin monitors number of workers, total memory used and average memory per process for puma,
+# a ruby web server built for concurrency http://puma.io
+# Here are the symlinks to enable it
+#
+#        ln -s /usr/share/munin/plugins/puma_ /etc/munin/plugins/puma_average
+#        ln -s /usr/share/munin/plugins/puma_ /etc/munin/plugins/puma_memory
+#        ln -s /usr/share/munin/plugins/puma_ /etc/munin/plugins/puma_processes
```
steveschnepp: If it makes sense to always have all the graphics, it is usually much better to have a multigraph plugin, as it's called only once.

plugins/puma/puma_
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
steveschnepp: The only use of a bashism seems to be the array, which could be even integrated into the `awk` that is already spawned (as you already did in the end of the plugin).

Then you could even use a direct `/bin/sh` which is usually lighter (until `sh` is symlinked to `bash`)

plugins/puma/puma_
```diff
@@ -0,0 +1,71 @@
...
+        fi
+
+elif [ "$mode" = "processes" ]; then
+        if [ "$1" = "config" ]; then
+                echo "graph_title puma Processes"
+                echo "graph_vlabel Processes"
+                echo "graph_category puma"
+                echo "processes.label active processes"
+        else
+                echo -n "processes.value "
```
steveschnepp: `echo -n` can be replaced with a `printf` which is much more standard when you don't want to emit a trailing new line char.

That said, your usual use of `eval` is perfectly sane : only use `printf` when a arg-less `echo` isn't enough.

plugins/puma/puma_
```diff
@@ -0,0 +1,71 @@
...
+        echo 'puma_average.draw LINE2'
+        echo 'puma_average.info The average process size for puma'
+        else
+                echo -n "puma_average.value "
+                ps awwwux | egrep "puma.* worker" | grep -v grep | grep -v master | awk '{total_mem = $6 * 1024 + total_mem; total_proc++} END{printf("%d\n", total_mem / total_proc)}'
+                exit 0
+        fi
+
+fi
+exit 0
```
steveschnepp: :+1: on that, as it is usually forgotten

plugins/puma/puma_
```diff
@@ -0,0 +1,71 @@
...
+        echo 'graph_title puma Average Process Size'
+        echo 'graph_args --base 1024 -l 0 '
+        echo 'graph_vlabel Average Process Size'
+        echo 'graph_category puma'
+        echo 'puma_average.label Average Process Size'
+        echo 'puma_average.draw LINE2'
+        echo 'puma_average.info The average process size for puma'
+        else
+                echo -n "puma_average.value "
+                ps awwwux | egrep "puma.* worker" | grep -v grep | grep -v master | awk '{total_mem = $6 * 1024 + total_mem; total_proc++} END{printf("%d\n", total_mem / total_proc)}'
```
steveschnepp: That's better than using a bashism.

That said, it might be even better to use a simpler `for`

alexanikanov:
thanks for your comments. I'll think about multigraph a bit later. The rest I've fixed I guess)

steveschnepp:
Btw, there's no need to wipe out all the commits in the branch (by rebase+squash). 

The journey is as important as the destination, in which is can tell things that are not in the doc, specially for coding conventions :)

PR #782, opened 2016-12-07
Munstrap big updates and fixes

CristianDeluxe:
Hi, I have recently started to use munstrap and I would like to share some changes that (I think) improve it.

Updated Bootstrap and JQuery to latest.
Integrated JQuery for avoid make external connections (before it was using a CDN) this allow to use it in a computer without internet connection.
Use LazySizes for load images "on-demand", it will only load the visible images, this improves page load speed and avoid unnecessary request to the server.
Now when you change the tab url will be auto updated and when page automatically refresh will back to previously selected tab.
All Javascript and CSS has been minimized and optimized (Using UglyJS for JS and YUI-Compressor for CSS) to faster load times.
Moved all scripts to HTML footer for (again) faster load times.
Reworked categories menu CSS for better look.
Sanitize IDs for avoid error with exotic category names like "system:network" from "interfaces" plugin, now it works with any special char.


sumpfralle:
Thank you for working on the design!
Currently no one with the required html/css/js knowledge seems to be around for proper review. I asked on the mailing list, as well.
Thus I would suggest, that I just merge your proposal as it is. Or do you think that there are details to be discussed or improved?
Could you please rebase your changes?

CristianDeluxe:
I think it can be merged "as is", but if someone wants to review it, you will be welcome.

I miss a sumpfralle's forensic code review!! 😄

sumpfralle:
Instead of missing my review, you should enjoy the incredible amount of lifetime you just saved :)

PR #786, opened 2016-12-18
Update oracle_sysstat_ plugin

shakemid:
I updated oracle_sysstat_ plugin. I have tested it in a few days and it looks working fine. Please consider this pullreq.

* Add modules
  logon
  cputime
  blockrw
  redosize
  eventwaitnetwork
  eventwaitadministrative
  eventwaitconfiguration

* Update modules
  cachehit: Changed to calculate hit ratio between munin-node update intervals by CDEF
  sessionwait: Add CPU status, Add graph_info
  table_fetch: Add table_fetch_continued_row field
  pgastat: Add aggregate PGA target parameter field

* Improvement
  Change not to execute sqlplus in for loop in getvalue funcs to improve performance
  Change to add field.min 0 when type is DERIVE
  Add label_max_length in config() to suppress new line in graph image
  Change graph_title for sorting

* Fix styles
  Enable set -o nounset
  Add double quote to suppress shellcheck warning
  Change some quote styles
  Remove meaningless double quote around $( )
  Add whenever sqlerror exit sql.sqlcode

### DIFF
--- a/plugins/oracle/oracle_sysstat_
+++ b/plugins/oracle/oracle_sysstat_
@@ -411,7 +411,15 @@ getvalue_func[$key]=getvalue_cputime
 key=cachehit
 # buf_hitratio = 1 - physical_reads / ( db_block_gets + consistent_gets )
 # lib_hitratio = 1 - reloads / pins
-# dic_hitratio = ( gets - misses ) / gets
+# dict_hitratio = ( gets - misses ) / gets
+field_info=$( for field in buf_physical buf_logical lib_pins lib_reloads dict_gets dict_getmisses
+   do
+       echo "${field}.graph no"
+       echo "${field}.type DERIVE"
+       echo "${field}.min 0"
+       echo "${field}.label ${field}"
+   done
+)
 global_attrs[$key]="
     graph_title Oracle Cache Hit Ratio
     graph_category db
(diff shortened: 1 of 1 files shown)


plugins/oracle/oracle_sysstat_
```diff
@@ -531,7 +678,7 @@ getvalue_sysstat() {
     do
         [ -z "$field" ] && continue
 
-        ${sqlplus} "${oracle_auth}" <<EOF
+        cat <<EOF
```
sumpfralle: Instead of `cat <<EOF` you could also use a simple echo statement. In this case it would not even interfere with the quoting.

plugins/oracle/oracle_sysstat_
```diff
@@ -314,25 +370,73 @@ global_attrs[$key]="
...
+    )
+
+    buf_hitratio.cdef 100,1,buf_physical,buf_logical,/,-,*,FLOOR
+    lib_hitratio.cdef 100,1,lib_reloads,lib_pins,/,-,*,FLOOR
+    dic_hitratio.cdef 100,dic_gets,dic_getmisses,-,dic_gets,/,*,FLOOR
 "
 data_attrs[$key]="
-    buf_hitratio  GAUGE LINE Buffer Cache Hit Ratio
-    lib_hitratio  GAUGE LINE Library Cache Hit Ratio
-    dict_hitratio GAUGE LINE Dictionary Cache Hit Ratio
```
shakemid: That makes sense. I'd like to use dict_*.

plugins/oracle/oracle_sysstat_
```diff
@@ -366,7 +470,7 @@ global_attrs[$key]="
     graph_category db
     graph_args --base 1000 --lower-limit 0 --rigid
     graph_vlabel microseconds
-    graph_info Oracle Wait Events
+    graph_info Oracle Wait Events - It may looks wierd that Y-axis indicates 'microseconds per second'. Although number of times of wait event looks easier to understand, in many cases the number of events does not matter, but wait time become more important to analyze bottle necks.
```
shakemid: Ah, thanks!

plugins/oracle/oracle_sysstat_
```diff
@@ -496,12 +640,15 @@ config() {
     local fields field type draw label
     while read -r field type draw label
     do
-        [ -z "$field" ] && continue
-        fields="${fields} ${field}"
-
-        echo "${field}.type"  "$type"
-        echo "${field}.draw"  "$draw"
-        echo "${field}.label" "$label"
+        [ -z "${field:-}" ] && continue
```
shakemid: It seems my misunderstanding. (:-) is not necessary.

plugins/oracle/oracle_sysstat_
```diff
@@ -496,12 +640,15 @@ config() {
     local fields field type draw label
```
shakemid: I added field= .

plugins/oracle/oracle_sysstat_
```diff
@@ -496,12 +640,15 @@ config() {
...
     while read -r field type draw label
     do
-        [ -z "$field" ] && continue
-        fields="${fields} ${field}"
-
-        echo "${field}.type"  "$type"
-        echo "${field}.draw"  "$draw"
-        echo "${field}.label" "$label"
+        [ -z "${field:-}" ] && continue
+        fields="${fields:-} ${field}"
```
shakemid: All right. I fixed using initialization fields= .

shakemid:
Thank you for your helpful advices! Please check it again. Should I rebase those fixes?

sumpfralle:

Rebasing (for squashing multiple commits) is usually not necessary (or helpful) for future readers of the history, as long as the base branch is not too old. In the end the separation of topics into commits is a question of taste and style.
Have fun!

shakemid:
My pleasure. Thank you for accepting my pullreq! And I agree for rebasing.

PR #788, opened 2016-12-18
p/jmx: extract the source from the jar

steveschnepp:


sumpfralle:
Probably we can remove the following files, as well?
```
.classpath
.gitignore
.project
META-INF/MANIFEST.MF
```

Please also adapt the following details in order to keep working with the extracted files:
* usage output in line 28 of "JMXQuery.java"
* java commandline in "jmx_"
* installation description in "readme.txt"

Alternative:
Somehow I am tempted to remove this plugin completely, instead of putting more effort into it. There is already a munin core plugin with the same name. I do not understand the difference between both (the above plugin's documentation does not mention the core plugin). Maybe it is an over-generic replacement for the core plugin? Hard to tell ...

steveschnepp:
> Please also adapt the following details 

Yeah, it was lying around on my lappy for a while, and I just pushed it in case it would be useful to someone.

> Maybe it is an over-generic replacement for the core plugin? 

It is much more generic indeed. The core plugin cannot be easily extended and this one is fairly nice in that respect. 

> There is already a munin core plugin with the same name

Note that I'm aiming at promoting this one to core instead once it's in a better shape 😅

steveschnepp:
... But then I'm also thinking about rewriting it all from scratch, as this one has a blurred license lineage.

sumpfralle:
OK. It looks like you know, what you are doing :)

sumpfralle:
Is this still relevant?

steveschnepp:
> Probably we can remove the following files, as well?

I should just provide a proper building maven/gradle building mechanism, merge it, and be done with it.

As a rewrite might not happen soonish.

steveschnepp:
I just merged it. Better than nothing ;)

PR #791, opened 2017-01-05
dont waste time by retrieving per script stats

staabm:
as per http://php.net/opcache_get_status passing false to `opcache_get_status()` will reduce unnecessary generation of per scripts stats data.


sumpfralle:

Regarding the other plugin: it is fine, that there are people out there, maintaining their own munin plugins. Of course, it would be great, if the author (or whoever feels like) would move it here, before potential lack of time or interest will let the code become stale (somewhen in the future).

PR #792, opened 2017-01-07
Http load  fix long domains

shtrom:
Munin 1.0 supports fieldnames of at most 19 characters. For domain names
longer than that, this plugin simply truncate the names in its own cache
down to 19.

This creates an issue with the `response` graph, which appends the HTTP
status code to the hostname to make multiple variables. Truncating to
this string loses this information, leading to an empty graph.


sumpfralle:
This looks good to me. Thank you!

PR #793, opened 2017-01-14
wordpress-multisite: wrong variable name

null-git:
The variable NETWORK_SIZE was filled by network_size if declared. But documentation said that the other variable was called networksize. Changed it to the latter.

(Hope this is the right place after the old pull request was done?)


sumpfralle:
yes, this is the proper way

PR #794, opened 2017-01-16
Added SNMP plugins for HP LaserJet consumption levels

kboenke:


plugins/printer/snmp__hpclj
```diff
@@ -0,0 +1,185 @@
...
+	my $oid_cur = $_[1];
+	my $oid_max = $_[2];
+	
+	if(not oidExists($oid_cur) || not oidExists($oid_max)){
+		exit 0;
+	}
+	
+	my $val_max	= $session->get_single($oid_max) || 'U';
+	my $val_cur	= $session->get_single($oid_cur) || 'U';
+	if ($val_max ne 'U' && $val_cur ne 'U') {
```
sumpfralle: Wouldn't it be useful to print the undefined value ("U"), as well?

plugins/printer/snmp__hpclj
```diff
@@ -0,0 +1,185 @@
...
+
+
+#####
+# Config
+###
+if (defined $ARGV[0] and $ARGV[0] eq "config") {
+	my ($host) = Munin::Plugin::SNMP->config_session();
+	print "host_name $host\n" unless $host eq 'localhost';
+	print "graph_title HP Printer Consumption Levels
+graph_info This graph shows Consumption-Levels on HP Printers.
```
sumpfralle: You should add a graph_category - here it should be "printing".

sumpfralle:
Regarding the graph category: please take a look at the list of [recommended categories](http://munin-monitoring.org/wiki/graph_category_list). "printing" would be the right one.

sumpfralle:
Regarding the other plugins: do you think, you could merge some of them?
Their code structure follows a repetitive pattern (but maybe this is the same for all SNMP plugins).

kboenke:
I added graph_category - thanks for the hint. These plugins have been running for me for some time now (as in years) and I honestly have not looked into combining them so far. I'll put it on my personal backlog. :)

sumpfralle:
I need to insist on the graph_category, since you picked "printer" instead of "printing" :)

Regarding the other plugins: we can certainly merge your commits even in their current state. There are just two details:
* "graph_category": "sensors" is the proper choice for UPS statistics (see [common categories](http://munin-monitoring.org/wiki/graph_category_list))
* handling of invalid values: there should be an output value for *every* field defined in "config". The value "U" (undefined) is used in case of problems. Please change all plugins in a way, that they print the "U" value (now there is nothing printed in case of problems).

It is up to you to decide, if you are happy with the current state of the plugins or if you would like to open a new merge request later on.

kboenke:
My bad. Fixed to "printing".

sumpfralle:
Thank you for the "printing".
I am a bit confused by your "closing" and "opening" of this pull request. I understand "close" as an expression for "please ignore this pull request". Is this your intention?

Regarding the "undefined" values: what do you think?

kboenke:
My understanding of closing is that this (particular) merge is completed. This is actually my very first Pull request, so please beare with me here... :)

Reverting to "undefined" was due to those OID returning "NaN" in some cases (different printer models). I decided to rather stick with U in those cases to avoid errors.

sumpfralle:
"closing" is (IMO) usually used to reject the pull request (by a maintainer) or to indicate that there is no further interest (by the contributer).
If you want to differentiate between an unfinished and a finished ("ready to merge") pull request, then it is probably common and well understood to prefix the title of unfinished requests with "WIP:". But IMO the textual interaction is usually sufficient here.
Anyway: welcome to the specific approach of contributing via pull requests :)

Regarding "undefined": I think, we misunderstand each other.
Please take a look at the [old documentation](http://munin-monitoring.org/wiki/PluginConcise#fetchmandatory) (I could not find this specific paragraph in the new documentation, right now):
`If the plugin, for any reason, has no value to report it may send the value U for undefined. It should report values for every field it has configured during config.`
Thus it would be perfect, if you would output the "U" whenever there is an unreasonable/invalid result. Right now I understand, that you skip these values and rather output nothing at all.

sumpfralle:
Ping :)

From my point of view the printer plugins are ready for merging. If you want to add the `U` output for undefined values, this would be fine - but it is not important.

Regarding the cyperpower plugins: I think, you could easily merge them into a single wildcard plugin. What do you think?
If you want to keep them here for a combined pull request, this is ok. But for cleanliness you could also move them to a new pull request. You decide.

sumpfralle:
Any thoughts?

sumpfralle:
Ping?

plugins/printer/snmp__hpclj
kboenke: The problem here is that get_single() returns 'U' in case a '0' is received. I am not quite sure if this is a bug with the function or my implementation.

Either way, here we are assuming that an undefined value for the Max-Counter (which should always be >0) determines that the counter is not available, thus return no value.

sumpfralle: OK - thank you for explaining it.

sumpfralle:
Last details:

> Regarding the cyperpower plugins: I think, you could easily merge them into a single wildcard plugin. What do you think?
If you want to keep them here for a combined pull request, this is ok. But for cleanliness you could also move them to a new pull request. You decide.

Keep it as it is or do you want to merge the CyberPower plugins into a single wildcard plugin?

kboenke:
They (the Cyberpower ones) will be translated it into a single Multigraph Plugin. Once ready I will create a new Pull for it. The HP Plugin is now a Multigraph as well. I had it running locally for some time to make sure it works. I guess it is easiest to create a new Pull for it?

Thanks for pointing out the Multigraph functionality! This should make it easier to maintain (and write new ones). 👍

sumpfralle:
I am glad, that you like the multigraph style :)

Please add a licence to the HP plugin.

Regarding the separate pull request for the HP plugin: yes - this is probably the cleanest approch.

I am looking forward to the cyberpower plugin!

PR #795, opened 2017-01-20
monit_parser: Support Authentication

recentralized:
Add optional preemptive HTTP Basic Authentication to HTTP request in order to support monit configuration like "set httpd port 2812 allow munin:s3cr3t read-only"


sumpfralle:
Perfect - thank you!

PR #797, opened 2017-02-09
Correctly detect the Routing Engines on Virtual Chassis

clonedagain:
The current recommended firmware for EX3400 (JUNOS 15.1X53-D55.5) lists the routing engines as 'Routing engine <member id>' when the switches are stacked in a Virtual Chassis.
The former code failed to detect these strings, so I relaxed it a bit.
This matches the documentation found here: https://kb.juniper.net/InfoCenter/index?page=content&id=KB17526&actp=search


sumpfralle:
This looks good - thank you!

PR #798, opened 2017-02-10
Added plugin for Powercom UPSMON PRO software

pru-mike:


plugins/sensors/upsmonpro_
```diff
@@ -0,0 +1,207 @@
...
+  /etc/munin/plugins/upsmonpro_status -> /usr/share/munin/plugins/upsmonpro_
+  /etc/munin/plugins/upsmonpro_temp -> /usr/share/munin/plugins/upsmonpro_
+  /etc/munin/plugins/upsmonpro_voltage -> /usr/share/munin/plugins/upsmonpro_
+
+=head1 CONFIGURATION
+
+Environment variables:
+
+  host - UPSMON PRO server host, default localhost
+  port - UPSMON PRO port, default 2601
```
sumpfralle: Listing the environment variables is correct and sufficient. But some (novice) users would probably appreciate something like the following for easier understanding:
```
Example configuration (optional):

  [upsmonpro_*]
  env.host localhost
  env.port 2601
```
What do you think?

plugins/sensors/upsmonpro_
```diff
@@ -0,0 +1,207 @@
...
+
+=head1 MAGIC MARKERS
+
+  #%# family=auto
+  #%# capabilities=autoconf
```
sumpfralle: Since you are offering multiple modes (temp, ...), it would be great if you could add support for the [suggest capability](http://munin-monitoring.org/wiki/PluginConcise#suggest).

plugins/sensors/upsmonpro_
```diff
@@ -0,0 +1,207 @@
...
+  temp    => [q/temp/],
+  status  => [qw/power_failure low_battery voltage_status ups_status  battery_test/]
+);
+our @TYPES = keys %TYPES;
+
+{
+  no strict 'refs';
+  for my $t (@TYPES) {
+    for my $sub (qw/autoconf/) {
+      *{ __PACKAGE__ . "::run_${sub}_$t" } = "run_$sub";
```
sumpfralle: My perl capabalities are quite rusty, thus I cannot judge, if you are using a common approach for dynamic function calls here.
If you think, this is easily understandable, then leave it like it is.
Otherwise I would suggest to use something less concise, but easier to read in order to make life easier for future contributors.

sumpfralle:
What do you think about moving this plugin from `/sensors` to `/ups`?

pru-mike:
Hello sumpfralle,
Thank you for your review.
I think i fix that.

sumpfralle:
Yes - almost! :)
But I am afraid, I did not express myself clearly with regards to sensors/ups: I did not mean the graph category (`sensors` is fine - it is mentioned in the [list of recommended graph categories](http://munin-monitoring.org/wiki/graph_category_list)).
Instead I wanted to refer to the location of the plugin within the repository. There is a plugin directory calles `plugins/ups` - I think this would be the proper location for your plugin.
What do you think?

pru-mike:
I think plugins/ups is good. And as i think i change location already. 
I see this
65  plugins/sensors/upsmonpro_ → plugins/ups/upsmonpro_
at /munin-monitoring/contrib/pull/798/commits/c9bd24bda7fd0f9e85d5691bfdfa5d77a6cabd5f#diff-a03c582d592aed4fc2078de45d4e4cc9
Pleas, tell me, am i missing something?
And should i revert graph category to sensors? (I think ups or sensors is fine)

sumpfralle:
Sorry - my bad: I overlooked your change of the path. This is fine now.
Regarding the graph category: yes, please switch it back to `sensors`, since `ups` is currently not one of the recommended graph categories.
Then it should be fine. Thank you!

sumpfralle:
Thank you!

PR #799, opened 2017-02-15
Add a plugin for the pacman package manager.

bertptrs:
This pull request adds a munin plugin for the `pacman` package manager, used on Arch Linux and related systems. It supports autoconf, and can be run as `nobody`.


plugins/other/pacman
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+# DESCRIPTION
```
sumpfralle: Please use the POD style documentation header. Maybe take a look at `plugins/munin/munin_events`.
This will allow the usage of `munindoc pacman`.

plugins/other/pacman
```diff
@@ -0,0 +1,43 @@
...
+#
+# This plugin monitors the number of pending updates in pacman.
+#
+# WARNINGS
+#
+# This plugin will issue a warning state whenever any package can be updated.
+#
+# ENVIRONMENT
+#
+# env.WARNING The number of updates at which a warning will be issued. If not
```
sumpfralle: I would suggest to use a more descriptive environment variable name here. Maybe some like `outdated_packages_warning_threshold`. Lower case names are common for environment variables used in munin plugins.

plugins/other/pacman
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
sumpfralle: You may want to switch from `bash` to `sh`, since you are using only two trivial bash-specific operators, which can be easily replaced (`[[ ... ]]` -> `[ ... ]` and `&>` -> `>/dev/null 2>&1`).
Is `bash` considered to be "essential" in an arch system?
Anyway - it is probably just a matter of taste ...

plugins/other/pacman
```diff
@@ -0,0 +1,43 @@
...
+updates.label updates
+updates.info Current number of pending updates
+EOM
+		if [[ -n $WARNING ]]; then
+			echo updates.warning $WARNING
+		fi
+		;;
+
+	autoconf)
+		hash checkupdates &> /dev/null && echo yes || echo no
```
sumpfralle: Are you sure that `hash` is the proper tool for verifying if a program is available?
Wouldn't `which` be more common?
Regarding the "no" branch: please help the user by explaining the issue: `echo "no (failed to find 'checkupdates' executable)"`.

plugins/other/pacman
```diff
@@ -0,0 +1,43 @@
...
+		;;
+
+	*)
+		echo updates.value $(checkupdates | wc -l)
+		;;
+esac
```
sumpfralle: Please specify an explicit exitcode here (`exit 0`).
Otherwise the next contributer may accidentally decide to simplify the lines 31..33 into something like `[ -n "$WARNING" ] && echo "updates.warning $WARNING"`, which would cause the exitcode of this script to be non-zero if no warning theshold was defined.

plugins/other/pacman
sumpfralle: Novice users will appreciate the documentation of the plugin configuration with an example, e.g.:
```
The number of upgradeable packages that would trigger a warning can be configured:

[pacman]
env.WARNING 3
```

plugins/other/pacman
```diff
@@ -0,0 +1,43 @@
...
+#
+# #%# family=contrib
+# #%# capabilities=autoconf
+
+case $1 in
+	config)
+		cat <<'EOM'
+graph_args --base 1000 -l 0
+graph_title Pending updates
+graph_vlabel updates
```
sumpfralle: Please add the graph category - probably `security` (see the [list of recommended graph categories](http://munin-monitoring.org/wiki/graph_category_list)).

plugins/other/pacman
bertptrs: On Arch Linux, the only package providing `sh` is `bash`, so it is always available. I think I'll stick with the small bash-isms.

plugins/other/pacman
bertptrs: Thanks, I'll update the PR.

plugins/other/pacman
bertptrs: The convention for shell scripts is ALL_CAPS for environment variables, but if Munin prefers lower case then I'll change that.

plugins/other/pacman
bertptrs: Will do, thanks.

plugins/other/pacman
bertptrs: I used to think so to, until I stumbled upon [this Stackoverflow answer](http://stackoverflow.com/a/677212) and now I use hash.

Regarding the no, since `checkupdates` is part of the `pacman` package and operating systems without it shouldn't use this plugin, isn't adding that too much clutter?

plugins/other/pacman
bertptrs: Can you clarify this? Add it to the `config` section or to all sections?

bertptrs:
Thank you for your thorough review! I added some comments/questions to yours, and will update the PR when I understand your requests.

plugins/other/pacman
sumpfralle: In a weird world, I could imagine, that:
* someone tries to use this plugin on a non-arch system
* somewhen `pacman` switches to another provider of the `checkupdates` functionality (with a different name)
Thus I would recommend to tell the user why this plugin fails.

Regarding `hash`: thank you for sharing this. As a dash/sh user, I will try to remember `command -v` from now on.

plugins/other/pacman
sumpfralle: I took a look at the plugin, noticed the absence of an explicit `exit` call and went through all branches of the `case` structure in order to check their respective last commands and possible resulting exit codes. The plugin is doing everything fine right now in this regard.
Adding an explicit `exit 0` after `esac` would clarify this without further delay (at least for me) and would add robustness against (less careful) future changes.

plugins/other/pacman
sumpfralle: Yes, I share your style preference.

There is no explicit style definition for munin - I just ran a quick "survey" over the plugins of this contrib repository (looking for the most common documentation style):
```
$ grep -Frh "env." plugins/ | sed -n '/^.*env\.[a-z]/p' | wc -l
1734
$ grep -Frh "env." plugins/ | sed -n '/^.*env\.[A-Z]/p' | wc -l
107
```
Thus, you may feel free to stick to the ALL_CAPS style.

bertptrs:
I implemented your suggestions, let me know what you think.

sumpfralle:
It looks good!
Regarding the environment variable example: you added a colon before the number (`:10`) - was that on purpose?

bertptrs:
Yes, that was on purpose, to make it more explicit that any valid Munin range may be given there. Just in case someone really wants to set the warning level at something like `3:7`, though I don't yet know why you would want that.

sumpfralle:
In this case I am happy with the current state of your plugin.

PR #800, opened 2017-02-16
Update avm-fritzbox-wan-traffic to remove spikes

Indiana8000:
Switched from COUNTER to DERIVE and added .min = 0 to remove spikes on re-connect of the WAN port (which cause a reset of the counter).


sumpfralle:
Thank you!

PR #801, opened 2017-02-16
dirsizes: fix directory labels

njoyard:
This PR changes value name computation by only keeping alphanumeric characters, hyphens and underscores. The previous list (`/\-`) was not enough to ensure value names were valid.

### DIFF
--- a/plugins/disk/dirsizes
+++ b/plugins/disk/dirsizes
@@ -26,6 +26,7 @@
 #
 
 use strict;
+use Munin::Plugin;
 my @watchdirs;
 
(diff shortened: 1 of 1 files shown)


sumpfralle:
What do you think about using the `clean_fieldname` function that is part of the munin module?

```
use Munin::Plugin;

...

my $fieldname=clean_fieldname($dev);
```

njoyard:
@sumpfralle indeed, that's way better. Updated.

sumpfralle:
Thank you!

PR #802, opened 2017-02-19
apticron

JoHartmann:
Test


JoHartmann:
Hi Lars,
I am back again for another plugin of my collection. I hope this time it will be a little bit easy to come to a final version.
Also I have changed the graph example for quota2percent_. Now it is Munin-Gallery compatible.

Greetings from the home of CeBIT
Jochen

plugins/system/apticron
```diff
@@ -0,0 +1,228 @@
...
+     exit 0
+  fi
+
+###################################################
+# Data reading sections                           #
+###################################################
+
+# call apt-get command and sepate the results
+  AptInfo=$(apt-get -u dist-upgrade --print-uris --yes | grep upgraded.)
+  Ausgabe=($(echo "${AptInfo}" | awk 'match($0,/([0-9]+) upgraded, ([0-9]+) newly installed, ([0-9]+) to remove and ([0-9]+) not upgraded./,arr){print arr[1], arr[2], arr[3], arr[4]};'))
```
sumpfralle: This may work with the current version of apt-get. But it will fail whenever there is the slightest change in the status output. `apt-get` is an interactive program. Its output is targeted at humans and thus should not be parsed.

Please look for another way to get these numbers.

plugins/system/apticron
```diff
@@ -0,0 +1,228 @@
...
+
+It showns graphs for
+
+  - upgraded
+  - newly installed
+  - to remove
+  - not upgraded
+
+The plugin itself can be stored in any directory.
+A symbolic link must be createdin the /etc/munin/plugins/ directory as follows:
```
sumpfralle: The use of symlinks in /etc/munin/plugins/ is Debian-specific.
You do not need to describe here, how munin plugins are installed. This is a task for package maintainers or the general munin documentation.

plugins/system/apticron
```diff
@@ -0,0 +1,228 @@
...
+  env.critical_holdit  [value]        (default:   2% of env.critical_update)
+
+=head1 DESCRIPTION
+
+This plugin is a graphic version of die I<apicron package>, but it sends no mails.
+
+It showns graphs for
+
+  - upgraded
+  - newly installed
```
sumpfralle: If I understand your interpretation of `apt-get`'s output correctly, then this number just describes, how many "new" packages appeared in the local package list after the last run of `apt-get update`. Are you sure, that you want to measure this state?
If yes: how could you describe it more precisely?

JoHartmann:
@sumpfralle 
Hello Lars,
thanks for your replay. I will delete the plugin! After this, please merge the new graph for quota2percent. 
Greetings from the home of CeBIT
Jochen

**PS:** In my installation (jessie / German) man apt-get show this text:

```
-u, --show-upgraded
     zeigt Pakete, von denen ein Upgrade durchgeführt werden soll; Gibt eine Liste
     aller Pakete aus, von denen ein Upgrade gemacht wurde. Konfigurationselement:
     APT::Get::Show-Upgraded.
```

sumpfralle:
Hello Jochen,

interestingly the `-u` switch disappeared somewhere between jessie and (current) stretch. Just one more indication, that there should be a better way :)

Regarding the graph: please remove the superfluous README file. Then I will merge the new graph. Thank you!

JoHartmann:
Hello Lars,
the file is deleted.
Jochen

PR #810, opened 2017-02-21
Category Tree: Reduce number of categories

dipohl:
Review of "other" plugins. If the gallery build script doesn't find pure string `graph_category  some-category` the plugin will be put into default category "other". With this commit, I fixed some of these.

Btw. We have to think about multi-plugins. The build script is at the time not able to handle multiple categories..

In this commit I made changes on one multi-plugin that defined sort of sub-categories referring all to "memcached". Therefore I set them all to their /top-category/ to reduce the number of categories we present per default in the Munin tree. If a user want's to have more or other categories shown in her/his Munin instance: set your own categories for the plugins in Munin-Master configuration (Node-Tree of munin.conf)

### DIFF
--- a/plugins/disk/disk-io-multigraph-munin-plugin
+++ b/plugins/disk/disk-io-multigraph-munin-plugin
@@ -1,2 +1,20 @@
-Check http://aouyar.github.com/PyMunin/ to get the most recent versionof the 
+graph_category other
+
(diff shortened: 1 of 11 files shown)


dipohl:
will try again and see if Travis has been healed inbetween

PR #812, opened 2017-02-21
Category Tree: Reduce number of categories

dipohl:
continue with review of plugins in category "others"


steveschnepp:
:+1: for all those PR, thx !

PR #813, opened 2017-02-21
fix crash if access.log is empty

fthiery:
If the only file found is empty (access.log with the default pattern), the following error happens:
```
/etc/munin/plugins/nginx_error: line 104: line_counts[${tmp[1]}]: bad array subscript
```
This patch will prevent the crash by testing if any values were found.


sumpfralle:
Thank you!

PR #814, opened 2017-02-21
Reduce number of categories

dipohl:
I do this with focus on usability concerning the plugin-gallery.
..users can always set up their own categories for plugins by changing the plugins source or with configuration on the Masters side in munin.conf.


sumpfralle:
hmmm - I like digital tidiness! :)

Do you have a specific approach for simplifying the categories? Or do you just go through the plugins one by one?

It looks like there is a lot of work ahead:
```
$ grep -hri "graph_category" plugins/ | sed 's/^.*graph_category[^a-zA-Z0-9_$-]*\([a-zA-Z0-9_$-]\+\).*$/\1/' | sort | uniq -c | sort -n | wc -l
234
```
(number of unique category names - including less than a dozen variable names)

I could also enjoy to work on this, if you don't mind.
But I surely don't want to to steal your meditative evening activity :)

dipohl:
Your welcome to join the party :-) To avoid merge conflicts it would be good to have a chat open. Are you on Munin IRC?

My approach at the time: 
Scroll down category list of [Plugin-Gallery](http://munin-gallery.dipohl.de/contrib/) and search for less generic terms that have only 1 or few members. Look at the single plugins and think in which general category they will fit. Change category accoringly and in some cases move plugins to other dirs to get a better 2nd level descriptor (which is the name of the dir) in the category page
e.g. [webserver](http://munin-gallery.dipohl.de/contrib/webserver-index.html)

Thanks for your feedback and offer to help ~

sumpfralle:
OK - thank you for sharing.
Regarding conflicts: I guess, I will follow my usual pattern of unexpected late night commits. Thus I guess, there will be no collisions.
I am also on the munin IRC channel, but I did not configure my client correctly, thus I tend to overlook messages sometimes.

I will probably join your efforts somewhen in the next days. Have fun!

Issue #819, opened 2017-02-25
plugins/varnish/varnish4 uses inconsistent colors


reven:
When weekly, monthly and yearly graphs are generated, they don't use the same color scheme than the daily graph. I'm guessing the colors are not defined by default and the order of the categories changes from run to run, but I don't know how to pinpoint it further. The colors of the daily graph also change with each update.

![varnish graphs](https://cloud.githubusercontent.com/assets/206993/23335362/48fe47a6-fb69-11e6-8270-d84cf798db23.png)

sumpfralle:
The problem seems to be line 1084:
```
foreach my $value (keys %values) {
```
My rusty perl knowledge ends here.
Who knows how to change this line in a way that keeps the order of keys stable?

reven:
Yes, that might be it. That would be easy to fix with:

`foreach my $value (sort keys %values) {`

I can't try it now, but I'll report back.

reven:
Ok, this seems to work after a couple of runs and forcing all the graphs to be generated again. I'll keep an eye on it and I can submit a pull request, though I don't know if this fix is the intended behavior. Is there a specific order that the values should follow that would make the graphs clearer? The option in that case would be to specify the color for each var in their definition.

sumpfralle:
It works like this:
* the munin master regularly asks for the current config
* the order of fields returned in the config is used for determining the color of each field
* the order of results (measurement values) does not matter - only the order of items returned by the "config" call

Thus your fix is technically the right one.
If you would like to produce a specific order (not based on the alphanumeric sorting of the config keys), then you would need to change a bit more (e.g. provide a "weight" attribute for each field). But from my point of view this would be clearly too much for a simple plugin :)

A pull request would be great. Thank you!

sumpfralle:
fixed by your contribution (#820)

PR #820, opened 2017-02-28
Fix inconsistent category order of varnish4_ plugin

reven:
Varnish4_ plugin displays categories in an arbitrary order, changing
the colors of the graphs with each run.

This fix introduces a 'sort' statement in the category iteration function
to force the same order consistently.

Fixes issue #819


sumpfralle:
cool - thank you!

PR #821, opened 2017-02-28
Fixed PHP shorttag. Shorttags are discouraged in newer PHP version

theyosh:
Fixed PHP script. Added full PHP tag to start. This is better for newer PHP versions


sumpfralle:
Well spot - thank you!

PR #822, opened 2017-03-04
Create twitter bot

digdilem:
Not altogether up to speed with github, but here's a working twitter stats script

https://github.com/digdilem/munin-twitter

[munin-twitter.zip](https://github.com/munin-monitoring/contrib/files/818834/munin-twitter.zip)


sumpfralle:
Thank you for contributing a new plugin!

First of all: your pull request does not contain the plugin script. Maybe you tried to add a submodule instead of the file directly?
Thus the following comments refer to https://github.com/digdilem/munin-twitter/blob/master/twitter_stats_.pl instead.

* due to the scope of the plugin I assume that `twitter_account_stats_` could be a suitable name (instead of `twitter_stats_)
* the version of the plugin is given as `$Id$`. I thought this would be the subversion-way of dynamically referring to the revision number of the last commit of this file. Please use an explicit version string (or maybe even add a concise list of releases/versions with dates).
* the four configuration settings (`consumer_key`, `consumer_secret`, `token` and `token_secret`) should be configurable via environment variables. Maybe you want to take a look at the plugin [munin_events](https://github.com/munin-monitoring/contrib/blob/master/plugins/munin/munin_events) for an example how to document such settings and how to retrieve them.
* you are offering `autoconf` - this is good. But your autoconf check always returns `yes` - this can be improved. In your case the following checks should be reasonable:
    * is `Net::Twitter` available?
    * is a username given?
    * are the required arguments given via the environment? (`consumer_*` and `token*`)
    * you should reply with `no (some reason for failure)` and exitcode zero, if one of these tests fail.
* regarding line 59: why is `no warning` necessary?

Please try to add the plugin file again and force-push your pull request.
Feel free to ask, if you are uncertain regarding the process.
Thank you for your time!

sumpfralle:
Are you still interested in this plugin?

digdilem:
Hi

I offered the plugin and I appreciated the time you took to explain the
submission process (Github is not something I use often and I'm not
familiar with it)

The plugin is still available, I use it myself on a couple of accounts and
anyone is welcome to copy it. I have not done any work to improve it as per
your suggestions, however - and probably won't as I have a number of other
priorities on my time - so I've left it dangling as "Working, but not
polished enough to be in the munin list".

S

sumpfralle:
Good - thank you for explaining your situation.

How would you like to proceed with this pull request now?
* leave it open (in this case I will ping you every few months)
* close it now (and maybe open it again somewhen if/when you want it to be included in the contrib repository)

Cheers,
Lars

digdilem:
Closing

PR #823, opened 2017-03-09
bandwidth_

jfebrerc:
Hello, I was looking for a munin plugin to calculate the bandwidth from the localhost to another host and I have not found anything, so I wrote this plugin that does that and I would like to share it so other people can use it.

### DIFF
--- a/plugins/network/bandwidth
+++ b/plugins/network/bandwidth
@@ -1,34 +1,64 @@
 #!/bin/sh
 
-#By Jose Manuel Febrer Cortés with Marco Bertola help.
 
-#bandwidth - plugin to monitor the bandwidth between localhost and another host.
+: <<=cut
 
-#Requirement: bing installed.
-#You can install bing by using (Ubuntu/Debian):
-#apt-get install bing
+=head1 NAME
+
+bandwidth_ - Plugin to monitor the bandwidth between localhost and another host.
+
+=head1 APPLICABLE SYSTEMS
+
+All systems with “bash”, and “munin”
+
+=head1 REQUIREMENTS
+
+bing installed.
+
+You can install bing by using (Ubuntu/Debian): apt-get install bing
+
+=head1 CONFIGURATION
+
+The following is the default configuration
+
+[bandwidth_*]
+user root
+env.samples 10
+env.small_packet_size 1000
+env.big_packet_size 8000
+
+- env.samples explanation: Reset stats after sending samples ECHO_REQUEST packets.
+
+- env.small_packet_size explanation: Specifies the number of data bytes to be sent in the small packets. The default and minimum value is 44.
+
+- env.big_packet_size explanation: Specifies the number of data bytes to be sent in the big packets. The default is 108. The size should be chosen so that big packet roundtrip times are long enough to be accurately measured.
 
 
-#Configuration example at munin-node:
-#       [bandwidth]
-#       user root
-#       env.host http://example.org/
-#       env.samples 10
-#       env.small_packet_size 1000
-#       env.big_packet_size 8000
-
-#env.samples explanation: Reset stats after sending samples ECHO_REQUEST packets.
-
-#env.small_packet_size explanation: Specifies the number of data bytes to be sent in the small packets. The default and minimum value is 44.
-
-#env.big_packet_size explanation: Specifies the number of data bytes to be sent in the big packets. The default is 108.
-#                                 The size should be chosen so that big packet roundtrip times are long enough to be accurately measured
+You need to do the symlink by adding the host with which you want to check the bandwidth after the "_". Example:
+ln -s /usr/share/munin/plugins/bandwidth_ /etc/munin/plugins/bandwidth_example.org
 
 
-#Stop after count resets of the stats.
-COUNT=1
+=head1 MAGIC MARKERS
+
+#%# capabilities=autoconf
+
+=head1 VERSION
+
+1.1.24
+
+=head1 AUTHOR
+
+Jose Manuel Febrer Cortés <https://www.linkedin.com/in/jfebrer/>
+Marco Bertola’s help <https://www.linkedin.com/in/bertolamarco/>
+
+=head1 LICENSE
+
+GPLv2
+
+=cut
 
 
+host="$(basename "$0" | cut -d"_" -f2)"
 case $1 in
    config)
                 echo graph_title Bandwidth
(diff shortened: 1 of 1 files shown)


plugins/network/bandwidth
```diff
@@ -0,0 +1,53 @@
...
+                echo 'graph_info This graph shows the bandwidth between localhost and ' $host
+                echo 'bandwidth.info Bandwidth statistics for ' $host
+                echo 'bandwidth.label ' $host
+        exit 0;;
+esac
+
+
+#Calculating the bandwidth
+printf "bandwidth.value ";
+bing localhost $host -c $COUNT -e $samples -s $small_packet_size -S $big_packet_size 2>/dev/null \
```
sumpfralle: Please add quotes wherever it is possible (see `shellcheck`).

Additionally you may want to add `-n` in order to avoid reverse DNS requests?

plugins/network/bandwidth
```diff
@@ -0,0 +1,53 @@
...
+        | egrep "bit$" \
```
sumpfralle: `grep` should be sufficient and easier to understand, or?

plugins/network/bandwidth
```diff
@@ -0,0 +1,53 @@
...
+        | cut -d " " -f3 \
+        | cut -c -5 \
```
sumpfralle: Maybe add an example output line in the comment below? This would make it easier for future reader to understand this.

plugins/network/bandwidth
```diff
@@ -0,0 +1,53 @@
...
+
+#Requirement: bing installed.
+#You can install bing by using (Ubuntu/Debian):
+#apt-get install bing
+
+
+#Configuration example at munin-node:
+#       [bandwidth]
+#       user root
+#       env.host http://example.org/
```
sumpfralle: Are you sure that you need an URL instead of an hostname or an IP here?

plugins/network/bandwidth
```diff
@@ -0,0 +1,53 @@
...
+#bandwidth - plugin to monitor the bandwidth between localhost and another host.
+
+#Requirement: bing installed.
+#You can install bing by using (Ubuntu/Debian):
+#apt-get install bing
+
+
+#Configuration example at munin-node:
+#       [bandwidth]
+#       user root
```
sumpfralle: Does bing require root privileges?

plugins/network/bandwidth
```diff
@@ -0,0 +1,53 @@
...
+#       env.host http://example.org/
+#       env.samples 10
```
sumpfralle: Maybe use `sample_count` instead.

plugins/network/bandwidth
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
+
+#By Jose Manuel Febrer Cortés with Marco Bertola help.
```
sumpfralle: Please turn the documentation into a perldoc style documentation header.
Maybe see plugins/munin/munin_events for an example.

plugins/network/bandwidth
```diff
@@ -0,0 +1,53 @@
...
+#env.samples explanation: Reset stats after sending samples ECHO_REQUEST packets.
+
+#env.small_packet_size explanation: Specifies the number of data bytes to be sent in the small packets. The default and minimum value is 44.
+
+#env.big_packet_size explanation: Specifies the number of data bytes to be sent in the big packets. The default is 108.
+#                                 The size should be chosen so that big packet roundtrip times are long enough to be accurately measured
+
+
+#Stop after count resets of the stats.
+COUNT=1
```
sumpfralle: Does this need to be configurable?
If yes: please use a more descriptive name.

plugins/network/bandwidth
```diff
@@ -0,0 +1,53 @@
...
+
+
+case $1 in
```
sumpfralle: I think, this plugin deserves an `autoconf` capability.
Take a look at the plugin guide: http://munin-monitoring.org/wiki/PluginConcise#autoconf

In this case you would check if `bing` is installed.

plugins/network/bandwidth
```diff
@@ -0,0 +1,53 @@
...
+   config)
+                echo graph_title Bandwidth
+                echo 'graph_vlabel Mbps'
+                echo 'graph_args --base 1000 -l 0'
+                echo 'graph_scale no'
+                echo 'graph_category network'
+                echo 'graph_info This graph shows the bandwidth between localhost and ' $host
+                echo 'bandwidth.info Bandwidth statistics for ' $host
```
sumpfralle: Maybe you want to go the extra mile of allowing to monitor the bandwidth towards more than one host.

One way of doing this would be by using the wildcard plugin style: http://munin-monitoring.org/wiki/PluginConcise#suggest

The other approach would be to interpret the "host" (or "hosts") parameter as a space separated list of hosts.

What do you think?

plugins/network/bandwidth
jfebrerc: Ok, done

plugins/network/bandwidth
jfebrerc: Yes

plugins/network/bandwidth
jfebrerc: No, you only need a hostname or an IP. My bad

plugins/network/bandwidth
jfebrerc: In the manual it appears as "samples".

plugins/network/bandwidth
jfebrerc: No, changed.

plugins/network/bandwidth
jfebrerc: Ok, done

plugins/network/bandwidth
jfebrerc: I've changed its name and I've written it as a wildcard plugin. I have also written a plugin called "multibing" which it does is show in a single graph the bandwidth of different hosts, now I will do a new pull request.

plugins/network/bandwidth
jfebrerc: Ok, done.

plugins/network/bandwidth
jfebrerc: Changed.

plugins/network/bandwidth
jfebrerc: Added.

sumpfralle:
That was a quick session of yours! :)

Your plugin looks good to me now.

Regarding your [multibandwidth plugin](#824):
It overlaps significantly in code and target audience with this plugin here. Thus I would ask you to decide which of the two plugins you want to contribute.
Contrary to popular belief the plugins in this repository are not only a gift for the world, but also a maintenance burden :)
Personally I would probably pick the other plugin with its multi-host environment setting.
What do you think?

PS: optional fine-tuning for this plugin (if you pick it): you may want to replace the outdated reference to bash and rephrase the "applicable systems" section (leaning towards circumstances where this plugin is useful instead of technical requirements)

jfebrerc:
Okay, I choose the "multibandwidth" plugin. I was trying to do something like this: https://i.gyazo.com/ad223ac828a2956709fd1139f5fc5fe7.png

Have a graph with all hosts to compare the bandwidth between them and then have the individual graphs of each host. But individual graphics can also be done using multibandwidth with different soft links.

sumpfralle:
OK - then let us continue the discussion in the other pull request (#824).
Closing this one ...

PR #826, opened 2017-03-26
Add Solaris io_disk plugin

shakemid:
Hello,
Recently I tried to fix Solaris io_* plugins in https://github.com/munin-monitoring/munin/pull/812 
After that, I tried to improve them by reference to linux diskstats plugin. It aims to replace io_* plugins because they seem to become quite old at this time. Please consider this pull request.

Improvements
- Merge io_* plugins into one multigraph plugin (like linux diskstats plugin)
- Change graph appearance like linux diskstats plugin
  But, RRD files generated by io_busy_, io_ios_, io_bytes_ can be taken over
- Add queue and latency functions
- Remove perl code and replace old Solaris::MapDev feature by using iostat
- Add POD style document
- Refactoring, using functions, replace variable names, check with  shellcheck, ...


plugins/solaris/io_disk
```diff
@@ -0,0 +1,301 @@
...
+
+    echo "graph_title $title_type $conf_title"
+    echo 'graph_category disk'
+    echo "graph_args $conf_gargs"
+    echo "graph_vlabel $conf_vlabel"
+    if [ -n "$graph_width" ]; then
+        echo "graph_width $graph_width"
+    fi
+
+    kstat -p -c "$class" -m "$module_regex" -s "/^${conf_in}\$/" | sed 's/:/ /g' | awk '{ print $3 }' | \
```
sumpfralle: What is the result of this chain of commands?
A list of full device paths?
Maybe assign the result to a variable with a suitable name? This would make this line easier to understand.

plugins/solaris/io_disk
```diff
@@ -0,0 +1,301 @@
...
+    while read -r dev
+    do
+        if is_excluded "$dev" ; then
```
sumpfralle: Another common pattern (maybe easier to read) would be the following:
```is_excluded "$dev" && continue```
(just a matter of taste)

File: plugins/solaris/io_disk [hunk 3]
```diff
@@ -0,0 +1,301 @@
...
+            continue
+        fi
+
+        # replace instance name to logical device name if exists
+        ref="name_$dev"
+        devname=${!ref:-}
```
sumpfralle: Maybe you want to add a comment regarding the indirect evaluation of `${!ref:-}`?
(at least I needed to look it up in the bash manual)

plugins/solaris/io_disk
```diff
@@ -0,0 +1,301 @@
...
+    echo "multigraph ${plugin_name}_${func}_${module}"
+
+    kstat -p -c "$class" -m "$module_regex" -s "$stat_regex" | sed 's/:/ /g' | awk '{ print $3,$4,$5 }' | \
+    while read -r dev stat value
+    do
+        if is_excluded "$dev" ; then
+            continue
+        fi
+
+        echo "${dev}_${stat}.value ${value%.*}"
```
sumpfralle: Do you use `${value%.*}` for removing fractional digits? Doesn't this unnecessarily discard precision?

plugins/solaris/io_disk
```diff
@@ -0,0 +1,301 @@
...
+
+=cut
+
+# Include plugin.sh
+. "${MUNIN_LIBDIR:-}/plugins/plugin.sh"
+is_multigraph "$@"
+
+# Shell options
+set -o nounset
+set -o pipefail
```
sumpfralle: Does `pipefail` serve any purpose without `errexit`?
If yes: maybe add a comment?

plugins/solaris/io_disk
```diff
@@ -0,0 +1,301 @@
+#!/bin/bash
+
+: << =cut
+
+=head1 NAME
+
+  io_disk - Multigraph plugin to monitor disks for Solaris.
+
+  These functions are implemented:
```
sumpfralle: Maybe add a comment how this plugin relates to similar existing plugins?

shakemid:
Thank you for your comment! 
I wrote this plugin to fully replace io_busy_, io_ops_, io_bytes_ plugins placed in plugins/node.d.sunos [https://github.com/munin-monitoring/munin/tree/master/plugins/node.d.sunos] (not contrib). I actually would like to replace them.
In most cases, those plugins are used as io_{busy,ops,bytes}\_sd (means SCSI, SAS, SATA disks). This plugin I wrote can simply replace them, and can take over RRDs generated by them. In case they used as io_*_{nfs,tape}, this plugin can take over data by setting env.module as nfs or tape.

plugins/solaris/io_disk
shakemid: I see. I will add comments.

plugins/solaris/io_disk
shakemid: It enables to get return code ($?) when command fails in front of pipe. (e.g bad_command | good_command; echo $?)
Without -o pipefail, shell returns status of good_command (0). 
But, I didn't use it in this plugin. Is it better to remove it?

plugins/solaris/io_disk
shakemid: Agree. I will add comments with sample output like following.

```
$ kstat -p -c disk -m '/^sd$/' -s '/^reads$/'
sd:0:sd0:reads  51432585
sd:1:sd1:reads  52671416
...
```

plugins/solaris/io_disk
shakemid: Yes, it looks better.

plugins/solaris/io_disk
shakemid: Yes, I do. I think rrdtool can accept only integers, can't it? Since, cumulative I/O ops and I/O bytes are quite large values (like millions), I think discard precision does not really matter.

plugins/solaris/io_disk
shakemid: Agree. I will add a comment. Indirect evaluation looks pragmatic, but it may be better than eval. 
I actually prefer hashes, but it is supported above bash 4. Solaris 11 has bash 4, but Solaris 10 has only bash 3.

shakemid:
I tried to fix it for your advice. Shell option pipefail and removing fractional digits still remain.
Would you check it again?

sumpfralle:
Thank you for your changes!
Regarding the fractional precision and `pipefail`: personally I would remove both, since they would tend to raise a question for future readers, even though both are of no specific use. But this is a matter of taste.

Should I merge it as it is or do you want to add more changes?

shakemid:
I agree. I would like to respect your opinion :-)
I will remove pipefail. But, I believe RRDtool DERIVE and COUNTER can accept only integer, right? (Recent RRDtool has DDERIVE and DCOUNTER)
So, rounding will be better than flooring in this case, like $( printf "%lu" "$value" ) . Does it look okay?

shakemid:
Ah, bash printf looks not to accept floating values. 
```
$ echo $( printf "%lu" 1.5 )
-bash: printf: 1.5: invalid number
0
```
Awk looks better.
```
$ echo $( echo 1.5 | awk '{ printf "%lu", $1 + 0.5 }' )
2
```

sumpfralle:
Indeed - this is a good point!
I was thinking about this, too - but spontaneously I failed to find details about rrd's data format (integer, float, ...).

Now I took another look and found this comment: http://stackoverflow.com/a/22265765
Thus it seems, that rrd would use floats even for COUNTER and DERIVE.

This would simplify your formatting code line :)

shakemid:
Thank you for your information. I didn't know about that.

I tried to update with floating values. But, rrdcached failed with following error. Maybe rrdupdate can accept floating value, but rrdcached can not, even my rrdtool version is the newest 1.6.0 .
```
rrdcached[27662]: [ID 899100 daemon.notice] queue_thread_main: rrd_update_r (....-io_busy_sd-sd0_wtime-d.rrd) 
failed with status -1. (...-io_busy_sd-sd0_wtime-d.rrd: not a simple signed integer: '761.702936605')
```
Anyway, integers look safe in any cases. And, removing fractional digits by variable substitution is smaller overhead than forking awk or perl. Thus, I would like to remove fractional digits in this case. What do you think?

shakemid:
And also, I think I should change magic marker.
```
  #%# family=contrib
```

sumpfralle:
Thank you for trying the interger/float issue out. Based on your experiment, my interpretation is obviously wrong. I am sorry for wasting your time!

Regarding the `family`: based on the [munin guide](http://guide.munin-monitoring.org/en/latest/architecture/syntax.html#family) I understand our current conversation within this pull request as a "check for conformity" - thus `auto` should be suitable.

Since my confusion now seems to be cleared up, only one open question is left.
Do you want to  remove `pipefail`? Otherwise I will just merge it as it is.

shakemid:
No problem! Thank you for your understanding.
I will remove pipefail. And, I would like to add some code and test more. Please give me some time.

shakemid:
I have removed pipefail, and added some codes. I have tested with several Sol 10 and 11 hosts about a day, and it looked working fine.
Would you check it?

sumpfralle:
Your changes look good to me.
I would merge the plugin, if you deem it to be finished.

shakemid:
Yes, please. Thank you so much for your helpful review!

PR #831, opened 2017-04-07
dspam_ fixes

jolan78:
- typo
- cleaning invalid lockfile
- new graph relprocessed in messages / minute

### DIFF
--- a/plugins/dspam/dspam_
+++ b/plugins/dspam/dspam_
@@ -61,7 +61,7 @@ EXAMPLES and USAGE for details).
 Link this plugin to /etc/munin/plugins/ and restart the munin-node. The link
 should be in the format: dspam_<graph>_<target>, where:
 
- graph      - One of: accuracy, processed, absprocessed.
+ graph      - One of: accuracy, processed, absprocessed, relprocessed.
  target     - The uid that DSPAM generates in dspam_stats output,
               but converted to Munin internal name format. Normally
               this means that non-alphabetic and non-numeral characters
@@ -87,6 +87,10 @@ The plugin supports the following graph types:
                 sorted by the classification that DSPAM uses. The
                 numbers are stacked, making the height of the column
                 display the increase of processed messages over time.
+
+ relprocessed - Shows the same data as dspam_absprocessed_, but as
+                messages per minute instead of ever-growing asolute
+                values.
 
  processed    - Shows the same data as dspam_absprocessed_, but as
                 percentage of the total amount of processed messages,
@@ -340,6 +344,7 @@ print_suggest() {
 	echo accuracy_ALL
 	echo processed_ALL
 	echo absprocessed_ALL
+	echo relprocessed_ALL
 }
 
 #
@@ -438,7 +443,7 @@ print_config() {
 			echo nc.draw AREASTACK
 			;;
 
-		absprocessed)
+		*processed)
 			if [ -n "$pattern" ]; then
 				debug env.pattern was set, so use it: $pattern
 				local uid=$description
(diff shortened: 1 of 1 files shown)


plugins/dspam/dspam_
```diff
@@ -211,7 +215,15 @@ file_is_locked() {
 	local file=$1
 	local lock=$1.lock
 
-	[ -f $lock ] && debug file $file is locked && return 0 # EX_OK
+	if [ -f $lock ];then
```
sumpfralle: Please quote variables, whenever it is possible.
Probably `shellcheck -s dash dspam_` would help you finding all occurrences.

plugins/dspam/dspam_
```diff
@@ -211,7 +215,15 @@ file_is_locked() {
...
+		local pid=$(cat $1.lock)
+		if [ $(ps h -p $pid -o comm=|grep -c dspam_) -eq 0 ]; then
```
sumpfralle: Maybe the following is easier to read?
```if ps h -p "$pid" -o comm= | grep -q dspam_; then```
(and switch both branches)

File: plugins/dspam/dspam_ [hunk 3]
```diff
@@ -211,7 +215,15 @@ file_is_locked() {
...
+			debug lock for file $file is no longer valid
+			file_remove_lock $file
+		else
+			debug file $file is locked && return 0 # EX_OK
+		fi
+	fi
 	debug file $file is not locked
```
sumpfralle: The following two lines should probably move to the "no lock" branch above?

plugins/dspam/dspam_
```diff
@@ -211,7 +215,15 @@ file_is_locked() {
...
 	local lock=$1.lock
 
-	[ -f $lock ] && debug file $file is locked && return 0 # EX_OK
+	if [ -f $lock ];then
+		local pid=$(cat $1.lock)
+		if [ $(ps h -p $pid -o comm=|grep -c dspam_) -eq 0 ]; then
+			debug lock for file $file is no longer valid
+			file_remove_lock $file
+		else
+			debug file $file is locked && return 0 # EX_OK
```
sumpfralle: Maybe split both commands?
This will make the flow-breaking `return` a bit more obvious.
Additionally it removes the conditional relationship between these (unrelated) commands.

plugins/dspam/dspam_
```diff
@@ -452,26 +465,34 @@ print_config() {
 			echo "graph_title Processed messages for $uid"
 			echo graph_category spamfilter
 			echo graph_args --base 1000
-			echo graph_vlabel Messages
+			[ $graph == absprocessed ] && echo graph_vlabel Messages
```
sumpfralle: This plugin uses `/bin/sh` (not `/bin/bash`) - thus you should avoid bashisms:
```[ "$graph" = absprocessed ] && echo graph_vlabel Messages```

jolan78:
I updated the commits according to your observations.
regarding quoted variables, I was trying to stick to the general code style of the file. I fixed some of the unquoted variables but didn't change those which are unquoted in the remaining code. tell me if that's OK for you.

sumpfralle:
It is much better thank before - thank you!

sumpfralle:
btw: I am heavily confused by the usage of the "statsfile" within in this plugin (a third of the plugin code deals with the locking of that file). As far as I understand the code, the statsfile is used only for storing intermediate processing results. Thus this data could be easily stored in a simple variable instead of working around temporary files and locks. Or do I miss something?
(just being curious)

jolan78:
the statefile is used to avoid running dspam_stats too much. if the file is less than 3 minutes old (by default) the file is not generated again. it's usefull when you have several graphs type , when you want one graph per user , or when you have several munin instances pulling the data from your server.

sumpfralle:
Ah - I see. Thank you!

PR #832, opened 2017-04-13
support hostname grep in postfix_stats

4nd3r:


plugins/postfix/postfix_stats
```diff
@@ -110,7 +121,11 @@ fi
 
 # Variable to store the pflogsumm result.
 # shellcheck disable=SC2086
```
sumpfralle: Please move this shellcheck attribute down to the line it belongs to. See the output of `shellcheck -s dash postfix_stats`.

plugins/postfix/postfix_stats
```diff
@@ -59,7 +64,9 @@ GPLv2
 =cut
 
 #set -xv
-SYS_LOG="${logfiles:-/var/log/syslog /var/log/syslog.0}"
+MAIL_LOG="${logfiles:-/var/log/mail.log /var/log/mail.log.1}"
+
+GREP_HOSTNAME=$(echo $0 | sed -r 's/^.+postfix_stats_?//')
```
sumpfralle: Please take care for proper quoting (here `"$0"`). See the output of `shellcheck -s dash postfix_stats` for details.
Maybe `GREP_HOSTNAME=$(basename "$0" | sed -r '^postfix_stats_?//')` would be even more precise?

plugins/postfix/postfix_stats
```diff
@@ -27,6 +28,8 @@ env.pflogsumm The "pflogsumm" script, can be pflogsumm.pl if it was manually
 downloaded and installed, or "pflogsumm" if it was installed by a package 
 manager (like apt-get).
 
+Use the last part of the symlink name for grepping the hostname from log.
```
sumpfralle: Please provide an example here - otherwise this feature could be hard to understand.

sumpfralle:
Thank you!

There is one thing, that I did not notice before: you are using `grep` without the `-F` switch, even though you are probably looking for a fixed string (ignoring dots and so on). Maybe you want to change this (or change the description of the environment variable).

In order to avoid misunderstandings: is your use-case a server where the logs of multiple mailservers end up (e.g. received via rsyslog from remote hosts)? Maybe you should mention this, in order to clarify the meaning of `HOSTNAME`.

4nd3r:
Thank you for your time. As I'm not professional programmer, your comments were really helpful.

sumpfralle:

PR #833, opened 2017-04-17
Add interrupts, forks, fsstat_bytes, fsstat_act_ plugins for Solaris

shakemid:
Hello,
I would like to add some plugins for Solaris OS, that look lacking in comparison with Linux OS. Please consider thus pull request.

- interrupts: plugin to monitor cpu interrupts, context switches and related events. It can be equivalent of Linux interrupts plugin. 
- forks: Plugin to monitor fork and exec rate. It can be equivalent of Linux forks plugin.
- fsstat_bytes: Plugin to monitor file system statistics. 
- fsstat_act_: Plugin to monitor detail activities of file system statistics.

### DIFF
--- a/plugins/solaris/forks
+++ b/plugins/solaris/forks
@@ -1,0 +1,112 @@
+#!/bin/bash
+
+: << =cut
+
+=head1 NAME
+
+  forks - Munin plugin to monitor Solaris fork and exec rate
+
+=head1 CONFIGURATION
+
+  Make symlink:
+    cd /path/to/munin/etc/plugins
+    ln -s /path/to/munin/lib/plugins/forks .
+
+=head1 AUTHOR
+
+  K.Cima https://github.com/shakemid
+
+=head1 LICENSE
+
+  GPLv2
+
+=head1 Magic markers
+
+  #%# family=contrib
+  #%# capabilities=autoconf
+
+=cut
+
+# Include plugin.sh
+. "${MUNIN_LIBDIR:-}/plugins/plugin.sh"
+
+# Shell options
+set -o nounset  # Like perl use strict;
+
+# Graph settings
+global_attr="
+    graph_title Fork and exec rate
+    graph_category processes
+    graph_args --base 1000 --lower-limit 0 --rigid
+    graph_vlabel count per second
+    graph_info Fork and exec rate
+"
+# data_attr format: field type draw label
+#   label can contain white-spaces.
+data_attr="
+    sysfork  DERIVE LINE fork
+    sysvfork DERIVE LINE vfork
+    sysexec  DERIVE LINE exec
+"
+
+# Functions
+
+autoconf() {
+    if [ -x "$( which kstat )" ]; then
+        echo yes
+    else
+        echo "no (failed to find executable 'kstat')"
+    fi
+}
+
+config() {
+    local label_max_length=45
+
+    # print global attributes
+    sed -e 's/^  *//' -e '/^$/d' <<< "${global_attr}"
+
+    # print data source attributes
+    # split line into field,type,draw,label 
+    local fields field type draw label
+    fields=
+    while read -r field type draw label
+    do
+        [ -z "$field" ] && continue
+        fields="${fields} ${field}"
+
+        echo "${field}.type ${type}"
+        echo "${field}.draw ${draw}"
+        echo "${field}.label ${label:0:${label_max_length}}"
+        if [ "${type}" = DERIVE ]; then
+            echo "${field}.min 0"
+        fi
+    done <<< "${data_attr}"
+
+    echo graph_order "$fields"
+}
+
+getvalue() {
+    local field type draw label
+    while read -r field type draw label
+    do
+        [ -z "$field" ] && continue
+        value=$( kstat -p "cpu::sys:${field}" | awk '{ sum += $2 } END { print sum }' )
+        echo "${field}.value ${value}"
+    done <<< "${data_attr}"
+}
+
+# Main
+case ${1:-} in
+autoconf)
+    autoconf
+    ;;
+config)
+    config
+    [ "${MUNIN_CAP_DIRTYCONFIG:-}" = "1" ] && getvalue
+    ;;
+*)
+    getvalue
+    ;;
+esac
+
+exit 0
--- a/plugins/solaris/forks
+++ b/plugins/solaris/forks
@@ -52,7 +52,7 @@ data_attr="
 # Functions
 
 autoconf() {
-    if [ -x "$( which kstat )" ]; then
+    if which kstat >/dev/null ; then
         echo yes
     else
         echo "no (failed to find executable 'kstat')"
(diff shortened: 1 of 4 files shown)


plugins/solaris/fsstat_act_
```diff
@@ -0,0 +1,175 @@
+#!/bin/bash
```
sumpfralle: Here you are using only one bash feature: `${...^^}` (uppercase conversion).
Maybe you would like to switch to `/bin/sh` and use `tr '[a-z]' '[A-Z]'` for case conversion?

plugins/solaris/fsstat_bytes
```diff
@@ -0,0 +1,172 @@
...
+    return 1
+}
+
+get_zone_id() {
+    local osver zonename zoneid
+
+    zoneid=0
+    osver=$( uname -r | cut -d. -f2 )
+    
+    if [ "$osver" -ge 11 ]; then
```
sumpfralle: Same as above: why is the version test necessary?

plugins/solaris/fsstat_bytes
```diff
@@ -0,0 +1,172 @@
...
+        zonename=$( zonename )
+        zoneid=$( /usr/sbin/zoneadm list -p | awk -F: '$2 == "'"$zonename"'" { print $1 }' )
+    fi
+
+    echo "$zoneid"
+}
+
+autoconf() {
+    if [ -x "$( which kstat )" ]; then
```
sumpfralle: Same as above - may remove test?

plugins/solaris/fsstat_bytes
```diff
@@ -0,0 +1,172 @@
...
+}
+
+config() {
+    local fs
+
+    # Print global attributes
+    echo "${global_attr}" | sed -e 's/^  *//' -e '/^$/d' 
+
+    # Get fs names by kstat
+    kstat -p "unix:${zone_id}:${name_regexp}:${data_in}" | \
```
sumpfralle: Multiline statement with indentation?

plugins/solaris/fsstat_bytes
```diff
@@ -0,0 +1,172 @@
...
+    sed -e 's/vopstats_//' -e 's/:/ /g' | awk '{ print $3 }' | sort -u | \
+    while read -r fs
```
sumpfralle: Is it safe to assume that `fs` is always "clean"?

plugins/solaris/fsstat_bytes
```diff
@@ -0,0 +1,172 @@
...
+
+    # Get fs names, stat names and values by kstat
+
+    # kstat output example:
+    #  $ kstat -p 'unix::/^vopstats_[a-z]/:nread'
+    #  unix:0:vopstats_autofs:nread    2
+    #  unix:0:vopstats_hsfs:nread      407790
+    #  ...
+
+    kstat -p "unix:${zone_id}:${name_regexp}:/^(${data_in}|${data_out})\$/" | \
```
sumpfralle: Multiline statement with indentation?

plugins/solaris/fsstat_bytes
```diff
@@ -0,0 +1,172 @@
+#!/bin/bash
```
sumpfralle: I think, here you are using no bash features at all. Maybe switch to `/bin/sh`?

plugins/solaris/interrupts
```diff
@@ -0,0 +1,132 @@
...
+    rw_wrfails     DERIVE LINE dummy
+    syscall        DERIVE LINE system calls
+    trap           DERIVE LINE traps
+"
+# They can be shown by vmstat -s or mpstat
+
+# Functions
+
+autoconf() {
+    if [ -x "$( which kstat )" ]; then
```
sumpfralle: Maybe remove test?

plugins/solaris/forks
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
```
shakemid: As you said, `echo` looks enough in this case. I would like to change `<<<` to `echo`. 
But, I believe command substitution with `$()` is the feature of bash (or POSIX shell). Unfortunately, `/bin/sh` of Solaris 10 (or less) is not POSIX shell and does not support `$()`. I don't think back-quotes are good for some reason (nesting, readability...).
In my opinion, `/bin/bash` seems safer and compatible for Solaris. It should work with Solaris 8 or above.

plugins/solaris/forks
```diff
@@ -0,0 +1,112 @@
...
+data_attr="
+    sysfork  DERIVE LINE fork
+    sysvfork DERIVE LINE vfork
+    sysexec  DERIVE LINE exec
+"
+
+# Functions
+
+autoconf() {
+    if [ -x "$( which kstat )" ]; then
```
shakemid: I agree. I will change them.

plugins/solaris/fsstat_act_
```diff
@@ -0,0 +1,175 @@
...
+  can be shown by fsstat command.
+
+  Make symlink:
+    cd /path/to/munin/etc/plugins
+    ln -s /path/to/munin/lib/plugins/fsstat_act_ fsstat_act_zfs
+    ln -s /path/to/munin/lib/plugins/fsstat_act_ fsstat_act_nfs3
+    ln -s /path/to/munin/lib/plugins/fsstat_act_ fsstat_act_nfs4
+    ...
+
+=head1 ENVIRONMET VARIABLES
```
shakemid: Thanks! I will fix it.

plugins/solaris/fsstat_act_
```diff
@@ -0,0 +1,175 @@
...
+
+# Functions
+
+get_zone_id() {
+    local osver zonename zoneid
+
+    zoneid=0
+    osver=$( uname -r | cut -d. -f2 )
+    
+    if [ "$osver" -ge 11 ]; then
```
shakemid: Because Solaris 10 does not support fsstat with zone id. I will add a comment.

plugins/solaris/fsstat_act_
```diff
@@ -0,0 +1,175 @@
...
+        echo yes
+    else
+        echo "no (failed to find executable 'kstat')"
+    fi
+}
+
+suggest() {
+    # Print file systems which look active
+
+    kstat -p "unix:${zone_id}:${name_regexp}:/^(read_bytes|write_bytes)\$/" | \
```
shakemid: I agree. I will change style like this.
```
kstat ... \
| sed ... \
| awk ...
```

plugins/solaris/fsstat_act_
```diff
@@ -0,0 +1,175 @@
...
+config() {
+    local stat
+
+    # Print global attributes
+    echo "${global_attr}" | sed -e 's/^  *//' -e '/^$/d' 
+
+    # Get stat names by kstat
+    kstat -p "unix:${zone_id}:vopstats_${fs_type}:${stat_regexp}" | \
+    sed -e 's/vopstats_//' -e 's/:/ /g' | awk '{ print $4 }' | sort -u | \
+    while read -r stat
```
shakemid: Yes, stat and fs contain only alphabets and numbers and not 'root'.

plugins/solaris/fsstat_act_
```diff
@@ -0,0 +1,175 @@
...
+    do
+        # Print data attributes
+        echo "${stat}.label ${stat#n}"
```
shakemid: kstat returns stat names like nread, nwrite, nopen, nclose... I felt better to remove prefix n for graph labels.

plugins/solaris/fsstat_act_
shakemid: I agree. `^^` is short but `tr` looks easy to understand.

plugins/solaris/fsstat_bytes
shakemid: Yes, fs only contain alphabets and numbers.

shakemid:
Thank you for your review! I have fixed for your advice. Would you check again?

sumpfralle:
Regarding sh/bash and `$(...)` on Solaris: what an unpleasant surprise! Sticking with bash is surely the best option under these circumstances.

All plugins look good to me. Thank you!

PR #835, opened 2017-04-18
Use $MUNIN_PLUGSTATE instead of hardcoded path

dmzkrsk:
`/var/lib/munin/plugin-state` even doesn't exist on Ubuntu 16.04


sumpfralle:
Thank you!

PR #837, opened 2017-04-24
graph_period for postfix_mailstats

SteffenRitschel:
graph_period is now configurable for low-traffic systems.
also fixed issue with syntax for graph_period in config-function

### DIFF
--- a/plugins/postfix/postfix_mailstats
+++ b/plugins/postfix/postfix_mailstats
@@ -26,6 +26,7 @@
 #   group adm
 #   env.logdir /var/log
 #   env.logfile mail.log
+#   env.graph_period seconds #minute
 
 use strict;
 use warnings;
@@ -34,6 +35,7 @@ use Munin::Plugin;
 
 my $LOGDIR  = $ENV{'logdir'}  || '/var/log';
 my $LOGFILE = $ENV{'logfile'} || 'mail.log';
+my $graph_period = $ENV{'graph_period'} || 'seconds';
 my $logfile = $LOGDIR .'/'. $LOGFILE;
 
 my $delivered;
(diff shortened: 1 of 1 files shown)


sumpfralle:
Hello Steffen,

thank you for taking a look at this plugin!

But I think, your approach went slightly into the wrong direction:
* `${graph_period}` is a verbatim template pattern that can be used in the configuration part of a munin plugin. It is not supposed to be substituted by the shell - thus the escaping of the `$` (sadly this is not well documented and can easily confuse readers due to its similarity with shell variables)
* you could change the default time unit of a plugin by adding `graph_period minute` to the plugin configuration (see [graph_period attribute](http://munin-monitoring.org/wiki/graph_period))
* alternatively you can also change this value in the munin node configuration for this plugin (see [faq](http://munin-monitoring.org/wiki/faq#Q%3aCanIrunmuninatdifferenttimeintervalsthanthedefault))

Thus I think, that your use-case can already be accomplished via configuration. Or what do you think?

SteffenRitschel:
Hi sumpfralle,

yes, usually this is right, but as far as i tested the changes in the config did not work as expected.
I got ${graph_period} as string, and not as replaced pattern in my tests and never got it running to see other values in the graph. you can try this with munin-run postfix_mailstats and change  the variable env.graph_period to other values than second

sumpfralle:
> I got ${graph_period} as string, and not as replaced pattern in my tests [..]

Maybe this is a misunderstanding: `munin-run FOO config | grep "graph_vlabel"` is supposed to return the verbatim token `${graph_period}`. You may want to try this out with some core plugins. See for example the `swap` plugin:
```
$ munin-run swap config | grep "graph_period"
graph_vlabel pages per ${graph_period} in (-) / out (+)
```

This template pattern is replaced in a later stage (when drawing the graph).

If you are still confident, that something is wrong, then please describe:
* what you did
* what you observed
* what you expected

sumpfralle:
Ping?

PR #838, opened 2017-04-25
[debian/debsecan] Count unique CVEs

shtrom:
This adds a second mode to debsecan. There is also a symlink to the new name, for backward compatibility with existing installations.


plugins/debian/debsecan_
```diff
@@ -0,0 +1,183 @@
...
+fixed.critical ${FIXEDCRIT}
+EOF_
+  exit 0
+fi
+
+ALL=$(debsecan --suite "${SUITE}" 2> /dev/null)
+REMOTE=$(echo "$ALL" | grep 'remotely')
+NONREMOTE=$(echo "$ALL" | grep -v 'remotely')
+
+HIGH=$(echo "${NONREMOTE}" | grep 'high urgency')
```
sumpfralle: Please add `-w` to all grep expressions (for more precise matching).

plugins/debian/debsecan_
```diff
@@ -0,0 +1,183 @@
...
+MEDIUM=$(echo "${NONREMOTE}" | grep 'medium urgency')
+LOW=$(echo "${NONREMOTE}" | grep 'low urgency')
+OTHER=$(echo "${NONREMOTE}" | grep -v 'urgency')
+FIXED=$(echo "${ALL}" | grep '(fixed')
+
+case "${MODE}" in
+	'cve')
+		remote_count=$(echo "${REMOTE}" | cut -f "${CUT_FIELD}" -d " "| sort | uniq | wc -l)
```
sumpfralle: Maybe you would want to use a separate function for the repetitive call chain `echo "$FOO" | cut -f "${CUT_FIELD}" -d " "| sort`.
This would simplify the following lines and the lines at the end of the plugin.

plugins/debian/debsecan_
```diff
@@ -0,0 +1,183 @@
...
+		;;
+	'pkg' | *)
+		TITLE_ADD="package "
+		CUT_FIELD=2
+		;;
+esac
+
+if [ "$1" = "config" ] ; then
+  cat <<EOF_
+graph_title DebSecan : ${TITLE_ADD}vulnerabilities for ${SUITE}
```
sumpfralle: The whitespace before the colon is quite unusual, or?

shtrom:
Heh. I wrote most of the previous version, so might as well (:

sumpfralle:
Thank you for taking the time!

If you would want to go the extra mile, you could additionally add quotes around the `${FIELD}` instances (see `shellcheck`) and mark the `CUT_FIELD` variable in the function as `local`.
Again: I would also merge your changes as they are, if you are satisfied with the current state.

shtrom:
Odd, shellcheck didn't pick those up...

sumpfralle:
Hm - interesting ...
Anyway: thanks!

PR #841, opened 2017-04-27
Add zfs_arcstats

shakemid:
Hello,
I would like to add zfs_arcstats plugin for Solaris. This plugin monitors ZFS ARC (Adaptive Replacement Cache) statistics. 
I wonder which directory is suitable to place this plugin. Although there is 'zfs' directory, but those plugins under zfs directory seem to be for FreeBSD and Linux. And, this plugin is for Solaris. So, I think 'solaris' directory looks better. What should I do?


sumpfralle:
I also think, that "solaris" is probably more specific than "zfs". Thus this should be a good place for the script.
Maybe we will need to move it somewhen, if you (or someone else) adds the BSD or Linux specific data retrieval that you outlined in a comment.

Your plugin looks good - thank you for your contribution!

shakemid:
Thank you for merging! I would like to add support for Linux/FreeBSD in future!

PR #842, opened 2017-04-28
Improve oracle_sysstat plugin

shakemid:
I would like to improve oracle_sysstat plugin. Mainly I changed it wildcard-plugin to multi-graph plugin. 
I found multi-graph very smart. It can be easy to setup. But, usage of 'asmusage' module is compromising. 
Please consider this pull request.

Changes are following
- Change to multi-graph plugin
- Add dirtyconfig
- Add config for multi-instance databases
- Cleanup code


sumpfralle:
In general, I like your changes.

Regarding the switch from `suggest` to `multigraph`: do you see a specific benefit connected to this change?

shakemid:
multigraph plugin looks easier to setup than wildcard plugin, especially for multiple instance environment.
Sometimes multiple Oracle database instances cab be setup in single host for several reasons. With wildcard version, it is necessary to make dozens of symlinks to monitor multiple Oracle instances.
```
e.g.
  ln -s /path/to/lib/plugins/oracle_sysstat_ oracle_sysstat2_cputime
  ln -s /path/to/lib/plugins/oracle_sysstat_ oracle_sysstat2_cachehit
  ...
  ln -s /path/to/lib/plugins/oracle_sysstat_ oracle_sysstat3_cputime
  ln -s /path/to/lib/plugins/oracle_sysstat_ oracle_sysstat3_cachehit
  ...
```
That was tiring to me. With multigraph plugin, only one symlink is necessary for one instance. In addition, multigraph plugin can reduce forking process cost for munin-update. That's why I decided to switch this plugin from wildcard to multigraph.

shakemid:
Excuse me. I fixed typo in document.

sumpfralle:
I had a quick discussion with people on the [munin mailinglist](https://sourceforge.net/p/munin/mailman/message/35813351/) regarding the switch from suggest to multigraph. This kind of change (in general) was warmly welcomed by the participants of the discussion.

Thus I would merge your changes now, if you think, your are finished.

Just a last question: did you check if the names of the rrd files remain unchanged? This would be good for the current users of this plugin.

shakemid:
Thank you for discussing about this. I am a frequent contributor:-)
I am glad to know multigraph and dirtyconfig are preferable!
Regarding rrd file names, yes, file names remain unchanged. Thus, rrd files generated by wildcard version can be taken over.
I think I am finished. Please go ahead to merge.

sumpfralle:
Great - thank you!

PR #843, opened 2017-05-15
changed category from search to redis

denics:


sumpfralle:
Ah - yes. `search` is indeed a really unconventional graph category.

What do you think about `db` instead of `redis`? (see the [list of well-known categories](http://munin-monitoring.org/wiki/graph_category_list))

denics:
Hi @sumpfralle, first of all thanks for the module. I just think it needs a more explicit README, it took me a bit to understand how to configure it, and I had to look at the code to configure the socket. I can provide you with few lines if you want. Concerning the categories, it is true that in munin we have the "well known categories",  but being redis a quite complex service, I would stick with "redis", as you have, for example, php-fpm or apache categories. As a sysadmin using redis for caching I would expect it under memory, if not under redis. But if you are a db admin, maybe you would like to see these info under db. This is why I would prefer Redis to have its own category: redis

dipohl:
Hi denics, 

if we set "redis" as category in the delivered plugin, we will get bad retrieval in the Munin-Gallery (which should operate as _Plugin-Shop_) when using a search engine, because on all index pages we have the list of categories in the navigation bar on the left side. 

But the search for `redis` gets significant hits if redis is **not the name of the category but of the directory where the plugin is stored in the contrib repo** as in that case `redis` is printed as header of the sections where redis plugins are listed.

For demonstration see here: 

1. Search in my instance of the Munin Plugin Gallery which is cleared of outdated gallery pages:

https://www.google.de/?q=site%3Ahttp%3A%2F%2Fmunin-gallery.dipohl.de%2F+redis

2. Search in the Munin projects gallery, which still contains outdated pages referring to your _former redis category_

https://www.google.de/?q=site%3Ahttp%3A%2F%2Fgallery.munin-monitoring.org+redis

You may set different categories (system, db, cache, ..) for the different graphs and this way the redis Plugin will show up in multiple categories in the Munin Plugin Gallery, each time significantly marked with header `redis`.

Please don't break the concept of a small collection of well-known categories, which should **not** be product names.

my2cents and thanks for your contribution!

Gabriele

PS: Concerning the users Munin overview: Here users can overwrite the default category per Munin Master configuration `munin.conf` to adapt fit the terms to her/his needs or taste.

denics:
Grazie Gabriele,
this was an useful explication. May I then suggest to add this very nice module to the contrib branch in official munin if @sumpfralle agrees? I've been testing for a while and it works fine.

dipohl:
Hi denics,

the decision about the plugin collection distributed by the Munin project lies in the hands of Steve Schnepp, the project lead. Recently I read [a statement of him that looks like your proposal will not be accepted](https://github.com/munin-monitoring/munin/issues/826) for branch stable-2.0.

Did you test the redis plugin already successfully with the development version of Munin? Then it might be possible that it can be added to the master branch of the Munin repository.

In any case a maintainer for the plugin will be needed then and at the time the redis plugin doesn't meet [the requirements for vetted plugins](http://munin-monitoring.org/wiki/requirements-vetted) as it has no perldoc section. 

On my wishlist: It would be nice if you add a doc page and example graphs for the visitors of the contrib collection in the Munin Gallery :-)

Cheers,
Gabriele

dipohl:
I added the explication to the [Munin Guide](http://guide.munin-monitoring.org/en/latest/reference/graph-category.html) now as well as the list of _well known categories_

sumpfralle:
Just for the cleanup: @denics - this pull request regarding the change of the category can be closed, right?

PR #844, opened 2017-05-29
Add missing appid in URL

mxjeff:
Add missing ``appid`` in http request and rename env to avoid collision with potential other api key.

### DIFF
--- a/plugins/weather/openweather_
+++ b/plugins/weather/openweather_
@@ -16,9 +16,10 @@
 ## From Oct 9 2015 OpenWeather needs you to register and get an APIKEY
 # include this key by setting 'env.apikey' in munin plugin config, i.e.:
 # [openweather_*]
-# env.apikey XYZ
+# env.owapikey XYZ
 
 query_string=$(printf '%s' "${0#*_}" | tr '_' '=')
+query_string="$query_string&appid=$owapikey"
 plugin_name=$( basename $0 )
 
 OWAPI=$( curl -s "http://api.openweathermap.org/data/2.5/weather?mode=xml&${query_string}")
--- a/plugins/weather/openweather_
+++ b/plugins/weather/openweather_
@@ -16,16 +16,13 @@
 ## From Oct 9 2015 OpenWeather needs you to register and get an APIKEY
 # include this key by setting 'env.apikey' in munin plugin config, i.e.:
 # [openweather_*]
-# env.owapikey XYZ
+# env.apikey XYZ
 
-query_string=$(printf '%s' "${0#*_}" | tr '_' '=')
-query_string="$query_string&appid=$owapikey"
+location=$(printf '%s' "${0#*_}" | tr '_' '=')
+query_string="${location}&appid=${apikey}&units=metric"
 plugin_name=$( basename $0 )
 
 OWAPI=$( curl -s "http://api.openweathermap.org/data/2.5/weather?mode=xml&${query_string}")
-# API returns temp in K, we have to convert it in C
-# &units=metric would change that ;-)
-KELVIN_BIAS=273
 
 CITY=$( expr "$OWAPI" : '.*\<city.*name=\"\(.*\)\"><coord.*' )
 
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for contributing to this plugin!

Sadly the previous usage of the `apikey` (that still remained in the documentation header) got lost recently (in January) in another pull request. Thus your submission is very welcome!

I have two additional thoughts / questions:

1. Maybe you would like to stick with the original name of the environment variable name `apikey`. Otherwise you would need to update the documentation header.
2. Maybe you could combine the two `query_string=...` lines into a single one for better readability? (optional)

Thank you!

mxjeff:
I made some others changes (mainly switch to metric unit system to get rid of kelvin offset).

But in the mean time I noticed there is an issue drawing the main graph.  
For some reason munin is looking for the wrong rdd file. I still have to figure out what's wrong (not even sure the plugin is responsible).

Somebody else testing the plugin would be very welcome :)

plugins/weather/openweather_
```diff
@@ -37,33 +35,30 @@ then
...
-		temp_max.cdef $KELVIN_BIAS,-
 
 		multigraph $plugin_name.humidity
 		graph_title Humidity in ${CITY}
 		graph_vlabel %
 		graph_category sensors
+		graph_args --upper-limit 100 --lower-limit 0
 		graph_info This graph show the humidity in ${CITY}
-		humidity.label humidity
+		humi.label humidity
```
sumpfralle: Please use an easily understandable (but still distinct) fieldname. Mabye `field_humidity`?

mxjeff: Agreed, idem with pressure field.

mxjeff:
Indeed, kelvin is a better choice. I followed existing comments without thinking about it.

Having some plugin conf to choose unit between metric, imperial and defaulting to kelvin would be even better (but it deserves another pull request imho).

mxjeff:
I did change back to kelvin+cdef but the [previous cdef config](https://github.com/munin-monitoring/contrib/blob/f2de5497bdcd94604048c4492739f6d83b55b470/plugins/weather/openweather_#L40) was triggering a ``rpn expressions without DEF or CDEF variables are not supported``. I'm not fluent enough in rdd, I let you judge my changes.

My settings: munin 2.0.25, rddtool  1.4.8,  Debian 8.8

plugins/weather/openweather_
```diff
@@ -37,33 +36,34 @@ then
...
+		temp_avg.cdef temp_avg,$KELVIN_BIAS,-
 
 		multigraph $plugin_name.temp
 		graph_title Temperature in ${CITY}
 		graph_vlabel Celsius
 		graph_category sensors
 		graph_info This graph show the temperature in ${CITY}
 		temp_avg.label avg
-		temp_avg.cdef $KELVIN_BIAS,-
+		temp_avg.cdef temp_avg,$KELVIN_BIAS,-
```
sumpfralle: I guess, the original cdef was ok? (`temp_avg.cdef $KELVIN_BIAS,-`)

sumpfralle:
Ping?

sumpfralle:
@mxjeff: any news regarding your proposal?

plugins/weather/openweather_
mxjeff: Hi,

I still have an error trying to graph : ``rpn expressions without DEF or CDEF variables are not supported``

I got it solved adding ``temp_avg``.

Setup:
  * Debian 9.4 vanilla
  * munin 2.0.33-1
  * rddtool 1.6.0-1

How could I debug further?
I'll try to reproduce with another set up from scratch.

Cheers and sorry for my delayed answer.

sumpfralle: After taking at a look at the `cdef` again, I think, your change was always on the right track - at least I cannot think of a good way, how the previous `cdef` was interpretet. Probably the `cdef` parser in munin was a bit more sloppy in an older release?

Anyway: if you are happy with the current state, I will merge it.
What do you think?

mxjeff: I'm ok with it.

I am not familiar with github workflow though.
Should I rebase my branch on top of master or is it ok like that ?

sumpfralle: Your changes still do not collide with the current master - thus they can be applied without rebase.

Thank you!

PR #846, opened 2017-06-12
[http_load_] Document running from cron with MUNIN_PLUGSTATE

shtrom:
When running from cron, nothing sets MUNIN_PLUGSTATE as needed, so it
needs to be done manually, otherwise the plugin cannot find its cache,
and no data is output.

### DIFF
--- a/plugins/http/http_load_
+++ b/plugins/http/http_load_
@@ -146,6 +146,13 @@ sub filter{
 	# status=1 => do download (default)
 	# status=0 => do not download
 
+	# For links, the 'rel' is more relevant that the 'src' attribute
+	if("$tag" =~ /^link/){
+		$status=0;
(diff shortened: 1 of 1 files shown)


shtrom:
Hum, CPAN error for a shell plugin. I don't thing the Travis error is linked.

steveschnepp:
A better way would be to leverage the `munin-run` tool that will happily set every variable needed for the plugin.

shtrom:
Ah, right. In my quick test, `munin-run` didn't work, but I must have been doing something wrong. It works nicely, with the oddity that you need to point it to any one of your `http_load_` plugins, and it will update everything.

While I'm at it, here are some other fixes.

shtrom:
Deprecated in favour of #848

PR #847, opened 2017-06-12
[ssl-certificate-expiry] Replace ssl_ with a single graph showing all expiries

shtrom:


plugins/ssl/multi_ssl
```diff
@@ -0,0 +1,75 @@
...
+    PORT="${1##*_}"
+    VAR="$(clean_fieldname "$1")"
+    if [ "$PORT" = "$SITE" ]; then
+	PORT=443
+    fi
+
+    CERT=$(echo "" | openssl s_client -CApath /etc/ssl/certs -servername "${SITE}" -connect "${SITE}:${PORT}" 2>/dev/null);
+
+    if [[ "${CERT}" = *"-----BEGIN CERTIFICATE-----"* ]]; then
+	echo "${CERT}" | openssl x509 -noout -enddate | awk -F= 'BEGIN { split("Jan Feb Mar Apr May Jun Jul Aug Sep Oct Nov Dec", month, " "); for (i=1; i<=12; i++) mdigit[month[i]] = i; } /notAfter/ { split($0,a,"="); split(a[2],b," "); split(b[3],time,":"); datetime=b[4] " " mdigit[b[1]] " " b[2] " " time[1] " " time[2] " " time[3]; days=(mktime(datetime)-systime())/86400; print "VAR.value " days; }' | sed "s/VAR/${VAR}/g"
```
steveschnepp: Multiline would be great for readability :)

plugins/ssl/multi_ssl
```diff
@@ -0,0 +1,75 @@
...
+
+: << =cut
+
+=head1 NAME
+
+multi_ssl - Plugin to monitor CERTificate expiration on multiple services and ports
+
+=head1 CONFIGURATION
+
+  [multi_ssl_*]
```
sumpfralle: It should be `multi_ssl`, or?

plugins/ssl/multi_ssl
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
sumpfralle: This plugin uses only a single bash-specific expression (`SITE="${1/_*/}"`). Thus maybe you want to avoid the dependency on bash and stick with `sh` instead?

plugins/ssl/multi_ssl
```diff
@@ -0,0 +1,75 @@
...
+: << =cut
+
+=head1 NAME
+
+multi_ssl - Plugin to monitor CERTificate expiration on multiple services and ports
+
+=head1 CONFIGURATION
+
+  [multi_ssl_*]
+    env.services www.service.tld blah.example.net_PORT
```
sumpfralle: Is there a specific reason why you picked the underscore as the separator between hostname and Port? A color would probably feel more familiar.

steveschnepp: I suppose he didn't know he could use a colon.

(also, i edited your comment as i think you made the typo : color/colon)

sumpfralle: Much better - thanks! :)

shtrom: No, the underscore is on purpose, as colons might be confusing with literal IPv6 addresses.

plugins/ssl/multi_ssl
shtrom: There are others, but ok.

shtrom:
Well, I didn't want to break the ssl plugin, but in all fairness, I think it could be retired, as the multi-ssl does a superset of that.

plugins/ssl/multi_ssl
steveschnepp: Ah, that infamous IPv6  😛 

But very true.

sumpfralle: With IPv6 addresses, I would expect the common approach of [wrapping the address in square brackets](https://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_network_resource_identifiers). At least `ssh` is doing this.
This would complicate the the port splitting slightly, but a simple `sed` expression to be applied to each token should solve this:
```
sed 's/^\(.\+\):\([0-9]\+\)$/\1 \2/'
```
But anyway: the decision is up to you.

sumpfralle:
You new awk statement looks really neat compared to the original!

Regarding the old plugin: I agree, that yours supersedes its functionality.
How about the following approach?
* add a simple wildcard parsing in case of an empty `services` environment variable (e.g. `[ -z "$services" ] && services=${0#ssl_}`) for backwards compatibility
* remove the old plugin
* rename your plugin to something like `ssl-certificate-expiry` (I think, we should use this opportunity for changing the name to be more precise)

@steveschnepp: or would you think, that this removal / replacement with a different name would break the expectations of the current users of that plugin?

shtrom:
Ah, if the RegExp is given to me... I just implemented that, and it works alright, except for IPv6:

    shtrom@supahwinch:/etc/munin/plugins$ sudo -u munin munin-run --pidebug multi_ssl_[2001:0DB8::badc:ofee]:485
    # ERROR: Invalid plugin name 'multi_ssl_[2001:0DB8::badc:ofee]:485.

Munin is happy with the colon, but apparently not so much with the square brackets...

Any suggestion to work around this? Are the square brackets that bad there, or could Munin be too protective?

sumpfralle:
What a pity!
I am sorry for putting you on the wrong track - here munin's restrictions on proper plugin names seem to stop us.

Your original underscore approach looks much better now! :)

plugins/ssl/ssl-certificate-expiry
```diff
@@ -56,12 +74,16 @@ esac
...
-    SITE="$(echo ${1} | sed 's/_.*//')"
-    PORT="$(echo ${1} | sed 's/.*_//')"
+    SITE="$(echo "${1}" | sed 's/_.*//')"
+    PORT="$(echo "${1}" | sed 's/.*_//')"
+
     VAR="$(clean_fieldname "$1")"
     if [ "$PORT" = "$SITE" ]; then
 	PORT=443
     fi
+    if echo "$SITE" | grep -q ':'; then
```
sumpfralle: Please add a comment that this is required for IPv6 addresses.

plugins/ssl/ssl-certificate-expiry
```diff
@@ -10,14 +10,25 @@ multi_ssl - Plugin to monitor CERTificate expiration on multiple services and po
...
 
 To set warning and critical levels do like this:
 
   [multi_ssl]
     env.services ...
     env.warning 30:
 
+Alternatively, if you want to monitor hosts separately, you can create multiple symlinks named as follows.
+
+  multi_ssl_HOST:PORT
```
sumpfralle: Underscore instead of colon.

plugins/ssl/ssl-certificate-expiry
```diff
@@ -10,14 +10,25 @@ multi_ssl - Plugin to monitor CERTificate expiration on multiple services and po
 =head1 CONFIGURATION
 
   [multi_ssl]
-    env.services www.service.tld blah.example.net_PORT
+    env.services www.service.tld blah.example.net:PORT
```
sumpfralle: underscore instead of colon

sumpfralle:
There were no objections regarding the retirement of the old `ssl_` plugin.

Thus could you please do the final cleanups?
* remove the `ssl_` plugin
* replace all occurrences of "multi_ssl" with "ssl-certificat-expiry"

This was a good refreshing of that old plugin!

sumpfralle:
Great!

PR #848, opened 2017-06-16
[http_load_] Add multigraph support

shtrom:
This branch is on top of the fix-http_load_ PR #846


sumpfralle:
How about switching to the multigraph plugin by default (without any underscore suffix) and renaming the plugin to `http_load`? This would reduce resource usage by calling the plugin only once (since most users are probably interested in all of its graphs).

sumpfralle:
sorry - somehow I failed to notice your commit - but you surely still remember it :)

Regarding the "suggest" mode: does it still work? I was not sure, when taking a look at the removed lines (around line 594). But even if it would not work: I think, loosing this capability would be a reasonable compromise for being able to use the plugin in a multigraph style.
Thus just for cleaning up: do you think it is OK like it is now?

Since you are turning this plugin into a beauty again: would you mind changing the documentation header in to the perldoc style (and remove the subversion marker (`$Id: $`)), please?

What do you think about renaming this plugin to `http_load` (without a traling underscore) in order to make users clearly aware of the multigraph approach)?

shtrom:
`suggest` still works, but it still suggest all the subgraphs. I'll fix that.

shtrom:
Ah, and removing the `_` would be confusing: it's multigraph per site, but the part after the underscore still needs to specifiy the site. It can't be used an its own.

shtrom:
Ok, it should be ready to go!

sumpfralle:
Great - thank you!

PR #849, opened 2017-06-17
tor/tor_ all-in-one plugin

daftaupe:
I've forked and worked a bit on https://github.com/mweinelt/munin-tor but now it's not maintained anymore so I'm commiting my work here.

Available from this plugin :
tor_bandwidth
tor_connections
tor_countries
tor_dormant
tor_flags
tor_routers
tor_traffic

Please comment if needed.

Regards,


plugins/tor/tor_
```diff
@@ -0,0 +1,491 @@
+#!/usr/bin/python2
+
```
sumpfralle: Please add a documentation header here. This will help future readers and allows a proper visualization in the munin plugin gallery.

Other plugins should provide usable examples (e.g. [munin/munin_events](https://github.com/munin-monitoring/contrib/blob/master/plugins/munin/munin_events)).

plugins/tor/tor_
```diff
@@ -0,0 +1,491 @@
...
+from __future__ import print_function
+import collections
+import json
+import os
+import stem
```
sumpfralle: Please group non-core modules (here: `stem`) separately (according to [PEP8](https://www.python.org/dev/peps/pep-0008/)).

Additionally you should mention this dependency in the documentation header.

Please also note, that missing external dependencies need to handled gracefully: the should be reported when running `autoconf`.

plugins/tor/tor_
```diff
@@ -0,0 +1,491 @@
...
+    try:
+        password = os.environ['torpassword']
+    except KeyError:
+        raise AuthError("Please configure the 'torpassword' "
+                        "environment variable")
+
+    try:
+        controller.authenticate(password=password)
+    except stem.connection.PasswordAuthFailed:
+        print("Authentication failed (incorrect password)")
```
sumpfralle: Should this be redirected to `sys.stderr`?

btw: wouldn't an `AuthError` be suitable here, as well?

plugins/tor/tor_
```diff
@@ -0,0 +1,491 @@
...
+        # header
+        for key, val in graph.iteritems():
+            print('graph_{} {}'.format(key, val))
+        # values
+        for label, attributes in labels.iteritems():
+            for key, val in attributes.iteritems():
+                print('{}.{} {}'.format(label, key, val))
+
+    @staticmethod
+    def autoconf():
```
sumpfralle: Here you should also report a potentially missing `stem` module.

plugins/tor/tor_
```diff
@@ -0,0 +1,491 @@
...
+
+class TorBandwidth(TorPlugin):
+    def __init__(self):
+        pass
+
+    def conf(self):
+        graph = {'title': 'Observed bandwidth',
+                 'args': '-l 0 --base 1000',
+                 'vlabel': 'bytes/s',
+                 'category': 'Tor',
```
sumpfralle: `Tor` is not a commonly known category. Please pick one from the [list of well-known categories](http://munin-monitoring.org/wiki/graph_category_list) instead.
I guess, `network` would be suitable, here.

plugins/tor/tor_
```diff
@@ -0,0 +1,491 @@
...
+                countries_num = json.load(f)
+        except:
+            # Fallback if cache_dir_name is not set, unreadable or any other
+            # error
+            countries_num = self.top_countries()
+
+        for c, v in countries_num:
+            print("%s.value %d" % (c, v))
+
+    # Unused
```
sumpfralle: Please remove this unused code, if it serves no further purpose.

plugins/tor/tor_
```diff
@@ -0,0 +1,491 @@
...
+                 'category': 'Tor',
+                 'info': 'Is Tor not building circuits because it is idle?'}
+        labels = {'dormant': {'label': 'dormant', 'min': 0, 'max': 1, 'type': 'GAUGE'}}
+
+        TorPlugin.conf_from_dict(graph, labels)
+
+    def fetch(self):
+        with gen_controller() as controller:
+            try:
+                #controller.authenticate()
```
sumpfralle: please remove

plugins/tor/tor_
```diff
@@ -0,0 +1,491 @@
...
+            provider = TorDormant()
+        elif __file__.endswith('_flags'):
+            provider = TorFlags()
+        elif __file__.endswith('_routers'):
+            provider = TorRouters()
+        elif __file__.endswith('_traffic'):
+            provider = TorTraffic()
+        else:
+            print('Unknown plugin name, try "suggest" for a list of possible ones.')
+            sys.exit()
```
sumpfralle: This message should be redirected to stderr and followed by a non-zero exitcode.

plugins/tor/tor_
```diff
@@ -0,0 +1,491 @@
...
+
+        if param == 'config':
+            provider.conf()
+        elif param == 'fetch':
+            provider.fetch()
+        else:
+            print('Unknown parameter "{}"'.format(param))
```
sumpfralle: This message should be redirected to stderr and followed by a non-zero exitcode.

daftaupe:
Hello @sumpfralle, thanks for the time you took to look at the code, I will try my best to fix it according to your suggestions.

sumpfralle:
I am glad, that you were not overwhelmed by the details :)

If you feel really adventurous, then you may also want to turn the plugin into a [multigraph plugin](http://guide.munin-monitoring.org/en/latest/plugin/multigraphing.html). This should make the plugin easier to use (multiple symlinks are not required), faster to run and could even reduce the code by a few lines. Anyway: proceed as you like ...

Have fun!

daftaupe:
Well, actually I think I'll first try to fix what you suggested as a first step. Converting the plugin to multigraph sounds good for a second step.
I'm not so fluent in Python, so I will need some time, the plugin has been started in this language by the original author, so I had to adapt :)

daftaupe:
@sumpfralle, actually there will be some delay, when I was trying to fix the autoconf return values I hit an issue : https://tor.stackexchange.com/questions/15072/tor-and-stem-library-non-consistent-error-message-with-wrong-password
That causes the script to not behave correctly if the password is wrong (not getting always the same answer) when running it with munin-run.

daftaupe:
Hello @sumpfralle, I did some changes in the code according to your suggestions.
Actually there's an issue with the stem library used in the plugin, so the autoconf part can't work well at the moment, please see there : https://trac.torproject.org/projects/tor/ticket/22679
When autoconf is run, the library returns non-consistent errors, which makes it unreliable at the moment. So the code hasn't changed to signal that the password is wrong.

plugins/tor/tor_
```diff
@@ -477,15 +519,16 @@ def main():
...
+            sys.exit(1)
 
         if param == 'config':
             provider.conf()
         elif param == 'fetch':
             provider.fetch()
         else:
-            print('Unknown parameter "{}"'.format(param))
+            print('Unknown parameter "{}"'.format(param), file=sys.stderr)
+	    sys.exit(1)
```
sumpfralle: Here a tab sneaked into your indentation?

plugins/tor/tor_
```diff
@@ -1,28 +1,78 @@
...
+daftaupe <daftaupe[at]protonmail.com>
+'''
 
 from __future__ import print_function
 import collections
 import json
 import os
+import sys
+
+import GeoIP
```
sumpfralle: Regarding the external dependencies: please use something along the following lines in order to support the usual "autoconf" approach of munin plugins:
```python
try:
    import GeoIP
    import stem
    import stem.control
    import stem.connection
except ImportError:
    # missing dependencies are reported via "autoconf" - thus failure is acceptable here
    pass
```

plugins/tor/tor_
```diff
@@ -88,6 +138,18 @@ class TorPlugin(object):
     @staticmethod
     def autoconf():
         try:
+            import stem
+
+        except ImportError as e:
+            print('no ({})'.format(e))
```
sumpfralle: A manually written error message is probably easier to understand for non-pythonic people.
Maybe the following?
```python
print('no (failed to import the required python module "stem": {})'.format(e))
```
```

daftaupe:
@sumpfralle I did some changes following your suggestions :)
The import trick is nice, I didn't think about it at all ! And the error message makes sense, the original one is not really explicit.

sumpfralle:
Regarding the ImportError message: you may want to apply the same pattern to "GeoIP" in "autoconf" :)

> Concerning the gallery, I will check what is needed.

That would be great. Basically it is just about adding example graphs (e.g. just taken from your setup) to the repository (following a specific filename schema). Anyway: this is just for polishing - I can also merge this plugin as it is right now.

> It's good you took the time to explain the things to me. I have two more plugins coming soon.

I am happy to hear that!
Have fun!

daftaupe:
> Regarding the ImportError message: you may want to apply the same pattern to "GeoIP" in "autoconf" :)
Sorry I did the commit too fast, didn't see the GeoIP part !

Concerning the pictures I have to add for the gallery I read that it should be in a folder called example-graphs, with a name like : myplugin_something-day.png for example.
But in this particular case my plugin has 7 plugins available, but some of them have the same name as other plugins in the directory. tor_routers and tor_traffic. How should I manage it ? Because I guess the pictures will end up in the gallery of these plugins, but their output might be different, even with the plugin having the same name.

sumpfralle:
Here the term `plugin` is used for a single file (not including its wildcard options).
Thus the gallery would look for pictures like `tor_-day.png` and `tor_-N.png` (with `N` being a digit).

You could pick one picture per plugin function and thus end up with `tor_-1.png` to `tor_-7.png`. This would probably be a nice visualization.

daftaupe:
Alright, now this should be complete :)

sumpfralle:
This looks great - thank you!

PR #850, opened 2017-06-26
Fix OpenWrt SNMP memory plugin

Dismounted:
Basically a rewrite of the existing plugin:
- Added documentation
- Updated to use Munin::Plugin::SNMP
- Fixed OID for snmpd on latest builds of OpenWrt/LEDE

### DIFF
--- a/plugins/openwrt/snmp__memory_openwrt
+++ b/plugins/openwrt/snmp__memory_openwrt
@@ -1,76 +1,81 @@
-#!/usr/bin/perl -w
-#
-# Copyright (C) 2008 J.M.Roth
-#
-# This program is free software; you can redistribute it and/or
-# modify it under the terms of the GNU General Public License
-# as published by the Free Software Foundation; version 2 dated June,
-# 1991.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program; if not, write to the Free Software
-# Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
-#
-#
-# $Log$
-#
-# Inspired by snmp__processes
-#
-#%# family=snmpauto
-#%# capabilities=snmpconf
+#!/usr/local/bin/perl -w
+# -*- perl -*-
+# vim: ft=perl
+
+=head1 NAME
+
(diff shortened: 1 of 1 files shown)


plugins/openwrt/snmp__memory_openwrt
```diff
@@ -1,76 +1,81 @@
...
-if ($host eq '' || !defined $host)
-{
-	print "# Debug: $0 -- $1\n" if $DEBUG;
-	die "# Error: couldn't understand what I'm supposed to monitor.";
-}
+As a rule SNMP plugins need site specific configuration.  The default
+configuration (shown here) will only work on insecure sites/devices.
+
+   [snmp_*]
+	env.version 2
```
sumpfralle: I think, you mixed tabs/spaces here.

plugins/openwrt/snmp__memory_openwrt
```diff
@@ -1,76 +1,81 @@
...
-# Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
-#
-#
-# $Log$
-#
-# Inspired by snmp__processes
-#
-#%# family=snmpauto
-#%# capabilities=snmpconf
+#!/usr/local/bin/perl -w
```
sumpfralle: I am not aware of the specific situation of perl in openwrt, but I guess, it is installed in `/usr/bin/perl`?
Or is there a good reason to change this path?

plugins/openwrt/snmp__memory_openwrt
```diff
@@ -1,76 +1,81 @@
...
+reports the contents of the hrStorageSize and hrStorageUsed OIDs.
+
+=head1 MAGIC MARKERS
+
+  #%# family=snmpauto
+  #%# capabilities=snmpconf
+
+=head1 VERSION
+
+  $Id$
```
sumpfralle: I think, you should put a specifc version here - this is probably a remainder of former svn usage.

plugins/openwrt/snmp__memory_openwrt
Dismounted: Thanks, will fix.

plugins/openwrt/snmp__memory_openwrt
Dismounted: I had a look at the newest SNMP plugins included in base munin - they no longer contain this item, so I will do the same.

plugins/openwrt/snmp__memory_openwrt
Dismounted: Missed this - yep, I have a slightly different environment. Will add a fix for portability.

Dismounted:
@sumpfralle addressed all of you comments - thanks!

sumpfralle:
Great - thank you!

PR #851, opened 2017-06-27
Ethereum plugins

Cyclenerd:
Hi,

I wrote two plugins for cryptocurrency Ethereum (ETH). With this plugins you can monitor your miner and balances.

Best regards
Nils


plugins/ethereum/ethermine_
```diff
@@ -0,0 +1,96 @@
...
+
+MINER = OPTIONS[2]
+
+if MINER == "":
+    print "Rig name missing!"
+    sys.exit(9)
+
+if command == 'config':
+    print "graph_title Ethermine " + MINER
+    print "graph_info Ethermine Hashrate " + ETH_ADDRESS +"/" + MINER
```
sumpfralle: Please avoid string concatenation and use the more pythonic style via "format" or the "%" operator:
```python
print("graph_info Ethermine Hashrate {}/{}".format(ETH_ADDRESS, MINER))
```

plugins/ethereum/ethermine_
```diff
@@ -0,0 +1,96 @@
...
+    print ETH_ADDRESS + "_" + MINER + ".critical 10:"
+    print ETH_ADDRESS + "_" + MINER + ".label MH/s"
+    sys.exit(0)
+
+
+POOL_URL = 'https://ethermine.org/api/miner_new'
+STATS = POOL_URL + '/' + ETH_ADDRESS
+
+mining_req = urllib2.Request(STATS)
+mining_req.add_header('User-Agent', 'Mozilla/5.0')
```
sumpfralle: Does the API force you to hide your user-agent? Otherwise I would suggest to keep the technically correct user-agent.

plugins/ethereum/ethermine_
```diff
@@ -0,0 +1,96 @@
...
+    print ETH_ADDRESS + "_" + MINER + ".warning 20:"
+    print ETH_ADDRESS + "_" + MINER + ".critical 10:"
+    print ETH_ADDRESS + "_" + MINER + ".label MH/s"
+    sys.exit(0)
+
+
+POOL_URL = 'https://ethermine.org/api/miner_new'
+STATS = POOL_URL + '/' + ETH_ADDRESS
+
+mining_req = urllib2.Request(STATS)
```
sumpfralle: `urllib2` is a python2-only module.
Please take care that this module works with python3.
Thus if you insist on preserving python2-compatibiltity (even though I cannot think of a good reason for this wish), then I would recommend the following:
```python
try:
    # python3
    from urllib.request import urlopen
except ImportError:
    # python2
    from urllib2 import urlopen
```

Regarding this specific line (`mining_req = urllib2.Request(STATS)`) - you do not need this, since you can use the URL for all requests. Otherwise you would need another conditional import of `urllib2.Request` or `urllib.requests.Request` for python3 compatibility.

plugins/ethereum/ethermine_
```diff
@@ -0,0 +1,96 @@
...
+mining_req.add_header('User-Agent', 'Mozilla/5.0')
+
+try:
+    mining_stats_raw = urllib2.urlopen(mining_req, None, 1.5 )
+except urllib2.HTTPError as e:
```
sumpfralle: For clarity I would recommend to handle all exceptions at once:
```python
except IOError as exc:
    print("Failed to request statistics: {}".format(exc))
```

plugins/ethereum/ethermine_
```diff
@@ -0,0 +1,96 @@
...
+except urllib2.URLError:
+    print "HTTP URL Error!"
+    sys.exit(9)
+except socket.timeout:
+    print "HTTP Timed out!"
+    sys.exit(9)
+
+try:
+    mining_stats = json.load(mining_stats_raw)
+except:
```
sumpfralle: Please do not use bare except clauses. Here `ValueError` should be fine.

plugins/ethereum/ethermine_
```diff
@@ -0,0 +1,96 @@
...
+    sys.exit(9)
+
+workers = mining_stats['workers']
+
+# ethermine.org sometimes has caching errors. You can see data from other miner. Always check your rig name.
+for worker in workers:
+    if workers[worker]['worker'] == MINER:
+        hash_rate = workers[worker]['hashrate']
+        hash_rate = hash_rate.replace(" MH/s", "")
+        print ETH_ADDRESS + "_" + MINER + ".value %s " % hash_rate
```
sumpfralle: Please use string formatting.

plugins/ethereum/ethermine_
```diff
@@ -0,0 +1,96 @@
...
+
+
+POOL_URL = 'https://ethermine.org/api/miner_new'
+STATS = POOL_URL + '/' + ETH_ADDRESS
+
+mining_req = urllib2.Request(STATS)
+mining_req.add_header('User-Agent', 'Mozilla/5.0')
+
+try:
+    mining_stats_raw = urllib2.urlopen(mining_req, None, 1.5 )
```
sumpfralle: Please use keyword arguments explicitly and do not rely on their order. Ths improves the readability of the code for your fellow readers.
```python
mining_stats_raw = urllib2.urlopen(mining_req, timeout=1.5 )
```

plugins/ethereum/ethermine_
```diff
@@ -0,0 +1,96 @@
...
+    print "graph_vlabel Ethermine Hashrate"
+    print "graph_category htc"
+    print ETH_ADDRESS + "_" + MINER + ".warning 20:"
+    print ETH_ADDRESS + "_" + MINER + ".critical 10:"
+    print ETH_ADDRESS + "_" + MINER + ".label MH/s"
+    sys.exit(0)
+
+
+POOL_URL = 'https://ethermine.org/api/miner_new'
+STATS = POOL_URL + '/' + ETH_ADDRESS
```
sumpfralle: Please use a more descriptivate variable name (e.g. "STATS_URL" - or maybe you can think of an even more precise name?).

plugins/ethereum/etherscan_balance_
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python
```
sumpfralle: Most of my comments above apply to this plugin as well, since they share a lot of code.
Maybe you want to combine both plugins into a [multigraph plugin](http://guide.munin-monitoring.org/en/latest/plugin/multigraphing.html)?
Just put a line like the following in front of every graph (the rest can remain unchanged):
```
multigraph ethermine_balance_0x3257bde8cf067ae6f1ddc0e4b140fe02e3c5e44f
```
Or maybe use a wildcard plugin?
We want to avoid duplicate code in the repository, since it is hard to maintain.

Cyclenerd:
Uff, thanks for the detailed answer. I have a lot to do :-) I am not really a Python programmer and copied everything from another plugins.

sumpfralle:
I glad, that you were not overwhelmed by the details! :)
To keep things simple, I would recommend, that you just switch to python3 (instead of keeping the code python2/3 compatible). This requires probably just three steps:
* change the shebang
* replace `print foo` with `print(foo)` or `print(foo, file=sys.stderr)` (for error output) - you should forget my comment with `from __future__...`
* use `import urllib.requests` and `urllib.requests.urlopen` instead of `import urllib2` and `urllib2.urlopen`

You may want to use "flake8" (`python3 -m flake8 YOUR_PLUGIN`) for style checks.

Have fun!

Cyclenerd:
I have the two plugin revised. Thanks again for your hints.

* I do not want to enter the Ethereum address over the configuration file. Via a system link is much better. The spelling of the address may differ from the different providers. No Multigraph. balances and hashrates of differed addresses can vary very much.
* I have created a new folder `currency`. Here is the folder `ethereum`. The plugin filenames I have also reworked. In this folder you could also move the `bitcoin` folder. If there are plugins for paper money, they can also be stored in the folder.
* I tried to keep all Python2 and Python3 compatible. Tests were successful.
* POD added

plugins/currency/ethereum/ethermine_hashrate_
```diff
@@ -0,0 +1,112 @@
...
+ethermine_api_url = 'https://ethermine.org/api/miner_new/' + eth_address
+
+mining_req = Request(ethermine_api_url)
+# User-Agent to bypass Cloudflare
+mining_req.add_header('User-Agent', 'Ethermine Munin Plugin/1.0')
+
+try:
+    mining_stats_raw = urlopen(mining_req, timeout=15)
+except IOError as exc:
+    print("Failed to request ethermine.org API: {}".format(exc), file=sys.stderr)
```
sumpfralle: Shouldn't you exit with an error, here?

plugins/currency/ethereum/ethermine_hashrate_
```diff
@@ -0,0 +1,112 @@
...
+except ValueError:
+    print("The filename of this plugin (or its symlink) should follow this pattern: "
+          "'ethermine_hashrate_<YOUR_PUBLIC_ETHEREUM_ADDRESS>_<YOUR_RIG_NAME>'", file=sys.stderr)
+    sys.exit(9)
+
+if command == 'config':
+    print("graph_title Ethermine {}".format(miner))
+    print("graph_info ethermine.org Mining Pool Hashrate for {}_{}".format(eth_address, miner))
+    print("graph_vlabel Ethermine Hashrate")
+    print("graph_category htc")
```
sumpfralle: Please pick a category name from the list of [well known categories](http://munin-monitoring.org/wiki/graph_category_list). I guess, `other` is (for now) the most suitable.

plugins/currency/ethereum/ethermine_hashrate_
```diff
@@ -0,0 +1,112 @@
...
+    mining_stats_raw = urlopen(mining_req, timeout=15)
+except IOError as exc:
+    print("Failed to request ethermine.org API: {}".format(exc), file=sys.stderr)
+
+reader = codecs.getreader("utf-8")
+
+try:
+    mining_stats = json.load(reader(mining_stats_raw))
+except ValueError:
+    print("Failed to parse JSON responce.", file=sys.stderr)
```
sumpfralle: "response"

plugins/currency/ethereum/ethermine_hashrate_
```diff
@@ -0,0 +1,112 @@
...
+    sys.exit(9)
+
+try:
+    workers = mining_stats['workers']
+except:
```
sumpfralle: Please avoid bare excepts. Here `KeyError` should be the right choice.

plugins/currency/ethereum/etherscan_balance_
```diff
@@ -0,0 +1,106 @@
...
+    print("Failed to parse JSON responce.", file=sys.stderr)
+    sys.exit(9)
+
+try:
+    float(etherscan_balance['result'])
+except:
+    print("JSON result error!", file=sys.stderr)
+    sys.exit(9)
+
+eth = float(etherscan_balance['result']) / 1000000000000000000
```
sumpfralle: What is the purpose of this division?
Aren't the iso multipliers automatically generated by munin (kilo, mega, giga, ...) sufficient?

Cyclenerd:
Thanks again for your help 👍 

I'd like to keep two files. Both have something to do with Ethereum, but use different API providers.

* `etherscan_balance_`: Is about credit balances. This will continue in the long term.
* `ethermine_hashrate_`: Is for Mining. Currently this is done with graphics cards. Due to the hype of Bitcoin, it is very popular at the moment. But this will not always work. Sometime it is not worth it anymore and Ethereum also wants to convert the verification of blocks. Blocks are then no longer verified with computing power (graphics cards), but with credit. It is assumed that it is introduced 2018.

The majority will only load `etherscan_balance_`.

With the ValueError I need your help. I checked it. If no JSON can be processed (website delivers HTML), this is output:

```
$ python etherscan_balance_0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d

  File "etherscan_balance_0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d", line 95, in <module>
    etherscan_balance = json.load(reader(etherscan_balance_raw))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 290, in load
    **kw)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/json/decoder.py", line 384, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
```

That's what I've found: https://stackoverflow.com/questions/8381193/python-handle-json-decode-error-when-nothing-returned

Is ValueError OK?

The division I have to do, because the API does not return a float. I want to output ETH.

sumpfralle:
ValueError and JSON: in your specific case, it is a good choice. You `try` clause consists of an decoding operation (which may throw a child of ValueError) and the JSON parsing (which could also throw a child of ValueError). You could add a comment referring these two failure causes to the exception branch.

Regarding the division: please add a comment, why you need to divide by this very specific number. It is still not clear to me :)

Besides this, I think, it is ready.

Cyclenerd:
I added the comments.

More details. If interested: http://ethdocs.org/en/latest/ether.html

> The smallest denomination aka base unit of ether is called Wei.

sumpfralle:
Ah - ok - thank you for the link.
In this case I would suggest do store the values in their natural unit (wei) and visualize the result as ether.

You would do this by printing the wei value (remove the division) and adding the following line to the config:
```python
print("{0}_{1}.cdef {0}_{1},1000000000000000000,/".format(eth_address, miner))
```
This would allow later changes of the output unit without complicated conversions and losses of historical data.

Or you can just leave it like it is - then I will merge it now.

Cyclenerd:
With `{fieldname}.cdef` it is much better... 👍 

```
python etherscan_balance_0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d config
graph_title ETH 0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d
graph_info Ethereum Account Balance for Address 0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d
graph_vlabel Ethereum Balance
graph_category other
0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d.cdef 0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d,1000000000000000000,/
0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d.label ETH
```

But I have now problems:

```
2017/07/01 16:10:37 [RRD ERROR] Unable to graph /var/vhosts/[...]/etherscan_balance_0xfbbc9f870bccadf8847eba2
9b0ed3755e30c9f0d-day.png : rpn expressions without DEF or CDEF variables are not supported
```
...
```
'DEF:a9cc724ae0b560b3=/var/lib/munin/[...]/[...].com-etherscan_balance_0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d-_xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d-g.rrd:42:MAX' \
'DEF:i9cc724ae0b560b3=/var/lib/munin/[...]/[...].com-etherscan_balance_0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d-_xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d-g.rrd:42:MIN' \
'DEF:g9cc724ae0b560b3=/var/lib/munin/[...]/[...].com-etherscan_balance_0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d-_xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d-g.rrd:42:AVERAGE' \
'CDEF:a58645e2bfd774b2=0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d,1000000000000000000,/' \
'CDEF:i58645e2bfd774b2=0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d,1000000000000000000,/' \
'CDEF:g58645e2bfd774b2=0xfbbc9f870bccadf8847eba29b0ed3755e30c9f0d,1000000000000000000,/' \
'CDEF:c58645e2bfd774b2=g58645e2bfd774b2' \
```

Servername removed with [...]

sumpfralle:
Ah - the leading zero seems to confuse the cdef parser, or?

How about adding a prefix to the fieldname?

Or: (since you want to use only a single dataset per diagram, anyway) you just pick a trivial name like "hashrate" or "balance" instead of the full ethereum token.

Cyclenerd:
Now everything works :-)

It was really the field name. Thanks for the tip 👍 

If you do not have any comments and improvements, I would have finished my work.

sumpfralle:
Cool - thank you!

Issue #852, opened 2017-06-27
Travis build broken by old Sys-Virt-0.9.8


Cyclenerd:
Sys-Virt-0.9.8 does not exist anymore. Oldest version is 1.3.0.

```
--> Working on DANBERR/Sys-Virt-0.9.8.tar.gz
Fetching http://www.cpan.org/authors/id/D/DA/DANBERR/Sys-Virt-0.9.8.tar.gz ... FAIL
! Download http://www.cpan.org/authors/id/D/DA/DANBERR/Sys-Virt-0.9.8.tar.gz failed. Retrying ... 
! Download http://www.cpan.org/authors/id/D/DA/DANBERR/Sys-Virt-0.9.8.tar.gz failed. Retrying ... 
! Download http://www.cpan.org/authors/id/D/DA/DANBERR/Sys-Virt-0.9.8.tar.gz failed. Retrying ... 
! Failed to download http://www.cpan.org/authors/id/D/DA/DANBERR/Sys-Virt-0.9.8.tar.gz
! Failed to fetch distribution Sys-Virt-0.9.8
The command "cpanm --notest DANBERR/Sys-Virt-0.9.8.tar.gz" failed and exited with 1 during .
```
On the Travis server should be `libvirt-dev` 1.2.22 (https://packages.ubuntu.com/trusty/libvirt-dev).

I immediately make a pull request to fix but can not test it.

daftaupe:
Should this be worked on ?
@Cyclenerd, I see in the pull request that you did upload Sys-Virt on your server, but it seems that the link is dead now.
It seems it can be found there : http://lutung.lib.ums.ac.id/cpan/authors/id/D/DA/DANBERR/Sys-Virt-0.9.8.tar.gz

Cyclenerd:
Hi,

https://www.nkn-it.de/backup/sys-virt/Sys-Virt-0.9.8.tar.gz is not deleted. I can also load the file from there.

daftaupe:
Weird, it didn't work earlier for me. 
Anyway, I just show that the build didn't complete on my last PR and I saw this was fixing the thing but wasn't closed.
What's blocking then ?

sumpfralle:
Fixed in 3706fe0051ffa81305938bbfa63df70859338efc

PR #853, opened 2017-06-27
Fix broken Travis build Issue #852

Cyclenerd:
Fix issue https://github.com/munin-monitoring/contrib/issues/852
Sys-Virt-0.9.8 does not exist anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: perl
 install:
   - sudo apt-get update
   - sudo apt-get --no-install-recommends install devscripts python python3 ruby php5-cli gawk ksh zsh pylint
-  - sudo apt-get --no-install-recommends install pkg-config libdb-dev libvirt-dev libsys-virt-perl libexpat-dev
+  - sudo apt-get --no-install-recommends install pkg-config libdb-dev libvirt-dev libexpat-dev
   # - Munin/Plugin.pm is in "munin-node" on precise
   - sudo apt-get --no-install-recommends install munin-node
 
@@ -53,7 +53,8 @@ install:
   - cpanm --notest Switch
   - cpanm --notest nvidia::ml
   - cpanm --notest experimental
-  - cpanm --notest Sys::Virt
+  # Old Sys-Virt version
+  - cpanm --notest https://www.nkn-it.de/backup/sys-virt/Sys-Virt-0.9.8.tar.gz
   # Modules used by plugins, but missing on cpan
   # - File::Tail::Multi
   # - Sun::Solaris::Kstat


Cyclenerd:
Travis works again. Found an old Sys-Virt-0.9.8 version in an archive and put the file on my server.

https://travis-ci.org/Cyclenerd/contrib/builds/247649729

You could also use the version of Ubuntu (the attempts you see in my Commits). Perl seems to be looking at other places.

sumpfralle:
Thank you for tackling this issue!

Since you probably managed to get a better understanding of the problem, than I have, maybe you can share some details?
* does the plugin `libvirt/libvirt` work with the newest `Sys::Virt` package? (thus: could we pick the one from Debian/Ubuntu?)
* do you know, whyt the `Switch` was not missing before?
* is the explicit selection of `trusty` required? Do you know why?
Thank you!

Cyclenerd:
`libvirt` does not work with the newest CPAN `Sys::Virt` 👎 

Switch was missed by `redis_` (https://travis-ci.org/Cyclenerd/contrib/builds/247642151)

```
# Can't locate Switch.pm in @INC (@INC contains: /usr/share/perl5
 /home/travis/perl5/perlbrew/perls/5.14/lib/site_perl/5.14.4/x86_64-linux 
 /home/travis/perl5/perlbrew/perls/5.14/lib/site_perl/5.14.4 
 /home/travis/perl5/perlbrew/perls/5.14/lib/5.14.4/x86_64-linux 
 /home/travis/perl5/perlbrew/perls/5.14/lib/5.14.4 .) at redis_ line 39.
```

With `trusty` you get Ubuntu 14.04 and no longer the EOF 12.04.

I think the best would be to use the Ubuntu Perl `libsys-virt-perl` package. Then the versions should fit. I tried that: https://travis-ci.org/Cyclenerd/contrib/builds/247645291

The perlbrew could not find the module but then. Maybe you have to expand the @INC. I do not know where Ubuntu stores Perl modules.

sumpfralle:
I followed you suggestion and tried to get the `libsys-virt-perl` package working within the current setup. But sadly the perl build used by travis (enabled via `language: perl`) was built without threading support. But the `libsys-virt-perl` packages was built with threading support.
Thus the `@INC` search path turned out to be not the real problem :)

I prepared [another pull request](contrib#854) and experimented a bit until I seemed to have found a suitable approach. Would you mind taking a look, since you already gained some experience with your experiments? :)

Cyclenerd:
I also tried something around. But do not come to any other solution than you.

Travis also has perlbrew threading support https://docs.travis-ci.com/user/languages/perl/#Perl-runtimes-with--Duseshrplib

I have tried it with `5.18-shrplib`. The problem is then that it will be version 5.18.4 and not 5.18.2 (like Ubuntu).

```
Perl lib version (5.18.2) doesn't match executable '/home/travis/perl5/perlbrew/perls/5.18-shrplib/bin/prove' version (v5.18.4)
```

Here you can see my experiments.`travisperl` branche: https://travis-ci.org/Cyclenerd/contrib/builds

What confuses me a little, that now more test errors come: https://travis-ci.org/munin-monitoring/contrib/builds/248980817#L1882
```
# syntax error at snmp__netscaler_connections line 151, near "$field qw(client server ssl)"
# Global symbol "$field" requires explicit package name at snmp__netscaler_connections line 152.
# syntax error at snmp__netscaler_connections line 155, near "}"
# snmp__netscaler_connections had compilation errors.
```

With the old Sys::Virt (this pull), the error does not occur: https://travis-ci.org/munin-monitoring/contrib/builds/247649751#L1500

**In summary:** I would completely change to the Perl version of Ubuntu. Like in your pull.

sumpfralle:
hehe - so I was not the only one triggering dozens of test runs :)

I agree, that we should probably not put too much effort into getting stuff to run with the exotic mixture of build variations offered by travis. Thus I will merge my branch, as you suggested. Thank you for taking a look and sharing your results!

The "new" failures indicated by the test runs are very confusing for me, too.

Since a local run of `prove` also exposes these errors, I assume that they were always there, but somewhere hidden/skipped due to the homebrew travis environment. My perl knowledge is a bit too weak to understand the cause, but I guess, we can safely proceed without breaking more than before.

I will take a look at the `checkbashisms` checks.

Do you have an opinion regarding the reported perl errors?

Cyclenerd:
The new tests with your pull are better!

I have looked at an error in detail. And it's true. `snmp__netscaler_connections`is broken. Since the brackets are missing:

https://github.com/munin-monitoring/contrib/blob/master/plugins/netscaler/snmp__netscaler_connections#L151

To test it:

```
#!/usr/bin/perl

use strict;
use warnings;

# BEST
my @blas = qw(client server ssl);
for my $bla (@blas) {
	print "$bla\n";
}

# OK
for my $bla (qw(client server ssl)) {
	print "$bla\n";
}

# FAIL
# syntax error at test.pl line 19, near "$bla qw(client server ssl)"
for my $bla qw(client server ssl) {
	print "$bla\n";
}
```

sumpfralle:
I finished the other set of changes now and mered them. The tests are working again ...

Thank you for preparing this!

PR #858, opened 2017-07-06
New plugin to get result data of Nanopool ethereum mining pool.

geschke:
Hello!

I've added a plugin to monitor the most important result data of Nanopool ethereum mining. It shows the hashrate, average hashrate, total balance and the worker data with their hashrate and number of shares per minute. 

Thanks in advance & kind regards,
   Ralf

### DIFF
--- a/plugins/currency/nanopool/nanopool_
+++ b/plugins/currency/nanopool/nanopool_
@@ -157,7 +157,7 @@ def write_config_worker():
     for i, val in enumerate(worker_data):
         print("worker" + str(i) + "_shares.label " + str(val["id"]))
         print("worker" + str(i) + "_shares.type COUNTER")
-        print("worker" + str(i) + "_shares.info Rating of worker '" + str(val["id"]) + "'")
+        print("worker" + str(i) + "_shares.info Accepted shares of worker '" + str(val["id"]) + "'")
         print("worker" + str(i) + "_shares.min 0")
         print("worker" + str(i) + "_shares.draw LINE1")
 
--- a/plugins/currency/nanopool/nanopool_
+++ b/plugins/currency/nanopool/nanopool_
@@ -2,14 +2,36 @@
 # -*- coding: utf-8 -*-
 # vim: set fileencoding=utf-8
 
-
 """
 =head1 NAME
 
-nanopool_ - Munin plugin to monitor nanopool eth user data.
+nanopool_ - Munin plugin to monitor nanopool ethereum user data.
+
+
+=head1 CONFIGURATION
+
+[nanopool_*]
+
+        - Copy to /usr/share/munin/plugins
+        - Create symlinks in /etc/munin/plugins to nanopool_<graph type>_<account_address>,
+          e.g. ln -s /usr/share/munin/plugins/nanopool_ /etc/munin/plugins/nanopool_hashrate_0x1234567890abcdef1234567890
+          Please use the account address in the format "0x<address>".
+
+        - To enable graphs, link to one or more of the graph types available
+        - Then restart munin-node
+
+Graph types and their links:
+        balance: link to nanopool_balance           Show current balance
+        hashrate: link to nanopool_hashrate         Show current calculated hashrate
+        avghashrate: link to nanopool_avghashrate   Show average hashrate of last hour
+        worker: link to nanopool_worker             Show worker data
+
+
+Thanks to the authors of other Python munin plugins. I've used some of
+them as inspiring example.
 
 =head1 VERSION
-0.0.1
+1.0.0
 
 =head1 AUTHOR
 L<Ralf Geschke|https://github.com/geschke>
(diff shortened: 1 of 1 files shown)


sumpfralle:
Hello Ralf,

thank you for your contribution!

Just a few days ago @Cyclenerd proposed two plugins that which looked (initially) quite similar to yours: #851.
Could you please take a look at them? Maybe they can be made configurable in order to support both pools (if that is the difference)?
(Code duplication is something that we really want to avoid due to the future maintenance burden)

Cheers,
Lars

geschke:
Hello Lars!

Thanks for your comments! I agree with you that code duplicating isn't a good idea. It seems that @Cyclenerd and I did develop something in parallel, so I was glad to find an appropriate folder to place the new plugin. ;-) 
I think the main difference is the API. The API of Nanopool (https://eth.nanopool.org/api) has a good documentation and let you request nearly anything you want. I didn't found a similar approach at Ethermine, only a link to JSON API in the details page (e.g. https://etc.ethermine.org/miners/05feb12f23f257229e7869c914808fc523bc3006 -> JSON API). The request shows very much information at one time. 
In my opinion it would be more effort to support both APIs with one plugin than having two (or more) plugins to support the different APIs. A generic plugin would end in many decision statements to differ between the pools. (Or you have to develop an abstraction layer to get data and transform into another format to feed Munin...) For me this seems to be more complicated than simple plugins - I would compare this approach to have one (or more) plugins per, let's say, web server. There are Nginx plugins, Apache plugins, Lighttpd plugins and so on, all of them have the same kind of server as their data source, but with different characteristics.
(I'm not sure about the future development of Munin and it's plugins, maybe you plan a more layered structure, so that not every plugin has to integrate all steps from getting data to write the values.)

Thanks for reading!
Kind regards,
   Ralf

sumpfralle:
Hello Ralf,

OK - I suspected, that you would want your own plugin :)

I have the following comments:
* in the "configuration parameters" section: I think, it would help people that are unfamiliar with munin, if you could add `[nanopool_*]` above the configuration line.
* `raise Exception` is rather unusual and will output a shocking amount of details to users - maybe you just want `print(..., file=sys.stderr)` followed by `sys.exit(1)`?
* around line 141 and 158 you used string concatenation. This is fairly uncommon with python and is hard to read in this case. Maybe you want to use one of the string formatting approaches (` % ...` or `"...".format(foo, bar)`)
* the single `print` statement without braces (e.g. in line 172) will not work as expected in python3 - just add `()`
* please add a final `else` clause (with an error message) at the end of `write_data`
* at the end you use structures like `len(sys.argv) == 2`. I am not sure, if this is a proper condition for all arguments given to a munin plugin. You should be future-proof and only check for `len(sys.argv) > 1`.
* why do you specify `graph_width` with its default value? If it is not really necessary, then you should not explicitly overwrite it.

Regarding the numbering of `worker_data`:
You are using the index of the `worker_data` list as part of your fieldnames.
This is a bad idea, if the sorting of the input source is not guaranteed to be stable.
Thus you may either want to switch to a specific detail of each list item (e.g. a name) or apply `.sort` to the `worker_data` (based on a suitable key) before.

Please also read the other pull request. SInce your code-base is very similiar, some comments should apply here, too.

geschke:
Thanks for your comments, I will check these issues! 

Kind regards,
   Ralf

plugins/currency/nanopool/nanopool_
```diff
@@ -0,0 +1,290 @@
...
+def request_data():
+
+    url = "https://api.nanopool.org/v1/eth/user/{0}".format(account_address)
+
+    req = Request(url)
+    # this fake is necessary to get values, otherwise the request ends in a 403 error
+    req.add_header('User-Agent', 'Nozilla/5.0')
+    try:
+        txt = urlopen(req).read()
+    except Exception as err:
```
sumpfralle: Please avoid generic exception traps.

Here you could either use `URLError` and `HTTPError` (with their respective imports for python2 and python3) or you could simpliy pick `OSError` (which is the parent of both of these exceptions).

plugins/currency/nanopool/nanopool_
```diff
@@ -0,0 +1,290 @@
...
+        exit(1)
+
+    return json.loads(txt.decode("utf-8"))
+
+
+def write_config_worker():
+    data = request_data()
+    worker_data = sorted(data["data"]["workers"], key=lambda element: element["id"])
+
+    print("multigraph worker_hashrate_{0}".format(account_address))
```
sumpfralle: For simplicity you can also use `{}` instead of `{0]` - but this is just a matter of taste. There is no need to change this.

plugins/currency/nanopool/nanopool_
```diff
@@ -0,0 +1,290 @@
...
+        worker_name = "_".join(val["id"].split())
+        print("multigraph worker_hashrate_{0}.worker_{1}".format(account_address, worker_name))
+        print("graph_title Hashrate in Mh/s of worker {0}".format(worker_name))
+        print("graph_args --base 1000 -l 0")
+        print("graph_vlabel Mh/s")
+        print("graph_category other")
+        print("graph_scale no")
+        print("whashrate.label hashrate")
+        print("whashrate.type GAUGE")
+        print("whashrate.info Hashrate of worker {0}".format(str(val["id"])))
```
sumpfralle: FYI: the parameters of the `format` function are always implicitly converted to their target data type (here: string). Thus you can remove `str`.

plugins/currency/nanopool/nanopool_
```diff
@@ -0,0 +1,290 @@
...
+    print("")
+    print("multigraph worker_shares_{0}".format(account_address))
+    print("graph_title Number of accepted shares ({0})".format(account_address))
+    print("graph_args --base 1000 -l 0")
+    print("graph_vlabel Shares per ${graph_period}")
+    print("graph_category other")
+    print("graph_scale no")
+    print("graph_period minute")
+
+    for i, val in enumerate(worker_data):
```
sumpfralle: Here you do not need `i` anymore, or?

plugins/currency/nanopool/nanopool_
```diff
@@ -0,0 +1,290 @@
...
+        print("wshares.min 0")
+        print("wshares.draw LINE1")
+
+
+def write_data_worker(data):
+    worker_data = sorted(data["data"]["workers"], key=lambda element: element["id"])
+
+    print("multigraph worker_hashrate_{0}".format(account_address))
+
+    for i, val in enumerate(worker_data):
```
sumpfralle: Remove `i`?

plugins/currency/nanopool/nanopool_
```diff
@@ -0,0 +1,290 @@
...
+
+    for val in worker_data:
+        print("")
+        worker_name = "_".join(val["id"].split())
+        print("multigraph worker_hashrate_{0}.worker_{1}".format(account_address, worker_name))
+        print("whashrate.value {0}".format(str(val["hashrate"])))
+
+    print("")
+    print("multigraph worker_shares_{0}".format(account_address))
+    for i, val in enumerate(worker_data):
```
sumpfralle: Remove `i`?

plugins/currency/nanopool/nanopool_
```diff
@@ -0,0 +1,290 @@
...
+
+    program = sys.argv[0]
+    try:
+        graph_type, account_address = program.split("_")[1:]
+    except ValueError:
+        print("Please configure the plugin as described in the configuration section.")
+        exit(1)
+
+    GRAPH_TYPES = define_graph_types()
+    if len(sys.argv) > 1 and sys.argv[1] == "autoconf":
```
sumpfralle: Please take a look at the [autoconf documentation](http://munin-monitoring.org/wiki/PluginConcise#autoconf) for some more details.

Since this plugin does not require additional dependencies (e..g non-core python modules), you can simply get rid of `autoconf`, I guess.

sumpfralle:
Great - thank you!

geschke:
Thanks for your comments and hints to improve the quality!

PR #860, opened 2017-07-08
Add deluge_ plugin

Neraud:
Initial version can monitor connections, bandwidth and states.


plugins/deluge/deluge_
```diff
@@ -0,0 +1,304 @@
+#! /usr/bin/env python
+#
+# Munin Wildcard Plugin for Deluge torrent client
+#
```
sumpfralle: You should mention the requirements of this plugin (twisted, deluge).

Neraud:
Thank you for your feedback.
This is the first plugin (and python script for that matter) that I've pushed on Github, so it's an understatement to say that I'm not used to python best practices ;p
I'll clean the whole thing.

About using a "torrent" or "deluge" folder : "torrent" was my first idea, but I've seen a "transmission" folder in the plugins root, so I created a "deluge" one. I'll rename it.

The (old) [example provided by deluge](http://dev.deluge-torrent.org/wiki/Development/1.3/UIClient) to code a complete client was to use twisted, and I followed that without too much thought.
The deluge.ui.client module handles the interactions with the daemon, so I don't think I can easily replace it (and twisted). I'll take a look in case I find a way. In any case, I'll properly document and check for the requirements.

sumpfralle:
> This is the first plugin (and python script for that matter) that I've pushed on Github, so it's an understatement to say that I'm not used to python best practices ;p
> I'll clean the whole thing.

I appreciate that. Welcome to the munin plugin repository!


> The (old) [example provided by deluge](http://dev.deluge-torrent.org/wiki/Development/1.3/UIClient) to code a complete client was to use twisted, and I followed that without too much thought.
> The deluge.ui.client module handles the interactions with the daemon, so I don't think I can easily replace it (and twisted). I'll take a look in case I find a way. In any case, I'll properly document and check for the requirements.

Ah - good catch.
If the deluge module already requires (not suggests) twisted and your plugin
requires deluge, then it is reasonable to depend on twisted, too.
Proper checks (careful imports) will be a good thing, nevertheless.

Thank you!

Neraud:
It should be much cleaner now ;p

Unfortunately, I couldn't migrate the plugin to python3 as deluge itself doesn't support it.

plugins/torrent/deluge_
```diff
@@ -0,0 +1,399 @@
...
+import os
+import string
+import sys
+
+try:
+    from deluge.log import setupLogger
+    from deluge.ui.client import client
+    from twisted.internet import reactor, defer
+    setupLogger()
+except (ImportError, NameError):
```
sumpfralle: Please change this to `except (ImportError, NameError) as exc` - this would work in python3, too (in case it is ever supported by deluge).

Regarding `NameError` - how exactly could this be triggered?

Maybe you want to keep things clean (i.e. only check for ImportError) and run `setupLogger` afterwards in an `if successful_import:` branch?

plugins/torrent/deluge_
```diff
@@ -0,0 +1,399 @@
...
+        defer.DeferredList(deferred_list).addCallback(
+            self.on_all_torrent_info_fetched)
+
+    def on_one_torrent_info_failed(self, torrent_id):
+        log.debug("Failed fetching torrent info {0}".format(torrent_id))
+        self.state["Error"] = self.state["Error"] + 1
+
+    def on_one_torrent_info(self, value, torrent_id):
+        log.debug("Got torrent info : {0} -> {1}".format(torrent_id, value))
+        if 'state' in value:
```
sumpfralle: A maybe more concise and readable alternative could be: `state = value.get("state", "Error")`.

plugins/torrent/deluge_
```diff
@@ -0,0 +1,399 @@
...
+            state = value['state']
+        else:
+            state = "Error"
+
+        if not (state in self.states):
```
sumpfralle: `flake8` would recommend `if state not in self.states` here.

plugins/torrent/deluge_
```diff
@@ -0,0 +1,399 @@
...
+            log.warn("State '{0}' is unknown !".format(state))
+            state = "Other"
+
+        self.states[state] = self.states[state] + 1
```
sumpfralle: `self.states[state] += 1` could be more concise.

plugins/torrent/deluge_
```diff
@@ -0,0 +1,399 @@
...
+
+class StatClient:
+
+    def __init__(self, conf, mode):
+        self.conf = conf
+        self.mode = mode
+        self.connected = False
+
+    def end_session(self, msg):
+        log.debug("end_session : {0}".format(msg))
```
sumpfralle: Maybe you want to use lazy evaluation for log messages (see [logging documentation](https://docs.python.org/3/library/logging.html#logging.debug)):
```python
log.debug("end session: %s", msg)
```
(thus the format strings would only evaluated if the log level is met)

plugins/torrent/deluge_
```diff
@@ -0,0 +1,399 @@
...
+            print(names_for_munin["state." + state_name] + ".draw AREASTACK")
+            print(names_for_munin["state." + state_name] + ".type GAUGE")
+            print(names_for_munin["state." + state_name] + ".min 0")
+            print(names_for_munin["state." + state_name] +
+                  ".info Number of torrents in the '" + state_name + "' state")
+
+
+def fetch_info(mode):
+    if not successful_import:
+        print('Missing imports, cannot run !')
```
sumpfralle: You should write this to stderr - otherwise the munin server would try to interprete it.

Maybe import `print_function` from `__future__` and use `print(foo, file=sys.stderr)`?

Or `sys.stderr.write(foo + os.linesep)`?

plugins/torrent/deluge_
```diff
@@ -0,0 +1,399 @@
...
+# Parse arguments
+if len(sys.argv) > 1:
+    action = sys.argv[1]
+    if action == "config":
+        print_config(get_mode())
+        sys.exit(0)
+    elif action == "autoconf":
+        if not successful_import:
+            print('no (required modules not found)')
+            sys.exit(1)
```
sumpfralle: `autoconf` should always return successfully (even for the `no` case).

plugins/torrent/deluge_
```diff
@@ -0,0 +1,399 @@
...
+        print('yes')
+    elif action == "suggest":
+        for mode in modes:
+            print(mode)
+        sys.exit(0)
+    elif action == "version":
+        print('Deluge Torrent Munin plugin, version {0}'.format(
+            plugin_version))
+        sys.exit(0)
+    elif action != "":
```
sumpfralle: `elif action:` would be quite common for this test.

plugins/torrent/deluge_
Neraud: As I'm not used to writing Python code, I tried the 2 following tests : 

`from [unknown.package] import [submodule]`
-> `ImportError`

`from [existing.package] import doestNotExist`
-> `NameError`

I assume it's rare to have a stable module renaming a submodule.

Is it common practice to only check for ImportError and assume that if/when a submodule is renamed, everything breaks down ?

sumpfralle: Hm - I really could not reproduce a NameError - neither with python2 nor python3.

From my point of view, ImportError should be sufficient.

sumpfralle:
Thank you for contributing this shiny new plugin!

PR #864, opened 2017-07-20
Allow to specify database password

adekock11:


sumpfralle:

I took a look at the `psql` manpage and failed to find `PGPASSWORD` being mentioned. Is this a documented feature?

Anyway: I have the feeling that this is just a configuration issue.
Are you using the plugin on the database server? (this would be the common usage approach for munin plugins)
In this case you could configure the plugin to be run as the system user `postgres` - which would grant the plugin all required permissions for reading the database.

Or how about using a `~/.pgpass` file? This would prevent the exposure of the password to the environment of the psql process.

In general: this plugin is rather cumbersome to use. Currently you need to manipulate the plugin itself, if you want to change the database server and database user. Maybe you would like to prepare a pull request and make this plugin a bit more configurable. This could be done by replacing `dbhost=localhost` with `dbhost=${dbhost:-localhost}`. Or maybe you would take the opportunity and offer a configuration via an environment variable like `psql_args` - here a user could specify even more exotic parameters for psql.

What do you think?

sumpfralle:
@adekock11: ping?

sumpfralle:
I just merged it now, since it does not hurt.

sumpfralle:
Thank you!

PR #865, opened 2017-07-22
Fixed output of transactions script

adekock11:
The current output of this script is:

```
root@master:/etc/munin/plugins# ./pg_transactions 
 \ncommits.value xxxx\nrollbacks.value xxxx
```

By removing the first `\n` and by adding an `E` before the second `\n` the output is:

```
root@master:/etc/munin/plugins# ./pg_transactions 
 commits.value xxxx +
 rollbacks.value xxxx
```

And lastly by adding the `--no-align` attribute you get the correct output:

```
root@master:/etc/munin/plugins# ./pg_transactions 
commits.value xxxx
rollbacks.value xxxx
```


sumpfralle:
This looks good!

PR #872, opened 2017-08-24
New SNMP Plugin for HP LaserJet Printers

kboenke:
Provides consumption levels for HP LaserJet Printers.
Now with multigraph support in a single pull. ;)


sumpfralle:
Great - thank you!

PR #873, opened 2017-08-25
SNMP Plugin for CyberPower UPS

kboenke:
Plugin to monitor status & environment readings readings for CyberPower UPS units with an RMCARD202 module.
Cleaned-up version as Multigraph-Plugin. ;)

### DIFF
--- a/plugins/snmp/snmp__cyberpower
+++ b/plugins/snmp/snmp__cyberpower
@@ -1,0 +1,186 @@
+#!/usr/bin/perl -w
+
+=head1 NAME
+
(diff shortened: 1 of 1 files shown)


sumpfralle:
That looks good - thank you!

btw: since you are a perl-capable person: [there is a recent issue](https://github.com/munin-monitoring/munin/issues/867#issuecomment-324758021) that would be happy to receive a short translation from human language to perl - but I feel way too clumsy for that. Maybe you are in the mood to take a look?
Anyway: thank you!

Cheers,
Lars

PR #874, opened 2017-08-27
Update LXC plugins

zorun:
Large cleanup of the LXC plugins to make them more usable with a non-trivial amount of containers:

- `lxc_cpu` and `lxc_ram` are now wildcard plugins, to generate one graph per container
- use stacked layout to improve readability of graphs
- drop `lxc_cpu_time` which is redundant with `lxc_cpu` (see commit messages)
- other small cleanups

See individual commit messages for details


zorun:
Here is a screenshot of `lxc_proc` before this change and after this change:

Before:
lxc_proc-day

After:
lxc_proc-week

I will post screenshots of the other changed plugins, once enough data has been collected :)

plugins/lxc/lxc_cpu_
```diff
@@ -0,0 +1,100 @@
...
+        echo "graph_title CPU usage for LXC container $CONTAINER "
+        # Set default graph upper limit to 100 (i.e. 1 CPU fully used), so
+        # that containers with low CPU usage have a sensible scale.
+        # Of course a container may use several CPU cores, in which case
+        # the graph will have a scale higher than 100.
+        # TODO: it could be interesting to figure out how many CPU cores a
+        # container is allowed to use, and set this value as upper limit.
+        echo 'graph_args -l 0 -u 100 --base 1000'
+        echo 'graph_vlabel %'
+        echo 'graph_category lxc'
```
sumpfralle: Please use "virtualization" instead of "lxc" (or take a look at the list of [well-known plugin categories](http://guide.munin-monitoring.org/en/latest/reference/graph-category.html)). This will improve the visual appearance in the plugin gallery and keeps the list of categories reasonably short.

zorun:
Hi, thanks for the suggestion!

I will send images to the gallery then.

As for the multigraph plugin: I like being able to enable/disable data collection for each container individually, and using `munin-node-configure` it is quite easy to configure all containers in one go.  However resources usage might indeed be an issue, and I found another issue: containers with forbidden characters in their name (such as dash) fail to generate a graph. I guess this is an issue with all wildcard plugins, but the multigraph will probably solve it.  I will give it a try!

plugins/lxc/lxc_cpu_
zorun: Good idea, fixed.

sumpfralle:
Great - I am looking forward to your multigraph experiences!

Regarding the selection of containers: maybe you want to support a configurable whitelist or blacklist for people who do not want to track all containers (probably rare).

Regarding valid fieldnames: probably you will want to use `clean_fieldname` when determining a suitable fieldname for a given container. This function is already imported as part of munin's shell library file in the lxc plugins.

zorun:
Yes, I know about `clean_fieldname`, the issue is that the name of the plugin itself would need to be cleaned (because just having e.g. `/etc/munin/plugins/lxc_cpu_my-container` already causes trouble).

The whitelist/blacklist is a good idea.

sumpfralle:
@zorun: may I help you somehow? Do you need advice?

sumpfralle:
Ping?

sumpfralle:
I think your proposal is already useful, as it is. But I would like to avoid the change of the plugins to wildcard plugins. Thus if you are short on time, maybe you just want to remove this change?

sumpfralle:
Ping?

ls-initiatives:
Hi, if I'm not mistaken your branch does not include the new multigraph XLC plugin which was merged from https://github.com/munin-monitoring/contrib/pull/1009

I don't think this can be merged anymore, unlesss you can rebase your changes on the new plugin.

sumpfralle:
> I don't think this can be merged anymore, unlesss you can rebase your changes on the new plugin.

Thank you for taking a look and commenting this!

PR #875, opened 2017-09-18
Allow for _ in original dataset name

yoe:
The current code escapes / (which is legal as part of a dataset name for ZFS, but illegal as part of a munin graph name) into _ (which is legal for both). The result is that any _ characters in the original filename get changed into / in the legend, which is confusing.

Fix by escaping any existing _ characters into __ first, and then later replacing // into _ again. // is a sequence that will not occur in a ZFS dataset name, so any double-/ sequence must therefore have been a _ originally.


sumpfralle:
This looks good - thank you!

PR #877, opened 2017-09-28
Use pdns_control utility instead of pdns init script.

sebastic:
The init script is not usable on systems using systemd (e.g. Debian stretch).

autoconf is also made conditional on the existence of the executable to not enable the plugin on systems where pdns is not installed.

### DIFF
--- a/plugins/pdns/pdns_errors
+++ b/plugins/pdns/pdns_errors
@@ -9,11 +9,16 @@
 #%# family=auto
 #%# capabilities=autoconf
 
-command="/etc/init.d/pdns dump"
+command="/usr/bin/pdns_control list"
 
 if [ "$1" = "autoconf" ]; then
-        echo yes
-        exit 0
+        if [ -e /usr/bin/pdns_control ]; then
+                echo yes
+                exit 0
+        else
+                echo no
+                exit 1
+        fi
 fi
 
 if [ "$1" = "config" ]; then
--- a/plugins/pdns/pdns_errors
+++ b/plugins/pdns/pdns_errors
@@ -9,15 +9,16 @@
 #%# family=auto
 #%# capabilities=autoconf
 
-command="/usr/bin/pdns_control list"
+pdns_control="/usr/bin/pdns_control"
+command="$pdns_control list"
 
 if [ "$1" = "autoconf" ]; then
-        if [ -e /usr/bin/pdns_control ]; then
+        if [ -e "$pdns_control" ]; then
                 echo yes
                 exit 0
         else
-                echo no
-                exit 1
+                echo "no (missing $pdns_control)"
+                exit 0
         fi
 fi
 
--- a/plugins/pdns/pdns_errors
+++ b/plugins/pdns/pdns_errors
@@ -43,4 +43,4 @@ if [ "$1" = "config" ]; then
 fi
 
 
-$command | sed 's/=\([0-9]\+\),/.value \1\n/g' | grep corrupt'\|'servfail'\|'timedout | sed 's/-/_/g'
+$command | sed 's/=\([0-9]\+\),/.value \1\n/g' | egrep "corrupt|servfail|timedout" | sed 's/-/_/g'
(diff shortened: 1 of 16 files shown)


plugins/pdns/pdns_errors
```diff
@@ -9,11 +9,16 @@
...
+command="/usr/bin/pdns_control list"
 
 if [ "$1" = "autoconf" ]; then
-        echo yes
-        exit 0
+        if [ -e /usr/bin/pdns_control ]; then
+                echo yes
+                exit 0
+        else
+                echo no
```
sumpfralle: Please return an exitcode of zero even in case of "no" (see [PluginConcise](http://munin-monitoring.org/wiki/PluginConcise#autoconf).
Additionally it would be great if you would provide a reason for "no". The documentation recommends the following pattern: `no (missing /usr/bin/pdns_control)`.

plugins/pdns/pdns_rel
```diff
@@ -31,17 +36,19 @@ if [ "$1" = "config" ]; then
         exit 0
 fi
 
-hits=$($command packetcache-hit | awk -F= '{print $2}')
-queries=$($command udp-queries | awk -F= '{print $2}')
-old_hits=$(cat $state_file | head -n1)
-old_queries=$(cat $state_file | tail -n1)
+hits=$($command packetcache-hit)
+queries=$($command udp-queries)
+if [ -f $state_file ]; then
```
sumpfralle: Please quote variables in order to avoid whitespace issues.
Maybe you want to try `shellcheck` - it is a great tool for verifying the intricate details and pitfalls of shell syntax.

plugins/pdns/pdns_rel
```diff
@@ -31,17 +36,19 @@ if [ "$1" = "config" ]; then
...
+        old_hits=$(cat $state_file | head -n1)
```
sumpfralle: shellcheck would advise the following: `head -n 1 "$state_file"`

plugins/pdns/pdns_rel
```diff
@@ -31,17 +36,19 @@ if [ "$1" = "config" ]; then
...
+        old_queries=$(cat $state_file | tail -n1)
+fi
 
 if [ -f $state_file ] && [ $(ls -l --time-style=+%s $state_file | awk '{print $6}') -gt $(date --date="7 minutes ago" +%s) ] ; then
```
sumpfralle: This is not your piece of the code, but maybe you are in the mood ...
`date --reference %s` is probably a much safer choice than manual parsing of `ls`.
Additionally quoting around all variables and command substitutions would be good.

plugins/pdns/pdns_rel
```diff
@@ -31,17 +36,19 @@ if [ "$1" = "config" ]; then
...
-	d_hits=$(($hits - $old_hits))
-	d_queries=$(($queries - $old_queries))
-	if [ $d_queries -gt 0 ] ; then
-	   echo packetcache_hitrate.value $(( $d_hits * 100 / $d_queries ))
-	fi
+        d_hits=$(($hits - $old_hits))
```
sumpfralle: You may (or should) omit `$` within arithmetic substitution.

plugins/pdns/pdns_rel
sebastic: Quoting the two commands is not needed because the output is a single number not a string.

sebastic:
While I only pushed the changes I made to the pdns plugins used at $DAYJOB, I've now pushed additional commits to fix the shellcheck issues and exit status for all pdns plugins.

sumpfralle:
Great - thank you for taking the time!

These plugins look much better now.
Regarding necessary and unnecessary quoting: this is usually hard to tell without taking a closer look. We will hopefully manage somewhen to improve the quality of the shell plugins in the repository to a state where we could use shellcheck for visible quality tests. Every step in this direction is helpful, from my point of view.

A small hint (maybe you know it): nested quoting/substitution/everything is perfectly fine within the new command substitution (`$(...)`), while it was impossible with the old backticks. Thus you could have changed `"$($rec_control get spoof-prevents)"` to `"$("$rec_control" get spoof-prevents)"`.
Anyway - it is great, as it is right now.
Thank you!

PR #879, opened 2017-10-03
Change commit and rollback types to "COUNTER"

adekock11:
By doing this, Munin knows to take the difference of each request, thereby showing you a true commit and rollback "per minute"

Before:
screen shot 2017-10-03 at 08 27 28

After:
screen shot 2017-10-03 at 08 27 35

### DIFF
--- a/plugins/postgresql/postgresql_transactions
+++ b/plugins/postgresql/postgresql_transactions
@@ -33,10 +33,12 @@ if [ "$1" = "config" ]; then
 
    echo 'commits.label commits'
    echo 'commits.min 0'
+   echo 'commits.type COUNTER'
    echo 'commits.info Number of transaction commits.'
 
    echo 'rollbacks.label rollbacks'
    echo 'rollbacks.min 0'
+   echo 'rollbacks.type COUNTER'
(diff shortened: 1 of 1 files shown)


sumpfralle:
Indeed, this looks better!

Could you please adjust the `info` and maybe the `vlabel` field in order to reflect this change?

steveschnepp:
I'd usually advocate on using `DERIVE` instead of `COUNTER`. It behave mostly the same, with a difference with wrap-around handling. 

If you emit a lower value than the previous one:
* `DERIVE` will think that it went backwards, but with a `.min` equal to `0` it will be replaced to `U`. 
* `COUNTER` will think that it has wrapped, and will add the remaining of `2^32` or `2^64`, whichever is appropriate. This can lead to *HUGE* pikes if the plugin source is just restarted for example.

sumpfralle:
@adekock11: what do you think?

* adjust `info` and `vlabel`
* use `DERIVE` and `min` instead of `COUNTER` (since this is no roll-over 32/64 bit value, I guess)

sumpfralle:
I applied the above changes now in 93a567afbc2bd79ce19863344317d91489161bf5.

PR #880, opened 2017-10-03
Add specific updates to extinfo field.

bertptrs:
As the title says, this update adds information about which updates are pending to the extinfo field. This makes warnings much more informational, since it tells you whether you really need to update now or if you can wait for a while.

### DIFF
--- a/plugins/other/pacman_pending_updates
+++ b/plugins/other/pacman_pending_updates
@@ -14,12 +14,6 @@ All systems with pacman as their package manager.
 
 The plugin needs no additional configuration and works out of the box.
 
-It is possible to add warnings for certain numbers of updates pending. The
-following will send a warning when there are more than 10 updates pending.
-
-  [pacman_pending_updates]
-  env.PENDING_UPDATES_WARNING :10
-
 =head1 INTERPRETATION
 
 This plugin will draw one line: the number of updates pending.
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for polishing this plugin up!

You are using a temporary file for storing the data source.
In this case you should use a `trap` for making sure, that the temporary file is removed under all (internal or external) circumstances.
But in this specific case you could simply store the data in a variable and `echo` it into `wc` and `paste`. This would free you from taking care for temporary files.

Additionally: please take care, that you do not introduce additional shell style issues (see `shellcheck`). In your case the `if` expression and the parameters for your `echo` calls lacked quotes.

What do you think?

bertptrs:
Great suggestion on using `shellcheck`. I added it to my arsenal.

My main reason for piping to a temp file rather than storing it in memory, is to work around output capture skipping trailing newlines. If I do `updates="$(checkupdates)"`, I cannot distinguish between 0 and 1 updates, since `wc -l` will report 0 lines either way. Piping to file removed this issue. Do you have a suggestion to work around this?

sumpfralle:
> My main reason for piping to a temp file rather than storing it in memory, is to work around output capture skipping trailing newlines. If I do updates="$(checkupdates)", I cannot distinguish between 0 and 1 updates, since wc -l will report 0 lines either way.

What an interesting detail! I never thought about that ...

How about checking for an non-empty string `[ -n "$updates" ]` before asking `wc`?

bertptrs:
Great suggestion. I've implemented that and your other suggestions. I also removed the explicit warning option since it was dupllicating a munin option.

sumpfralle:
Cool - thank you!

PR #881, opened 2017-10-05
jenkins: Added support for jobs folders

crispygoth:
Newer versions of Jenkins have support for grouping jobs into "folders",
for example when using a multibranch setup, the main "job" will actually
be a folder containing a job for each branch.

This changeset adds a "jobDepth" parameter that can be used to control
how far down the folder tree to report on.

### DIFF
--- a/plugins/jenkins/jenkins_
+++ b/plugins/jenkins/jenkins_
@@ -28,6 +28,7 @@ env.port 		Jenkins Port
 env.context		Jenkins Context path
 env.user 		User for the API Tokent
 env.apiToken 	Jenkins API Token (see https://wiki.jenkins-ci.org/display/JENKINS/Authenticating+scripted+clients)
+env.jobDepth    How far into job "folders" should the plugin check for jobs
 
 Example:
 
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for taking care for this.

Do you know, if the plugin still works with an older version of jenkins? Or would it fail with the new request parameters?

btw: If you feel like, you may also want to unify the whitespacing in the remainder of this plugin (tabs/spaces) in a separate commit ...

sumpfralle:
Hi crispygoth,

I just went through the open pull requests and merged #883.
And now I noticed, that your pull request handles the same issue (a bit differently).
I am sorry, that I noticed yours (again) only after merging the other one.

Could you please take a loot at the changeset that I merged and maybe comment or improve it, if you think, something (from your proposal) is missing?
Thank you!

Cheers,
Lars

crispygoth:
From a quick glance at that other PR, it looks like that version involves making multiple requests to the Jenkins API to recurse down the job tree. My version involves just a single request; it asks Jenkins to recurse down the tree to a certain depth instead. Personally I think that makes this PR superior?

sumpfralle:
@crispygoth: I finally took the time to merge your changeset into the one from @henti.

Please take a look at it for the following two details:
* I replaced the global `counts` hash with a local variable - please check, if it works properly (my perl is rusty)
* I do not really like the fact, that the "depth" needs to be configured (i.e. defaults to 1). It looks like we could just define an arbitrary higher default depth limit (e.g. 10?) in order to work in most cases. Or would this cause a problem?

sumpfralle:
I forgot to mention the location of my merge proposal: #918
Please feel free to either comment on it or refresh your pull request based on it.

sumpfralle:
Ping?

sumpfralle:
I merged my proposal (#918) now. Please comment on any issues you may encounter using it.

PR #886, opened 2017-11-15
Add strelaysrv plugin

daftaupe:
Hello,

this is a plugin targeted at strelaysrv, the Syncthing relay server.
I hesitated between having a special dir for it or putting it in the syncthing one I created in the other pull request. Tell me what's better, but I think it's sort of independent.

Regards,


plugins/strelaysrv/strelaysrv_
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env bash
```
sumpfralle: You are using no bash-specific features in this plugin (except for the irrelevant `function` keywords, which can be simply removed). Thus it would be good to use `/bin/sh` instead.

plugins/strelaysrv/strelaysrv_
```diff
@@ -0,0 +1,145 @@
...
+graph_category network
+graph_vlabel numbers
+strelaysrv_num_sessions.label sessions
+strelaysrv_num_connections.label connections
+strelaysrv_num_pending.label pending session keys
+strelaysrv_num_proxies.label proxies
+EOM
+        exit 0;;
+        *)
+            NS=$($CURL -s http://$STR_HOST:$STR_PORT/status | $JQ '.numActiveSessions ')
```
sumpfralle: Please call `shellcheck -s dash` (if using `/bin/sh`) or simply `shellcheck` (for `bash`) for this plugin. This will show you quite a number of places, where quoting would be suitable.

plugins/strelaysrv/strelaysrv_
```diff
@@ -0,0 +1,145 @@
...
+            NC=$($CURL -s http://$STR_HOST:$STR_PORT/status | $JQ '.numConnections ')
```
sumpfralle: Since you are requesting the same resource multiple times, you will slow down the plugin and risk an inconsistent mixture of data points. I would recommend to request the status once, store it in a variable and `echo` it into `jq` for each step.

plugins/strelaysrv/strelaysrv_
```diff
@@ -0,0 +1,145 @@
...
+            NK=$($CURL -s http://$STR_HOST:$STR_PORT/status | $JQ '.numPendingSessionKeys ')
+            NP=$($CURL -s http://$STR_HOST:$STR_PORT/status | $JQ '.numProxies ')
+            printf "strelaysrv_num_sessions.value $NS\nstrelaysrv_num_connections.value $NC\nstrelaysrv_num_pending.value $NK\nstrelaysrv_num_proxies.value $NP\n"
```
sumpfralle: This single-liner is a bit hard to read :)
Maybe the more common approach would be more intuitive?
```shell
printf "strelaysrv_num_sessions.value %s\n" "$NS"
```

plugins/strelaysrv/strelaysrv_
```diff
@@ -0,0 +1,145 @@
...
+graph_category network
+graph_vlabel bps
+graph_args --base 1000
+strelaysrv_transfer.label bps
+strelaysrv_transfer.cdef strelaysrv_transfer,1000,*
+EOM
+            exit 0;;
+	*)
+            printf "strelaysrv_transfer.value "
+            $CURL -s http://$STR_HOST:$STR_PORT/status | $JQ '.kbps10s1m5m15m30m60m[2] '
```
sumpfralle: Probably you should output the value in bytes instead (`jq` supports simple operations)? This will allow munin to do handle proper units automatically.

daftaupe:
As it makes more sense, I'm putting this plugin in the syncthing directory, thus closing this PR.

PR #891, opened 2017-12-28
Adding a zimbra spesific clamav plugin

Flums:


sumpfralle:
It looks like this plugin is very similar to `plugins/clamav/clamav`.
In order to keep this repository of plugins in a maintainable state, we would like to keep the number of code copies low and try to use configurable plugins instead.

In your case: I think, it is better to simply make the `log` variable in the original plugin configurable via the environment.

What do you think?

sumpfralle:
ping?

sumpfralle:
Please use a configurable log file instead of adding another plugin for your purpose.

PR #894, opened 2018-01-18
Add support for monitoring packets and link status for TPLink SG108E …

mad-ady:
…switches


plugins/tplink/tl_sg108e
```diff
@@ -0,0 +1,210 @@
...
+    exit 0;
+}
+
+# extract data from the switch - uses web scraping (tested with 1.0.2 Build 20160526 Rel.34615)
+
+# print STDERR "Connecting to $switch with user $username";
+my $mech = WWW::Mechanize->new(
+    autocheck             => 0,
+    requests_redirectable => [ 'GET', 'HEAD', 'POST' ],
+    user_agent => 'Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:57.0) Gecko/20100101 Firefox/57.0'
```
sumpfralle: Is it really necessary, to lie about our user agent?
If yes: please add a comment.

plugins/tplink/tl_sg108e
```diff
@@ -0,0 +1,210 @@
...
+
+if ( $ARGV[0] and $ARGV[0] eq "config" ) {
+    foreach my $graphType (qw/Packets Speed/) {
+        foreach my $port ( sort keys %ports ) {
+            print "multigraph ${graphType}_if_$port\n",
+              "graph_title $graphType for $switch port $port $ports{$port}\n",
+              "graph_info $graphType graph for port $port $ports{$port}\n",
+              "graph_args --base 1000 -l 0\n",
+              "graph_scale yes\n",
+              "graph_category switch\n";
```
sumpfralle: Please use `network` instead.
Or pick another suitable category from [the list of well-known categories](http://guide.munin-monitoring.org/en/latest/reference/graph-category.html#well-known-categories).

plugins/tplink/tl_sg108e
```diff
@@ -0,0 +1,210 @@
...
+if ( $data =~ /link_status:\[([0-9,]+)\]/ ) {
+
+    my $linkString = $1;
+    my @links = split( /,/, $linkString );
+    for ( 1 .. $numberOfPorts ) {
+        my $currentPort = $_;
+        my $link        = $links[ $currentPort - 1 ];
+        print "multigraph Speed_if_$currentPort\n";
+
+        print "p${currentPort}.value $link\n";
```
sumpfralle: Please transform the device-specific metric (0/5/6) into a human-readable metric here.
How about `0` for `not connected`, `100` for `5` and `1000` for `6`?
All other values could be represented by `U`.

plugins/tplink/tl_sg108e
```diff
@@ -0,0 +1,210 @@
...
+
+TP-Link SG108E switches with web management (http). Tested with software version 1.0.2 Build 20160526 Rel.34615
+
+=head1 CONFIGURATION
+
+Add this to the relevant munin-node config file. You can specify switch address, username, password and description for each port
+(the switch management doesn't allow port descriptions)
+
+  [tl_sg108e]
+  env.switch 192.168.1.12
```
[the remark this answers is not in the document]
mad-ady: Will do, thanks

plugins/tplink/tl_sg108e
```diff
@@ -0,0 +1,210 @@
...
+#populate the ports and descriptions based on ENV
+my %ports = ();
+for ( 1 .. $numberOfPorts ) {
+    my $i = $_;
+    if ( defined $ENV{"p$i"} ) {
+        $ports{$i} = $ENV{"p$i"};
+    }
+    else {
+        #no description
+        $ports{$i} = "";
```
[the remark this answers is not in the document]
mad-ady: We can do this as well.

plugins/tplink/tl_sg108e
mad-ady: I will check, otherwise I'll remove it.

plugins/tplink/tl_sg108e
mad-ady: Thanks, I didn't know about that. Will change category. I'd also like to change and have the switch display as a standalone "remote" device, but I haven't gotten it to work yet.

plugins/tplink/tl_sg108e
mad-ady: Sure, that's sensible. However, the values are not only speed, but speed/duplex (e.g. 5 is 100 Full Duplex, 6 is 1000 Full Duplex. Probably 4 is 100 Half Duplex.). I will need to map them out, but I'm not sure how to display them.

sumpfralle: Since you are visualizing distinct states without a real numerical representation, you could also just define a number of states and set the value to zero or one, if the state is currently active:
```
# config
speed_100_half_duplex.label 100 MBit/s Half Duplex
speed_100_full_duplex.label 100 MBit/s Full Duplex
speed_1000_full_duplex.label 1 GBit/s Full Duplex

# values
speed_100_half_duplex.value 0
speed_100_full_duplex.value 1
speed_1000_full_duplex.value 0
```

The resulting graph would be similar to this one: [channel used by a wireless device](https://monitor.opennet-initiative.de/munin/ap/AP2.76/wireless_channel_active_phy0.html)

sumpfralle:
Another note: you initial proposal for the target filename (`plugins/tplink/tl_sg108e`) was better than my suggestion (`plugins/switches/...`). At least this is official, since we wrote down [recommendations for picking the name scheme](https://github.com/munin-monitoring/contrib/tree/master/plugins#plugin-name-and-location) some days ago.

Thus (adding a bit of model variability) `plugins/tplink/tl_sg` could be the way to go?

mad-ady:
I've made some commits to address your recommendations. Do I need to do anything else (push them again?) so that they can be included in the official repo?

sumpfralle:
Great!

There is one last detail left: please rename the example graph to `plugins/tplink/example-graphs/tl_sg-day.png`. This name schema is relevant for the plugin gallery.
Thank you!

mad-ady:
Thanks for merging!

PR #895, opened 2018-01-23
Add varnish5_ plugin

### LINKED ISSUE #876: varnish4_ not working with varnish 5.2
After upgrading our varnish box to 5.2 the munin plugin varnish4 stopped working. 
Output of munin-node.log:
```
2017/09/19-01:25:05 [9833] Error output from varnish4_data_structures:
2017/09/19-01:25:05 [9833]      No such aspect
2017/09/19-01:25:05 [9833]      Known arguments: suggest, config, autoconf.
2017/09/19-01:25:05 [9833]      Run with suggest to get a list of known aspects.
2017/09/19-01:25:05 [9833] Service 'varnish4_data_structures' exited with status 1/0.
2017/09/19-01:25:05 [9833] Error output from varnish4_data_structures:
2017/09/19-01:25:05 [9833]      No such aspect
2017/09/19-01:25:05 [9833]      Known arguments: suggest, config, autoconf.
2017/09/19-01:25:05 [9833]      Run with suggest to get a list of known aspects.
2017/09/19-01:25:05 [9833] Service 'varnish4_data_structures' exited with status 1/0.
```

wintgen:
Resolves #876. In Varnish 5.2 some fields were merged in the XML output, breaking the varnish4_ plugin. This plugin takes these changes into account.

### DIFF
--- a/plugins/varnish/varnish5_
+++ b/plugins/varnish/varnish5_
@@ -1,0 +1,1177 @@
+#!/usr/bin/perl
+# -*- perl -*-
+#
+# varnish5_ - Munin plugin to for Varnish 5.x
+#
+# Copyright (C) 2009  Redpill Linpro AS
+# Original author: Kristian Lyngstol <kristian@bohemians.org>
+# Updated to Varnish 5: Marcus Wintgen <wintgen@uni-wuppertal.de>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+=head1 NAME
+
+varnish5_ - Munin plugin to monitor various aspects of varnish
+
+=head1 APPLICABLE SYSTEMS
+
+Varnish 5.x with varnishstat
+
+=head1 CONFIGURATION
+
+The plugin needs to be able to execute varnishstat.
+
+The configuration section shows the defaults
+  [varnish5_*]
+     group varnish
+     env.varnishstat varnishstat
+     env.name
+
+env.varnishstat can be a full path to varnishstat if it's
+not in the path already.
+
+env.name is blank (undefined) by default and can be used to specify a -n
+name argument to varnish if multiple instances are running on the same
+server.
+
+A few aspects are not linked by default. They are marked as
+'DEBUG' => 'yes' (or any other value). They are:
+
+vcl, bans, bans_lurker, lru, objects_per_objhead,
+losthdr, esi, hcb, shm, shm_writes, overflow,
+session, session_herd, gzip
+
+You can link them yourself with something like this:
+
+  ln -s @@LIBDIR@@/plugins/varnish5_ \
+    @@CONFDIR@@/plugins/varnish5_lru
+
+=head1 INTERPRETATION
+
+Each graph uses data from varnishstat.
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf suggest
+
+=head1 VERSION
+
+ $Id$
+
+=head1 BUGS
+
+The hit_rate graph requires munin r2040 or newer to display
+correctly.
+
+=head1 PATCHES-TO
+
+Please send patches to Kristian Lyngstol <kristian@bohemians.org>
+and/or varnish-misc@varnish-cache.org for significant changes. Munin SVN
+is the authoritative repository for this plugin.
+
+=head1 AUTHOR
+
+Kristian Lyngstol <kristian@bohemians.org>
+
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for taking the time to prepare the updated plugin!

I just noticed, that there is also a vagrant plugin in the "munin" repository (distributed as part of each munin release). We should probably focus on that plugin.
I am about to merge all missing fields from the `varnish4_` plugin (here in this "munin-contrib" repository) to the `varnish_` plugin in the "munin" repository.

Since you assembled the `varnish5_` plugin and probably tested it against a varnish server (I do not have one around), you may be able to help me with a detail.

You changed the handling of the `name` field (compared to the original `varnish4_` plugin).
Here your code seems to indicate, that the following contents of `name` are to be expected:
* `type.name`
* `type.ident.name`

After taking a look at the related [varnish issue](https://github.com/varnishcache/varnish-cache/issues/2377) and the [related varnish code change](https://github.com/varnishcache/varnish-cache/commit/a93c4555ef664ce2835cb43d4cccddd188a27ec#diff-ad564a5c05320f8f36acfd1c2689a11fR279) , it looks like only `type.name` (not `type.ident.name`) would be possible.
In this case (with `ident` always being undefined), the following [condition](https://github.com/munin-monitoring/contrib/pull/895/files#diff-6b3d47c9c386739ada3cb2b661401935R942) would never become true:
```perl
if (defined($type) and $type ne '' and defined($ident) and $ident ne '') {
    $data{$type}{$ident}{$name}{$key} = $data;
```

What do you think?

wintgen:
Originally there were always `type` and `name` fields in the XML output. Depending on the counter there was also an `ident` field, which could also contain a dot.

For example, a snippet from varnish-5.1.3:
```
<?xml version="1.0"?>
<varnishstat timestamp="2018-02-26T08:24:39">
...
	<stat>
		<type>MAIN</type>
		<name>sess_conn</name>
		<value>0</value>
		<flag>c</flag>
		<format>i</format>
		<description>Sessions accepted</description>
	</stat>
	<stat>
		<type>MEMPOOL</type>
		<ident>sess0</ident>
		<name>pool</name>
		<value>10</value>
		<flag>g</flag>
		<format>i</format>
		<description>In Pool</description>
	</stat>
	<stat>
		<type>VBE</type>
		<ident>boot.default</ident>
		<name>req</name>
		<value>0</value>
		<flag>c</flag>
		<format>i</format>
		<description>Backend requests sent</description>
	</stat>
...
</varnishstat>
```

In the current version there is only the `name` field which always contains `type` and `name` (`type.name`) and optionally `ident` (`type.ident.name`).

Snippet from varnish-5.2.1:
```
<?xml version="1.0"?>
<varnishstat timestamp="2018-02-26T08:29:12">
...
	<stat>
		<name>MAIN.sess_conn</name>
		<value>0</value>
		<flag>c</flag>
		<format>i</format>
		<description>Sessions accepted</description>
	</stat>
	<stat>
		<name>MEMPOOL.sess0.pool</name>
		<value>10</value>
		<flag>g</flag>
		<format>i</format>
		<description>In Pool</description>
	</stat>
	<stat>
		<name>VBE.boot.default.req</name>
		<value>0</value>
		<flag>c</flag>
		<format>i</format>
		<description>Backend requests sent</description>
	</stat>
...
</varnishstat>
```

The counters managed by varnish are listed in the [documentation](https://varnish-cache.org/docs/5.2/reference/varnish-counters.html).

`varnishstat -l` prints a complete list, with the ident fields included:
```
Varnishstat -f option fields:
Field name                     Description
----------                     -----------
MGT.uptime                     Management process uptime
MGT.child_start                Child process started
MGT.child_exit                 Child process normal exit
...
MAIN.summs                     stat summ operations
MAIN.uptime                    Child process uptime
MAIN.sess_conn                 Sessions accepted
...
LCK.backend.creat              Created locks
LCK.backend.destroy            Destroyed locks
LCK.backend.locks              Lock Operations
...
MEMPOOL.sess0.live             In use
MEMPOOL.sess0.pool             In Pool
MEMPOOL.sess0.sz_wanted        Size requested
...
SMA.s0.c_req                   Allocator requests
SMA.s0.c_fail                  Allocator failures
SMA.s0.c_bytes                 Bytes allocated
...
VBE.boot.default.pipe_in       Piped bytes from backend
VBE.boot.default.conn          Concurrent connections to backend
VBE.boot.default.req           Backend requests sent
```


The condition you mentioned would indeed never become true, if it weren't for this [line](https://github.com/munin-monitoring/contrib/pull/895/files#diff-6b3d47c9c386739ada3cb2b661401935R921), where `ident` is set, if found in the `name` field.

So i'm guessing that in the source code you linked the `ident` field is already concatenated.

plugins/varnish/varnish5_
```diff
@@ -0,0 +1,1177 @@
...
+	my $graph = $_[0];
+
+	# Need to double-check since set_aspect only checks this if there
+	# is no argument (suggest/autoconf doesn't require a valid aspect)
+	if (!defined($ASPECTS{$graph})) {
+		usage "No such aspect";
+	}
+	my %values = %{$ASPECTS{$graph}{'values'}};
+
+	print "graph_category webserver\n";
```
fr33l: Why this became `webserver` instead of `varnish`? It's kinda confusing

sumpfralle: It is recommended to use categories from the [list of well-known plugin categories](http://guide.munin-monitoring.org/en/latest/reference/graph-category.html#well-known-categories).

But maybe `loadbalancer` would be more suitable? I cannot tell.

codeurimpulsif:
Will this pull request be merged soon ? It seem to be working :-)

kenyon:
Done in #1131.

PR #896, opened 2018-01-28
count failed disks and report critical warning

dpavlin:
This addition is very useful for me, especially with spare disks which can be rebuilt when nobody is watching so you would miss failed disk if you didn't notice rebuild.


sumpfralle:
This looks good - thank you!

PR #899, opened 2018-02-17
update wifi_signal to correct a typo

batmac:
trivial change, needed to have  the right marker to enable autoconf


sumpfralle:
thank you!

PR #902, opened 2018-03-05
Implement plugin for monitoring nullmailer.

bertptrs:
This PR adds a plugin for monitoring the queue size for the nullmailer MTA. Nullmailer is a simple MTA that can only work with an SMTP smarthost rather than serve mail itself.

The plugin monitors queue size and the size of the "failed" folder. Folder locations are configurable from the environment.

### DIFF
--- a/plugins/mail/nullmailer_queue
+++ b/plugins/mail/nullmailer_queue
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 : <<=cut
 
@@ -22,7 +22,6 @@ Example:
 
 [nullmailer_queue]
 user nullmail
-
 env.queuedir /var/spool/nullmailer/queue
 env.errordir /var/spool/nullmailer/failed
 
@@ -56,8 +55,8 @@ errordir=${errordir:-/var/spool/nullmailer/failed}
 
 case $1 in
 	autoconf)
-		if hash nullmailer-queue &> /dev/null; then
-			[[ -r "$queuedir" ]] && echo yes || echo "no (queue dir not readable)"
+		if hash nullmailer-queue >/dev/null 2>/dev/null; then
+			[ -r "$queuedir" ] && echo yes || echo "no (queue dir not readable)"
 		else
 			echo "no (nullmailer not installed)"
 		fi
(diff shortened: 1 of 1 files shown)


plugins/mail/nullmailer_queue
```diff
@@ -0,0 +1,90 @@
...
+user nullmail
+
+env.queuedir /var/spool/nullmailer/queue
+env.errordir /var/spool/nullmailer/failed
+
+=head1 INTERPRETATION
+
+This plugin will draw a stack of 2: the number of emails in the queue, and
+the number of emails that have permanently failed sending. The latter has
+a warning attached, since those should never happen.
```
sumpfralle: Just for my understanding:
* number of emails in queue: this is a *state*
* number of permanently rejected emails: this feels like a *rate* to me (emails per seconds)
    * does this value grow forever? (with manual cleanup)

Does this combination result in an understandable graph?
Maybe two separate graphs would be better? (maybe produced by a single "multigraph" plugin)

plugins/mail/nullmailer_queue
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
```
sumpfralle: Below there is only a single occurrence of a bash-only feature: `&>` (which can be easily replaced). Maybe you want to turn this into a `/bin/sh` plugin?

bertptrs: Apparantly, I used two features, since I also used `[[ … ]]` which is also a bash-only feature. In general, it is more flexible and safe to use than `[ … ]` as well as being a shell syntax feature rather than an external command, though in bash and zsh, it is a builtin. Anyway, since I didn't use those features I changed it to `[ … ]`.

If you'd like, you can read more about it here: http://mywiki.wooledge.org/BashFAQ/031

plugins/mail/nullmailer_queue
bertptrs: Number of permanently rejected emails is a _state_, and is ever increasing with manual cleanup as you said. Think of it as the `hold` value for the upstream `postfix` plugin.

I chose this version, since there is, under normal operation, no reason that this should ever be non-zero, and anything else requires administrator intervention, since it is most likely a configuration error.

bertptrs:
I cleaned up according to your suggestions, let me know what you think

sumpfralle:
It looks fine - thank you for your explanations!

PR #903, opened 2018-03-06
Bugfix: fix graph scaling in munstrap.

bertptrs:
Bootstrap sets box-sizing to border-box to make lay-outing easier, but
this damages the scaling of the munin graphs, which get a border to
indicate their warning state.

This patch makes all images have box-sizing: content-box, so that the
graphs are actually the intended size, rather than the intended size
minus the border, causing a blurry image.

I didn't know what minifier was used to make the minified CSS file, so I used [this one](https://github.com/tdewolff/minify), which did not produce the exact same result but close enough not to matter.


sumpfralle:
This sounds like a reasonable change - thank you!

The pure existence of the minified css file in the repository feels like a mistake to me.
What do you think: should it be removed and instead delivered as-is (1800 bytes instead of 1300). Or should we add the minification step to the deployment description of the template?
(this is not the topic of your changeset - but maybe you want to share your thoughts)

bertptrs:
I think it's a two-parter here:
- We have the Bootstrap library, which is quite large and can be expected to (reasonably) never change, so that can be stored minified, if at all. In fact, I think there is very little reason to store the original version of that one.
- Storing the minfied version of the munstrap CSS (and JS, for that matter) is indeed a bit silly, and can indeed be better left to a deployment stage or for package maintainers.

I happen to maintain an [AUR package]](https://aur.archlinux.org/packages/munin-template-munstrap-git) of munstrap, and it would be very difficult to include a small build step if that would be the case.

On the other hand, the files aren't that large to begin with, and gzip makes that even less relevant. I think I'm in favor of removing them altogether.

On an even less related note, script execution could possibly be improved by moving the script declarations to the `<head>` and giving them `defer` attributes. This causes the parser to start loading them immediately (not executing them first) and only executing them after the entire page has been parsed and downloaded, excluding images.

Anyway, all of that is definitely outside the scope of this PR, which just intends to fix a rendering bug.

sumpfralle:
Thank you for sharing your thoughts!

Regarding the removal and the `defer`: I would welcome both. Do not hesitate to open another pull request, if you feel like.

PR #906, opened 2018-03-09
move all mail-related plugins to plugins/mail/

sumpfralle:
The following mail-related plugins are not part of this change:
* horde (webmail)
* mailman (mailinglists)

See #546


sumpfralle:
Please comment, if this looks like a step in the right direction ...

dipohl:
Hi Lars,

On Fri, 09 Mar 2018 03:31:04 +0000 (UTC)
sumpfralle wrote:

> Please comment, if this looks like a step in the right direction ...

no it doesn't from my point of view [1]

My concept for structuring the plugin gallery is

1st level hierarchy is determined from setting the plugins category,
2nd level hierarchy is determined from the name of the 1st level directory beneath "plugins".

Here an example to illustrate the impact, that your
suggested changes would have on the plugin gallery

> rename from plugins/qmail/qmailconn
> rename to plugins/mail/qmail/qmailconn

Now:
plugin "qmailconn" is shown 
in category index "mail"
in section "qmail"
http://gallery.munin-monitoring.org/contrib/mail-index.html

After:
plugin "qmailconn" will be shown
in category index "mail"
in section "mail"

Of course we can change the gallery build script 
so that it fetches the section name from the 
directory name where the plugin is located


but (!!!)

then the contributors will not see the 
product/application directories on the first
directory level under "plugins"

and (!!!)

as a product/application can be monitored in various ways 
plugins with relation to different categories
will spread all around in the directory tree..

e.g. postfix plugins for perspective of spam protection,
postfix plugins for perspective of memory usage,
postfix plugins for perspective of various measurements..

So if you move the plugins to directories that
are equal to the _category_, soon will occur
multiple product/application directories
with the same directory names under different
category directories..

and it will be difficult for the contributors
to place their plugins in the right one of the multiple directories, 

the number of duplicated plugins will rise,

users that search for product/application plugins
in the repo browser will have to browse all 
directories to look for subdirectories that
match their search string 

and ... so on with nasty usability issues.


Therefore I had come to the conclusion,
that it will be the best to avoid using
category names for the directories and
instead build a flat structure 
with specific product/application terms on the 1st level.

I hope you agree.

fyi and cheers!

Gabriele


[1] your PR will result in undoing my action from a year ago:
https://github.com/munin-monitoring/contrib/commit/54a91c13a4156fe72b87945a237c1ae80f364716

PR #909, opened 2018-03-13
varnish5 plugin

ivarmedi:
This is basically the Varnish4 plugin with a line added that strips the namespace that comes before the key being checked (`MAIN`, `VBE`, etc.).

I'm not 100% sure this is needed (or the right approach), as Varnish4 seems to have these namespaces as well, but when trying to use the varnish4 plugin with varnish 5, it fails with 
```
Error: <field> not part of varnishstat.
```

where `<field>` is whatever key I'm looking for

I currently can't easily test on a Varnish4 install

### DIFF
--- a/plugins/varnish/varnish5_
+++ b/plugins/varnish/varnish5_
@@ -1,0 +1,1170 @@
+#!/usr/bin/perl
+# -*- perl -*-
+#
+# varnish5_ - Munin plugin to for Varnish 5.x
+# Copyright (C) 2009  Redpill Linpro AS
+#
+# Author: Kristian Lyngstol <kristian@bohemians.org>
+# Author: Ivar Larsson <ivar@larsson.me>
+#
+# This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


plugins/varnish/varnish5_
```diff
@@ -0,0 +1,1170 @@
...
+sub xml_characters {
+	my $d = $_[1];
+	if ($state{'stat'} == 0) {
+		return;
+	}
+	if ($state{'field'} eq "type" && $d eq "MAIN") {
+		return;
+	}
+	$d =~ s/^\s*$//g;
+	$d =~ s/([A-Z]+\.)?//g;
```
ivarmedi: this line has been added, otherwise the script == varnish4_

gnuheidix:
I can confirm that this change works excellent with Varnish 6.0.2 as well.

codeurimpulsif:
Will this pull request be merged soon ? It seem to be working :-)

PR #910, opened 2018-03-15
Fix broken link in plugin README

bertptrs:
Missing protocol caused it to link to a nonexistent place on github.


sumpfralle:
Indeed - thank you!

PR #913, opened 2018-03-27
Plugin to get hashrate of Flypool Zcash mining pool

Cyclenerd:
Hi,

I wrote a new plugin for cryptocurrency Zcash (ZEC). With this plugin you can monitor your miner hashrate.

Earlier comments of my Ethereum plugins were considered:
https://github.com/munin-monitoring/contrib/pull/851

Sample:
https://imgur.com/gFws4Hc

Best regards
Nils

### DIFF
--- a/plugins/currency/zcash/zcash_flypool_hashrate_
+++ b/plugins/currency/zcash/zcash_flypool_hashrate_
@@ -85,22 +85,21 @@ except IOError as exc:
 # manage codec and error handling lookup process
 reader = codecs.getreader("utf-8")
 
+hash_rate = "U"
+
 try:
     mining_stats = json.load(reader(mining_stats_raw))
 except ValueError:
     print("Failed to parse JSON responce.", file=sys.stderr)
-    print("flypool_hs_{}_{}.value U".format(zcash_address, miner))
 else:
     try:
         worker = mining_stats['data']
     except (KeyError, TypeError):
         print("JSON result error!", file=sys.stderr)
-        print("flypool_hs_{}_{}.value U".format(zcash_address, miner))
     else:
         try:
             hash_rate = worker['currentHashrate']
         except (KeyError, TypeError):
             print("No current Hashrate!", file=sys.stderr)
-            print("flypool_hs_{}_{}.value U".format(zcash_address, miner))
-        else:
-            print("flypool_hs_{}_{}.value {}".format(zcash_address, miner, hash_rate))
+
+print("flypool_hs_{}_{}.value {}".format(zcash_address, miner, hash_rate))
--- a/plugins/currency/zcash/zcash_flypool_hashrate_
+++ b/plugins/currency/zcash/zcash_flypool_hashrate_
@@ -82,8 +82,10 @@ except IOError as exc:
     print("Failed to request Flypool API: {}".format(exc), file=sys.stderr)
     sys.exit(9)
 
+reader = codecs.getreader("utf-8")
+
 try:
-    mining_stats = json.load(mining_stats_raw)
+    mining_stats = json.load(reader(mining_stats_raw))
 except ValueError:
     print("Failed to parse JSON responce.", file=sys.stderr)
     sys.exit(9)
--- a/plugins/currency/zcash/zcash_flypool_hashrate_
+++ b/plugins/currency/zcash/zcash_flypool_hashrate_
@@ -82,13 +82,10 @@ except IOError as exc:
     print("Failed to request Flypool API: {}".format(exc), file=sys.stderr)
     sys.exit(9)
 
-# manage codec and error handling lookup process
-reader = codecs.getreader("utf-8")
-
 hash_rate = "U"
 
 try:
-    mining_stats = json.load(reader(mining_stats_raw))
+    mining_stats = json.loads(mining_stats_raw.read().decode("utf-8"))
 except ValueError:
     print("Failed to parse JSON responce.", file=sys.stderr)
 else:
--- a/plugins/currency/zcash/zcash_flypool_hashrate_
+++ b/plugins/currency/zcash/zcash_flypool_hashrate_
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 """
 =head1 NAME
@@ -7,7 +7,7 @@ zcash_flypool_hashrate_ - Munin plugin to monitor your zcash.flypool.org hashrat
 
 =head1 APPLICABLE SYSTEMS
 
-All systems with "python" and "munin"
+All systems with "python3" and "munin"
 
 =head1 CONFIGURATION
 
@@ -42,20 +42,11 @@ GPLv2
 =cut
 """
 
-from __future__ import print_function
-
 import sys
 import json
 import codecs
-
-try:
-    # python3
-    from urllib.request import urlopen
-    from urllib.request import Request
-except ImportError:
-    # python2
-    from urllib2 import urlopen
-    from urllib2 import Request
+from urllib.request import urlopen
+from urllib.request import Request
 
 command = ''
 if len(sys.argv) > 1:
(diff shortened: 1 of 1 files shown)


plugins/currency/zcash/zcash_flypool_hashrate_
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python
```
sumpfralle: Please use python3 for new plugins.

plugins/currency/zcash/zcash_flypool_hashrate_
```diff
@@ -0,0 +1,113 @@
...
+import sys
+import json
+import codecs
+
+try:
+    # python3
+    from urllib.request import urlopen
+    from urllib.request import Request
+except ImportError:
+    # python2
```
sumpfralle: Python2 ist EOL very soon - you should not clutter the code with support for it.

plugins/currency/zcash/zcash_flypool_hashrate_
```diff
@@ -0,0 +1,113 @@
...
+flypool_api_url = 'https://api-zcash.flypool.org/miner/' + zcash_address + '/worker/' + miner + '/currentStats'
+
+mining_req = Request(flypool_api_url)
+# User-Agent to bypass Cloudflare
+mining_req.add_header('User-Agent', 'Flypool Munin Plugin/1.0')
+
+try:
+    mining_stats_raw = urlopen(mining_req, timeout=15)
+except IOError as exc:
+    print("Failed to request Flypool API: {}".format(exc), file=sys.stderr)
```
sumpfralle: You should break here or provide a sane fallback value. Otherwise you will run into a `NameError` below due to the missing `mining_stats_raw`.

In case of errors (value is not retrievable at the moment), you should return the magic value `U`, e.g. `foo.value U`. This indicates to the munin master, that the plugin worked, but a value was not accessible.

plugins/currency/zcash/zcash_flypool_hashrate_
```diff
@@ -0,0 +1,113 @@
...
+
+try:
+    mining_stats = json.load(reader(mining_stats_raw))
+except ValueError:
+    print("Failed to parse JSON responce.", file=sys.stderr)
+    sys.exit(9)
+
+try:
+    worker = mining_stats['data']
+except:
```
sumpfralle: Here `except (KeyError, TypeError):` should be good.
Bare except clauses are not recommended and will fail with all python code quality checking tools.

plugins/currency/zcash/zcash_flypool_hashrate_
```diff
@@ -0,0 +1,113 @@
...
+mining_req = Request(flypool_api_url)
+# User-Agent to bypass Cloudflare
+mining_req.add_header('User-Agent', 'Flypool Munin Plugin/1.0')
+
+try:
+    mining_stats_raw = urlopen(mining_req, timeout=15)
+except IOError as exc:
+    print("Failed to request Flypool API: {}".format(exc), file=sys.stderr)
+
+reader = codecs.getreader("utf-8")
```
sumpfralle: What is the purpose of this `reader`? Isn't `json.loads(mining_stats_raw.decode("utf-8"))` sufficient?

plugins/currency/zcash/zcash_flypool_hashrate_
```diff
@@ -0,0 +1,113 @@
...
+    print("JSON result error!", file=sys.stderr)
+    sys.exit(9)
+
+try:
+    hash_rate = worker['currentHashrate']
+except:
+    print("No current Hashrate!", file=sys.stderr)
+    sys.exit(9)
+
+print("flypool_hs_{}_{}.value {}".format(zcash_address, miner, hash_rate))
```
sumpfralle: Here you are missing the linebreak at the end of the last line. Maybe your editor is misconfigured?

Cyclenerd:
Thank you for your comments. I could do almost anything. But I need `reader = codecs.getreader("utf-8")`. Without I have problems with Python <3.6.

`decode("utf-8")` is not working.

```
mining_stats = json.loads(mining_stats_raw.decode("utf-8"))
AttributeError: 'HTTPResponse' object has no attribute 'decode'
```

Everything works with this version. My tests were successful.

sumpfralle:
> Without I have problems with Python <3.6.

But we don't want no magic lines of code :)

> AttributeError: 'HTTPResponse' object has no attribute 'decode'

This one explains it: you are dealing with an HTTPResponse here - not with `bytes` (I assumed that before).
Thus the following should decode the body of the response object:
```python
mining_stats_raw.read().decode("utf-8")
```

For perfection, you should output "U" as the value in case of errors. You could do this by chaining the three last`try/except` blocks in the plugin into the (to be added) `else` branch of their respective previous `if/except` block. The `else` branch of an `if/except` block is executed, when there were no errors in the `try` block.
What do you think?

Cyclenerd:
Does not seem to be a Python version problem as I first suspected. Have now also version 3.4 under macOS. Python with operating system macOS outputs `str`. Under Ubuntu bytes. `codecs.getreader()` works with both. Do you have a different idea? I added a comment to clarify the line.

macOS:
```
python3.4 zcash_flypool_hashrate_XXXXXX_XXXX
  File "zcash_flypool_hashrate_XXXXXX_XXXX", line 86, in <module>
    mining_stats = json.load(mining_stats_raw.read().decode("utf-8"))
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/json/__init__.py", line 265, in load
    return loads(fp.read(),
AttributeError: 'str' object has no attribute 'read'
````

Ubuntu: no error

I implemented your "U" suggestion.

**Comment by Cyclenerd** [comment]
The JSON lib behaves differently:

```
$ python3.4 -c 'import json; from urllib.request import urlopen; raw = urlopen("http://json-schema.org/example/card.json"); json.load(raw); print("OK")'
  File "<string>", line 1, in <module>
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/json/__init__.py", line 268, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'

$ python3.6 -c 'import json; from urllib.request import urlopen; raw = urlopen("http://json-schema.org/example/card.json"); json.load(raw); print("OK")'
OK
```

With codecs.getreader() I have no problems.

**Comment by sumpfralle** [comment]
The difference is detailed in the [changelog for "json" in Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#json):
> json.load() and json.loads() now support binary input.

But all versions should support the following:
```python3
from urllib.request import urlopen
response = urlopen("https://fsf.org")
text = response.read().decode("utf-8")
```
The resulting text can be parsed via `json.load` in all versions since Python 3.0. This is the clean and readable approach, that should be suitable for your situation.

Cyclenerd:
I found it. I have to use `json.loads` not `json.load`.

load:
```
python3.4 -c 'import json; from urllib.request import urlopen; response = urlopen("http://json-schema.org/example/card.json"); text = response.read().decode("utf-8"); json.load(text); print("OK")'
  File "<string>", line 1, in <module>
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/json/__init__.py", line 265, in load
    return loads(fp.read(),
AttributeError: 'str' object has no attribute 'read'
```

loads:

```
python3.4 -c 'import json; from urllib.request import urlopen; response = urlopen("http://json-schema.org/example/card.json"); text = response.read().decode("utf-8"); json.loads(text); print("OK")'
OK
```

Cyclenerd:
ping @sumpfralle. The plugin runs without problems. You can merge it if you want.

sumpfralle:
Indeed - I am sorry for lettings this rest for so long :(
Thank you!

PR #915, opened 2018-03-28
Fix du-2 to use clean_fieldname()

Talkless:
Update du-2 to fix "No .label provided" error by using clean_fieldname().

Before:
munin-du-2
After:
munin-du-2-fixed


sumpfralle:
This looks good - thank you!

PR #919, opened 2018-06-13
change dependency from gawk to awk

WanpengQian:
As default installation of FreeBSD does not include gawk, change the dependency from gawk to awk will be nice for some platforms other than Linux.


sumpfralle:
Thank you!

PR #921, opened 2018-07-02
rtorrent version 0.9.7 compatible

sulaweyo:
In version 0.9.7 deprecated xmlrpc calls were dropped. The added files contain the needed changes for version 0.9.7+ of rtorrent.


sumpfralle:
Thank you for preparing these changes for the new interface!

I have two general questions regarding your proposal:
* Would older versions of rtorrent work with these new changes, too? (if yes: how old?)
* If you really think, you would need a separate plugin for the new xml_rpc_calls:
    * What could be a better name/suffix/prefix for these? ("_097_" will be quite confusing in a few years)
    * Could you join these four new plugins into one [multigraph plugin](http://guide.munin-monitoring.org/en/latest/plugin/multigraphing.html#plugin-multigraphing), please? This would reduce code duplication (and the future maintenance burden) a lot.

Thank you!

sulaweyo:
The new API calls are in since 0.9.0 which came out in Dec 2011 but until 0.9.7 they were just deprecated, in 0.9.7 they were dropped.

So a new set of files does not make a lot of sense, i simply did not dare to just drop the old code. I have now updated the pr with a better approach: the api is now a config option which defaults to the old api but allows to switch to the new one.

I don't do a lot with perl so i hope i did not break every coding style the way i did it :)

plugins/rtorrent/rtom_allsessions_vol
```diff
@@ -3,7 +3,7 @@
 # xmlrpc based munin plugin for monitoring rtorrent's torrent count
 # prerequisites:
 #  - rtorrent 0.7.5 or newer compiled with --with-xmlrpc-c
-# check http://libtorrent.rakshasa.no/wiki/RTorrentXMLRPCGuide for further information
+# check http://libtorrent.rakshasa.no/wiki/RTorrentXMLRPCGuide for further informations
```
sumpfralle: I am not a native speaker, but I am quite confident, that there is no plural for "information" in English :)

plugins/rtorrent/rtom_allsessions_vol
```diff
@@ -80,48 +83,54 @@ my $src         = $ENV{"src"} || "";
...
+        } else {
+                for $port (@ports)
+                {
+                        socket( SOCK, PF_INET, SOCK_STREAM, getprotobyname( "tcp" ) );
+                        connect( SOCK, sockaddr_in( $port, inet_aton( $ip ) ) );
+                        my $line = "${hlen}:${header},${line}";
+                        print SOCK $line;
+                        flush SOCK;
+                        while ( $line = <SOCK> ) {
+                                                if ( $line =~ /$pattern/ ) {
```
sumpfralle: Since you are changing the whitespace here (tabs to spaces), please feel free to fix the indentation in this block.
(the nesting levels look weird)

sulaweyo:
Applied consistent indentation on all files and renamed legacy to pre09.

sumpfralle:
Great!

There is just one more thing: please use `current` as the default (instead of `pre097`).

sulaweyo:
switched the default to the current API version

sumpfralle:
Great!
Thank you for your persistence ...

PR #924, opened 2018-07-20
Added git_commit_behind

Neraud:
Added a new plugin to monitor local git repositories and report how many commits behind their remote they are.


plugins/git/git_commit_behind
```diff
@@ -0,0 +1,260 @@
...
+Neraud (https://github.com/Neraud)
+
+=head1 LICENSE
+
+GPLv2
+
+=cut"""
+
+
+from __future__ import print_function
```
sumpfralle: This is not necessary for python3.

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,260 @@
...
+                        format='%(asctime)s %(levelname)-7s %(message)s')
+
+conf = {
+    'git_path':  os.getenv('git_path', '/usr/bin/git'),
+    'state_file': os.getenv('MUNIN_STATEFILE',
+                            '/var/lib/munin-node/plugin-state/nobody/' +
+                            'git_commit_behind.state')
+}
+
+repo_codes = set([re.search('repo\.([^.]+)\..*', elem).group(1)
```
sumpfralle: Just for detailed nitpicking: you can use an generator expression here instead of the list comprehension. Simply dropping `[` and `]` would do that.

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,260 @@
...
+                  for elem in os.environ.keys() if 'repo.' in elem])
```
sumpfralle: Probably `elem.startswith("repo.")` would be a more precise test?

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,260 @@
...
+# happen), but ensure that there will never be more than two hours
+# (7200 seconds) interval between updates.
+*/5 * * * * root if [ -x /etc/munin/plugins/git_commit_behind ]; then /usr/sbin/munin-run git_commit_behind update 7200 12 >/dev/null; fi
+
+=head1 CONFIGURATION
+
+Use your "/etc/munin/plugin-conf.d/munin-node" to configure this plugin.
+    [git_commit_behind]
+    user root
+    env.git_path /path/to/git
```
sumpfralle: Maybe mention the default for `git_path`, since most users will not need to override it.

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,260 @@
...
+        print('%s.warning %s' % (repo_code, repos_conf[repo_code]['warning']))
+        print('%s.critical %s' %
+              (repo_code, repos_conf[repo_code]['critical']))
+
+
+def generate_git_command(repo_conf, git_command):
+    if not repo_conf['user'] or repo_conf['user'] == os.environ['USER']:
+        cmd = [conf['git_path']] + git_command
+    else:
+        shell_cmd = 'cd %s ; %s %s' % (
```
sumpfralle: This string formatting is not safe and should probably not used in a root context.
Please use at least something like the module `shlex` for sanitizing the string.

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,260 @@
...
+                ['rev-parse', '--abbrev-ref', '--symbolic-full-name', '@{u}'])
+            logging.debug('remote_branch = %s' % remote_branch)
+
+            commits_behind = execute_git_command(
+                repos_conf[repo_code],
+                ['rev-list', 'HEAD..%s' % remote_branch, '--count'])
+
+            print('%s.value %d' % (repo_code, int(commits_behind)))
+        except CalledProcessError as e:
+            logging.error('Error executing git command : %s' % str(e))
```
sumpfralle: The logging module supports lazy evaluation for optimizing resource usage: simply use `, e` instead of `% str(e)`.

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,260 @@
...
+        try:
+            logging.info('Fetching repo %s' % repo_code)
+            execute_git_command(repos_conf[repo_code], ['fetch'])
+        except CalledProcessError as e:
+            logging.error('Error executing git command : %s' % str(e))
+        except FileNotFoundError as e:
+            logging.error('Repo not found at path %s' %
+                          repos_conf[repo_code]['path'])
+    logging.debug('Updating the state file')
+    open(conf['state_file'], 'w')
```
sumpfralle: What is the purpose of this line? Please comment or remove it.

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,260 @@
...
+                print('no (git seems to be broken ?!)')
+        else:
+            print('no (git is missing or not executable)')
+    elif action == 'version':
+        print('Git commit behind Munin plugin, version {0}'.format(
+            plugin_version))
+    elif action == 'update':
+        check_update_repos()
+    elif action:
+        logging.warn('Unknown argument \'%s\'' % action)
```
sumpfralle: You can improve readability by minimizing escaping with alternating quote characters.

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,260 @@
...
+
+def print_config():
+    print('graph_title Git repositories - Commits behind')
+
+    print('graph_args --base 1000 -r --lower-limit 0')
+    print('graph_vlabel number of commits behind')
+    print('graph_scale yes')
+    print('graph_info This graph shows the number of commits behind' +
+          ' for each configured git repository')
+    print('graph_category system')
```
sumpfralle: I could imagine, that `filetransfer` could be a more suitable category for this plugin.
What do you think?
Otherwise just leave it as it is ...

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,260 @@
...
+
+debug = int(os.getenv('MUNIN_DEBUG', os.getenv('DEBUG', 0))) > 0
+if debug:
+    logging.basicConfig(level=logging.DEBUG,
+                        format='%(asctime)s %(levelname)-7s %(message)s')
+
+conf = {
+    'git_path':  os.getenv('git_path', '/usr/bin/git'),
+    'state_file': os.getenv('MUNIN_STATEFILE',
+                            '/var/lib/munin-node/plugin-state/nobody/' +
```
sumpfralle: Please rely on the munin node to supply a `MUNIN_STATEFILE` path - otherwise this feature is simply not supported.

Neraud:
Thank you for your feedback.

It might indeed be unnecessary to use the cron job to perform the git fetch operations.
I wanted to make sure that the plugin execution was as fast as possible, and was afraid that if you were to configure many repositories, it could get slowed down too much.
I was also afraid of large fetches over poor networks.

I will add both options: 
By default, perform the fetch during the plugin execution. 
Optionally, use a cron job instead.

sumpfralle:
> I will add both options:
> By default, perform the fetch during the plugin execution.
> Optionally, use a cron job instead.

great!

plugins/git/git_commit_behind
Neraud: I've added details about those settings a few lines below (L51 for `git_path`)

plugins/git/git_commit_behind
Neraud: The apt plugins are in `system`, so that's the one I've selected by default.
In my mind, `file_transfer` makes more sense for downloads (wget, torrent, newsgroup), backups, rsync ...

I don't mind changing it, if you think `file_transfer` is better suited, I'll update it.

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,283 @@
...
+
+git_commit_behind - Munin plugin to monitor local git repositories and report
+how many commits behind their remote they are
+
+=head1 NOTES
+
+This plugin is similar to how apt_all works for apt packages.
+
+To be able to check how behind a git repository is, we need to run git fetch.
+To avoid fetching all repos every 5 minutes and slowing down the munin-node,
```
sumpfralle: Please update with regards to the optional nature of the cron job. How about the following?
> To avoid fetching all repos every 5 minutes (for each munin period) and thus slowing down the data collection, the `git fetch` operation is only randomly triggered (based on `env.update.probability`). In case of very time-consuming update operations, you can run them in a separate cron job.

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,283 @@
...
+
+Use your "/etc/munin/plugin-conf.d/munin-node" to configure this plugin.
+    [git_commit_behind]
+    user root
+    env.git_path /path/to/git
+    env.update.mode [munin|cron]
+    env.update.probability 12
+    env.update.maxinterval 7200
+
+user root : required (to be able to switch to each repo user)
```
sumpfralle: How about the following?
> user root : specify the owner of the repository checkouts or (in case of multiple different owners), use `root`

plugins/git/git_commit_behind
```diff
@@ -0,0 +1,283 @@
...
+import re
+from shlex import quote
+from subprocess import check_output, call, DEVNULL, CalledProcessError
+import sys
+import time
+
+
+plugin_version = "1.0.0"
+
+debug = int(os.getenv('MUNIN_DEBUG', os.getenv('DEBUG', 0))) > 0
```
sumpfralle: I think, `MUNIN_DEBUG` is sufficient here.

plugins/git/git_commit_behind
sumpfralle: Given that a git repository checkout can be used for a variety of purposes, I would prefer `file_transfer`.

Neraud:
Thank you for your feedback.
All those nitpicks (and bonus !) should be fixed ;p

sumpfralle:
Cool!

PR #926, opened 2018-07-25
Use JSON format to obtain data from 'ceph -s'

xavise:
'ceph -s' text output format strongly depends upon ceph version. Using JSON format to obtain data works across different ceph versions.

### DIFF
--- a/plugins/ceph/ceph_osd
+++ b/plugins/ceph/ceph_osd
@@ -44,6 +44,6 @@ if [ "$1" = "config" ]; then
 	exit 0
 fi
 
-echo "osds.value $(ceph -s | grep osdmap | awk '{ print $3 }')"
-echo "up.value $(ceph -s | grep osdmap | awk '{ print $5 }')"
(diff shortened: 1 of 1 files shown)


sumpfralle:
This change looks good! Thank you.

You *closed* the pull request: did you mean to retract it or did you want to indicate, that it is finished from your side?
(the usual github flow would be to close it only, if it is retracted or rejected)

xavise:
Thank you for comments. I was about to include another optimization. I am going to open a new pull request right away.

sumpfralle:
> I am going to open a new pull request right away.

great!

Regarding the evaluation of the data: I think, it would be good to run `ceph -s` only once and store the result in a variable (instead of requesting it multiple times). This would prevent potential inconsistencies based on the delay between successive requests.
Just a hint - I am not sure, how relevant this is for this specific plugin ...

xavise:
You're absolutely right. That's exactly the optimization I was talking about.

I have already opened a new pull request including it:
https://github.com/munin-monitoring/contrib/pull/927

Thanks again for your comments! :-)

PR #932, opened 2018-08-24
Plugin bitcoind improvements

sumpfralle:
* reformat documentation header
* migrate to python3
* introduce explicit setting for configuration file


Talkless:
@sumpfralle looks like it does not support `rpcauth` alternative of specifying login and hashed password. Would it be hard to add that?

PR #935, opened 2018-09-01
bbox: upgrade bbox version.

steveschnepp:
The UI isn't compatible anymore.

### DIFF
--- a/plugins/router/bbox
+++ b/plugins/router/bbox
@@ -109,32 +109,32 @@ then
 fi
 
 echo multigraph bbox_adsl_bw
-echo up.value $(jq ".[].wan.xdsl.up.bitrates" < xdsl.json)
-echo down.value $(jq ".[].wan.xdsl.down.bitrates" < xdsl.json)
+echo up.value $(jq -r ".[].wan.xdsl.up.bitrates" < xdsl.json)
+echo down.value $(jq -r ".[].wan.xdsl.down.bitrates" < xdsl.json)
 
 echo multigraph bbox_adsl_att
-echo up.value $(jq ".[].wan.xdsl.up.attenuation" < xdsl.json)
-echo down.value $(jq ".[].wan.xdsl.down.attenuation" < xdsl.json)
+echo up.value $(jq -r ".[].wan.xdsl.up.attenuation" < xdsl.json)
+echo down.value $(jq -r ".[].wan.xdsl.down.attenuation" < xdsl.json)
 
 echo multigraph bbox_adsl_snr
-echo up.value $(jq ".[].wan.xdsl.up.noise" < xdsl.json)
-echo down.value $(jq ".[].wan.xdsl.down.noise" < xdsl.json)
+echo up.value $(jq -r ".[].wan.xdsl.up.noise" < xdsl.json)
+echo down.value $(jq -r ".[].wan.xdsl.down.noise" < xdsl.json)
 
 echo multigraph bbox_adsl_pkt
-echo up.value $(jq ".[].wan.ip.stats.tx.packets" < ip-stats.json)
-echo down.value $(jq ".[].wan.ip.stats.rx.packets" < ip-stats.json)
-echo up_f.value $(jq ".[].wan.xdsl.stats.remote_fec" < xdsl-stats.json)
-echo down_f.value $(jq ".[].wan.xdsl.stats.local_fec" < xdsl-stats.json)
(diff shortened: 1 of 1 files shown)


plugins/router/bbox
```diff
@@ -0,0 +1,140 @@
...
+	echo uptime.label Uptime
+	echo uptime.cdef uptime,3600,/,24/
+
+	exit 0
+fi
+
+TMPFILE=$(mktemp -d)
+trap 'rm -Rf $TMPFILE' EXIT
+
+cd $TMPFILE
```
sumpfralle: Here you could also use quoting.

plugins/router/bbox
```diff
@@ -0,0 +1,140 @@
...
+	echo graph_category network
+	echo graph_vlabel days
+	echo uptime.label Uptime
+	echo uptime.cdef uptime,3600,/,24/
+
+	exit 0
+fi
+
+TMPFILE=$(mktemp -d)
+trap 'rm -Rf $TMPFILE' EXIT
```
sumpfralle: Here you should quote the `$TMPFILE` (double-quotes within the single-quotes).

plugins/router/bbox
```diff
@@ -0,0 +1,140 @@
...
+
+cd $TMPFILE
+
+curl -s http://$IP/api/v1/wan/ip > "ip.json" &
```
sumpfralle: Here you should quote the URL.

`shellcheck` is your friend :)

plugins/router/bbox
```diff
@@ -0,0 +1,140 @@
...
+curl -s http://$IP/api/v1/wan/ip/stats > "ip-stats.json" &
+curl -s http://$IP/api/v1/wan/xdsl > "xdsl.json" &
+curl -s http://$IP/api/v1/wan/xdsl/stats > "xdsl-stats.json" &
```
sumpfralle: Are you sure, that you need to store the result in files?
If you are quite certain, that the resulting data is small (compared to the currently expectable amount of RAM in a node), then I would recommend to store the data in a variable instead and get rid of temorary files handling.

plugins/router/bbox
```diff
@@ -0,0 +1,140 @@
...
+then
+	for i in *
+	do
+		echo "========= $i ========"
+		cat "$i"
+	done
+fi
+
+echo multigraph bbox_adsl_bw
+echo up.value $(jq ".[].wan.xdsl.up.bitrates" < xdsl.json)
```
sumpfralle: Are you sure, that all values are always numbers? Otherwise you should probably pass `-r` to `jq` in order to remove potential quotes from output?

plugins/router/bbox
```diff
@@ -0,0 +1,140 @@
...
+echo down.value $(jq ".[].wan.ip.stats.rx.packets" < ip-stats.json)
+echo up_f.value $(jq ".[].wan.xdsl.stats.remote_fec" < xdsl-stats.json)
+echo down_f.value $(jq ".[].wan.xdsl.stats.local_fec" < xdsl-stats.json)
+echo up_c.value $(jq ".[].wan.xdsl.stats.remote_hec" < xdsl-stats.json)
+echo down_c.value $(jq ".[].wan.xdsl.stats.local_hec" < xdsl-stats.json)
+echo up_u.value $(jq ".[].wan.xdsl.stats.remote_crc" < xdsl-stats.json)
+echo down_u.value $(jq ".[].wan.xdsl.stats.local_crc" < xdsl-stats.json)
+
+echo multigraph bbox_adsl_bytes
+echo up.value $(jq ".[].wan.ip.stats.tx.bytes" < ip-stats.json | tr -d '"')
```
sumpfralle: Maybe `-r`?

plugins/router/bbox
```diff
@@ -0,0 +1,140 @@
+#! /bin/sh
+# Parse bbox information
```
sumpfralle: This is a very, very short documentation header :)

Please excuse my ignorance, but I have no idea, what bbox means. Currently I think, it is some kind of network statistics collector. But initially I thought about spatial bounding boxes :)

Maybe you want to add a perldoc header?

plugins/router/bbox
steveschnepp: I usually prefer to use temp files in shell scripts. I'm therefore leveraging `mktemp -d` which can configured via setting `$TMPDIR`, to redirect it to a `tmpfs` filesystem.

steveschnepp:
Addressed your comments

sumpfralle:
Great!

PR #939, opened 2018-09-10
network.tc*: add the ability to detect systemd based network devices

RubenKelevra:
…ames

### DIFF
--- a/plugins/network/tc_
+++ b/plugins/network/tc_
@@ -61,7 +61,7 @@ case "$1" in
     suggest)
 	if [ -r /proc/net/dev ]; then
 	    awk '
-                /^ *(eth|tap|bond|wlan|ath|ra|sw)[0-9]+:/ {
+                /^ *(eth|tap|bond|wlan|ath|ra|sw|eno|ens|enp|wlp|wl)[0-9]*/ {
                    split($0, a, /: */);
                    gsub(/^ +/,"",a[1]);
                    if (($2 > 0) || ($10 > 0)) print a[1]; }' /proc/net/dev
--- a/plugins/network/tc_drops_
+++ b/plugins/network/tc_drops_
@@ -32,12 +32,10 @@ case $1 in
     suggest)
 	if [ -r /proc/net/dev ]; then
 	    awk '
-                /^ *(eth|tap|bond|wlan|ath|ra|sw)[0-9]+:/ {
+                /^ *(eth|tap|bond|wlan|ath|ra|sw|eno|ens|enp|wlp|wl)[0-9]*/ {
                    split($0, a, /: */);
                    gsub(/^ +/,"",a[1]);
                    if (($2 > 0) || ($10 > 0)) print a[1]; }' /proc/net/dev
-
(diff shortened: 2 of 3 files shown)


sumpfralle:
Thank you for taking care!

btw: did you add the "cake" plugin by mistake?

plugins/cake/cake_
```diff
@@ -0,0 +1,131 @@
...
+    config)
+	
+	echo "graph_title $DEVICE cake packet handling stats"
+	echo 'graph_args --base 1000'
+	echo 'graph_vlabel packets per second'
+	echo 'graph_category network'
+	echo "graph_info This graph shows the general packet handling status of egress traffic of the $DEVICE network interface."
+	
+	echo "backlog.label backlog";
+    echo "backlog.draw LINE2";
```
sumpfralle: Here you seem to mix intendation styles.

Just in case you are using vim: maybe the following is suitable for you?
```
set list
set listchars=tab:▸\ 
```

plugins/cake/cake_
```diff
@@ -0,0 +1,131 @@
...
+    autoconf)
+	if [ -r /proc/net/dev ]; then
+	    echo yes
+	    exit 0
+	else
+	    echo "no (/proc/net/dev not found)"
+	    exit 1
+	fi
+	;;
+    suggest)
```
sumpfralle: Just as a thought: personally I prefer multigraph plugins due to their ease of configuration (automatically following changes with regard to obsolete/new interfaces).
Maybe you want to turn this plugin into a multigraph plugin instead of a suggest-style plugin?
(just a matter of taste - the decision is up to you)

plugins/cake/cake_
```diff
@@ -0,0 +1,131 @@
...
+	if [ -r /proc/net/dev ]; then
+	    ifs="$(awk '
+                /^ *(eth|tap|bond|wlan|ath|ra|sw|eno|ens|enp|wlp|wl)[0-9]*/ {
+                   split($0, a, /: */);
+                   gsub(/^ +/,"",a[1]);
+                   if (($2 > 0) || ($10 > 0)) print a[1]; }' /proc/net/dev)"
+		cake_ifs=()
```
sumpfralle: Are you sure, that this array access works with `sh` (e.g. `dash`)?

plugins/cake/cake_
```diff
@@ -0,0 +1,131 @@
...
+}
+
+case "$1" in
+    autoconf)
+	if [ -r /proc/net/dev ]; then
+	    echo yes
+	    exit 0
+	else
+	    echo "no (/proc/net/dev not found)"
+	    exit 1
```
sumpfralle: The exitcode of an `autoconf` call [should be zero, even for "no"](http://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-concise.html?highlight=autoconf#autoconf).

RubenKelevra:
Thanks for your review, I saved it and will apply the ideas to the cake plugin... actually I wrote two, but as you can see, I guess I'm just too tired. Sorry for the mess. I'll open a new PR to commit the changes.

PR #940, opened 2018-09-10
network.tc*: add the ability to detect systemd based network devices

RubenKelevra:


sumpfralle:
Great!
And thank you for removing neighbouring cruft ...

PR #941, opened 2018-09-11
[WIP] cake plugins

RubenKelevra:
cake.cake_  renders the drops, overlimits and requeues as well as the backlog
cake_wlp3s0-day

cake.cake_tin_ renders the packets sent in each tin
cake_tin_wlp3s0-day
_This plugin can handle diffserv3, diffserv4 and diffserv8... while diffserv8 is somewhat incompatible with the others, since the tins are named differently._

@sumpfralle thanks for your initial feedback.

- No, the arrays won't work on POSIX shells, so I changed the shell required to bash.
- I thought again about the name. But it looks still fine to me.
- I really dislike those multigraph plugins, since they, well multigraph, which would be a mess with all those values. So I think I'll stick with the stacked area graphs per interface - thanks for your idea on that, anyway. :)
- The `exit 1` for autoconf is exactly like on network.tc_, so I like to stick with this for the moment. If you think, that's an issue, feel free to change my plugin as well, when you fix network.tc_


dtaht:
This is very cool. A couple notes:

0) I like seeing the backlog stat!
1) marks are happening more often now that apple deployed ecn. 

Having a daily stat would be awesome

RubenKelevra:
@dtaht 
the first plugin already supports backlog. The queue was just always empty, that's why it doesn't show up on that graphs.

The issue with this is, the value is just evaluated once in a 5 minute period, so it's not an average... (I currently have no idea how to fix this - without a service). But over a period of a day it should still show up a nice graph.

Marks is a great idea. :)

RubenKelevra:
I'm currently working on some more plugins after some reconsidering and feedback. :)

tohojo:
I would suggest using the JSON output of TC rather than trying to parse the human-readable output with regexes. Especially if you want to get at the tin stats...

Any version of TC that has support for CAKE is also going to support JSON output...

RubenKelevra:
@tohojo 
Hey Toke,

thanks for the idea, but parsing JSON in shellscript isn't an easy task. This is a task Python is suited much better for - which means a complete rewrite and quite a lot overhead for loading the python interpreter for each call.

I think I stick with simple screen parsing, also since the original network.tc_-plugin is written the same way.

Any thoughts on the graphs itself?

RubenKelevra:
The cake_tin_throughput_ looks quite boring, but should show some quite interesting graphs on systems with more traffic.

It combines the traffic limit set for cake with the average traffic for each tin.

cake_tin_throughput_wlp3s0-day

RubenKelevra:
@tohojo 
Well, no. Munin just calls executable files and grab the standard output. So you can build it in any language you like. Calling 10 python scripts every 5 minutes just to measure some numbers for one IF is likely an overkill.

> The backlog/dropped/overlimits/requeues graph is not CAKE-specific. Not sure how Munin organises this; is there a generic qdisc plugin as well / in addition to?

Yes, theres `tc_` in the network folder. But since it's designed for other output it probably won't output the stuff you like to see. It's also a lot bigger. I just wanted to see the drops and the backlog for cake, so I cutted the `tc_`-plugin down, to just graph those. Munin handles this change fine, since the plugin has a different name.

> Are you handling a varying number of tins correctly?

Those plugins read the settings with the keywords listed by tc, so diffserv3, diffserv4 and diffserv8 are implemented. They are named like the tc output, so for diffserv8 it's Tin 1, Tin 2... etc.

tohojo:
"@RubenKelevra" <notifications@github.com> writes:

> @tohojo 
> Well, no. Munin just calls executable files and grab the standard
> output. So you can build it in any language you like. Calling 10
> python scripts every 5 minutes just to measure some numbers for one IF
> is likely an overkill.

Right, fair enough; you're the one who has to update the script if the
output parsing breaks, so obviously your call ;)

>> The backlog/dropped/overlimits/requeues graph is not CAKE-specific.
>> Not sure how Munin organises this; is there a generic qdisc plugin as
>> well / in addition to?
>
> Yes, theres `tc_` in the network folder. But since it's designed for
> other output it probably won't output the stuff you like to see. It's
> also a lot bigger. I just wanted to see the drops and the backlog for
> cake, so I cutted the `tc_`-plugin down, to just graph those. Munin
> handles this change fine, since the plugin has a different name.

Right. Well, mostly a UI issue, I guess; and up to the Munin upstream if
they want two different-but-related plugins :)

>> Are you handling a varying number of tins correctly?
>
> Those plugins read the settings with the keywords listed by tc, so
> diffserv3, diffserv4 and diffserv8 are implemented. They are named
> like the tc output, so for diffserv8 it's Tin 1, Tin 2... etc.

Cool.

What happens if the CAKE configuration changes while the same monitor
instance is running?

dtaht:
I would plot ecn and drops on the same graph. And a five minute summary of those is probably fine for many systems. We really don't drop all that much. Can a log scale be used? In which case you could actually include packets. cool work, exciting! I wonder if snmp has a mib that includes ecn....

heistp:
Cool, I might use of this at an ISP in the process of trying to get cake deployed. I was hoping we'd be able to look at some of the cake tin stats alongside the SmokePing graphs.

I do second the idea that using `jq` might make the parsing more robust, and it's pretty lightweight, but I also understand if you're not ok with the dependency. Thanks for the effort!

RubenKelevra:
@tohojo 

> Right, fair enough; you're the one who has to update the script if the
output parsing breaks, so obviously your call ;)

I'm fine with this. ;)

> Right. Well, mostly a UI issue, I guess; and up to the Munin upstream if
they want two different-but-related plugins :)

Munin-Plugins are separated from this community contribution repo and an official one. So the barrier for community plugins is lower. :)

> What happens if the CAKE configuration changes while the same monitor
instance is running?

The graph is instantly reconfigured. This means old values are probably disappearing, cause they are incompatible and the new ones appear.

RubenKelevra:
@dtaht 

> I would plot ecn marks and drops on the same graph.

yes, thats an good idea. I'll implement it that way, today.

> Can a log scale be used?

Yep, that's possible. So you want packets, drops and ecn marks in one graph? That could be too much if you consider 8 tins... 

I guess individual graphs would be better than. We could use logarithm scale and a stacked graph as area for packets and drops, so you get easily the proposition and a line for ecn marks. That's the way the first graph is build, backlog is a line and everything else is a stacked area.

But I'm not sure about the direction, packets at the bottom, and a line to see how many packets are ecn marked?

> cool work, exciting!

Graphs of stuff you build are always exiting. Great to hear that from you, too. :)

RubenKelevra:
@heistp

> Cool, I might use of this at an ISP in the process of trying to get cake deployed. I was hoping we'd be able to look at some of the cake tin stats alongside the SmokePing graphs.

> I do second the idea that using jq might make the parsing more robust, and it's pretty lightweight, but I also understand if you're not ok with the dependency. Thanks for the effort!

Hey cool!

I would recommend to deploy it in a test environment first, to see if the graphs are working right. I'm especially worried about cake_tin_throughput_. It "should" work the way it's inteded, but I haven't found the time to check it with some synthetic throughput to measure that it's right.

I guess it would be also great if we can get some graphs with real traffic back, for a subpage on bufferbloat.net to show how it's working. :)

RubenKelevra:
I push today a change which makes the plugins able to detect the corresponding ingress device, by parsing `tc filter` output of the configured device for the plugin one.

If the ifb device is found, the graph is reconfigured to show the ingress traffic below, like you're used to by the traffic graphs.

Note that Munin is might not capable of handling reconfiguring graphs as expected. It could be wise to delete the corresponding rrd files after updating your plugin files with something like this:

`# rm -f /var/lib/munin/myhostname/myhostname-cake_*`

Here are some examples:

cake_wlp3s0-day
cake_tin_pps_wlp3s0-day
cake_tin_delay_wlp3s0-day

RubenKelevra:
@dtaht I decided to make a multi-graph for ecn, so all tins combined and if you click on it you get up to 8 graphs for each tin. That's the only way to get a useful graph out of all this data.

This might take a bit longer than anticipated, it's my first multi-graph :)

There are two values which would be helpful for graphs which are currently not shown:

- Packets in the backlog per tin - without it I cannot display backlog, since we're talking about packets not bandwidth.

- Cake can separate between hosts and flows at the same time. It would be interesting to get a hosts reading additionally to the flows one.

sumpfralle:
Hello Ruben,

thanks for this great and detailed data collection!

> No, the arrays won't work on POSIX shells, so I changed the shell required to bash.

Personally I would really like to use these plugins in the OpenWrt routers of our local wireless community. Thus I would appreciate `sh`-only plugins.

> I thought again about the name. But it looks still fine to me.

I would agree, if this plugin would cover *all* aspects of cake. But for now (as I understand it) it only tracks specific data. Thus you should rename it appropriately.

> I really dislike those multigraph plugins, since they, well multigraph, which would be a mess with all those values. So I think I'll stick with the stacked area graphs per interface - thanks for your idea on that, anyway. :)

Allow me another attempt of persuading you :)
There are two styles of multigraph plugins:
* nested plugins with two levels: an overview graph and multiple detailed/split graphs (you use this for `ecn`)
* multiple flat graphs in one plugin: I really think, this would be worthwhile for the whole set of plugins, since they share a lot of code. In this case you would just put all the configs and data into one plugin and prefix each block with something like `multigraph cake_foo`. This is quite easy to maintain, since it reduces a lot of code duplication.

> The exit 1 for autoconf is exactly like on network.tc_, so I like to stick with this for the moment. If you think, that's an issue, feel free to change my plugin as well, when you fix network.tc_

The specification of that exit code is quite clear. Please do no deviate on purpose.

Regarding the parsing of JSON instead of using the human-oriented output: please follow the advice of the other commentators and use `jq`. This is quite common among a lot of munin plugins and *really* eases long-term maintenance of plugins.

plugins/cake/cake_
```diff
@@ -0,0 +1,202 @@
...
+=head1 LICENSE
+
+GPLv2 or later
+
+=cut
+
+DEVICE=${0##*/cake_}
+
+tc_get_ifb_dev() {
+	dev="$(/sbin/tc filter show dev $DEVICE parent ffff: protocol all | grep "mirred" | grep "Egress Redirect to device")"
```
sumpfralle: Please use `local` for variables that are only used in functions.
Afterwards you will need to split the assignment from the `local` line in order to not hide the exit code of the command substitution.

Please quote variables (e.g. `$DEVICE`). `shellcheck` is your friend :)

`grep "mirrored"`: `-w` would be more specific here. Or maybe join both `grep` calls into one.

plugins/cake/cake_
```diff
@@ -0,0 +1,202 @@
...
+tc_cake_backlog() {
+	/sbin/tc -s qdisc show dev "$1" | grep -E "^ backlog" | tr 'p' ' ' 
+}
+
+DEVICE_IN="$(tc_get_ifb_dev)"
+has_if_in=$?
+
+case "$1" in
+	autoconf)
+	if [ -r /proc/net/dev ]; then
```
sumpfralle: Is there really no specific test for the actual use of `cake` on the system?

plugins/cake/cake_
```diff
@@ -0,0 +1,202 @@
...
+				cake_ifs+=("$if")
+			fi
+		done
+		echo "$cake_ifs"
+	fi
+	exit 0
+	;;
+	config)
+	
+	if [ $has_if_in -eq 0 ]; then
```
sumpfralle: Please quote variables. We are striving for `shellcheck` cleanliness in the plugins.
See the [hints for new plugins](https://github.com/munin-monitoring/contrib/tree/master/plugins#submit-a-new-plugin).

plugins/cake/cake_
```diff
@@ -0,0 +1,202 @@
...
+	print "dropped.value " $7
+	print "overlimits.value " $9
+	print "requeues.value " $11
+}'
+
+tc_cake_backlog "$DEVICE" | awk '{
+	print "backlog.value " $3
+}'
+
+if [ $has_if_in -eq 1 ]; then
```
sumpfralle: The logic of conditional should be clearer for the next reader, if you invert the expression and place the two blocks that are only relevant in case of an existing interface into the `then` branch.

plugins/cake/cake_tin_delay_
```diff
@@ -0,0 +1,377 @@
...
+	/sbin/tc -s qdisc show dev "$1" | head -n1 | grep -i "diffserv8" > /dev/null
+	if [ $? -eq 0 ]; then
+		echo 8
+	fi
+	return 0
+}
+
+convert_time() {
+	#converts us, ms, s into s without a unit
+	echo $1 | awk '{
```
sumpfralle: Please quote variables.

plugins/cake/cake_tin_delay_
```diff
@@ -0,0 +1,377 @@
...
+			time *= 3600;
+		} else if (unit == "d") {
+			time *= 86400;
+		}
+		printf("%.9f\n", time)
+	}'
+}
+
+tc_cake_get_delays() {
+	delays="$(/sbin/tc -s qdisc show dev "$1" | grep "^  av_delay" | sed -e 's/av_delay//')"
```
sumpfralle: Please use `local`.

plugins/cake/cake_tin_delay_
```diff
@@ -0,0 +1,377 @@
...
+	for e in ${delays[@]}; do
+		delay="$(convert_time $e)"
+		echo -en " $delay "
```
sumpfralle: What is the purpose of `-e` here?
Regarding `-n`: "printf" should be more portable (in case you switch from bash to sh).

plugins/cake/cake_tin_delay_
```diff
@@ -0,0 +1,377 @@
...
+}
+
+DEVICE_IN="$(tc_get_ifb_dev)"
+has_if_in=$?
+
+diffserv_no="$(tc_cake_get_diffserv "$DEVICE")"
+if [ $has_if_in -eq 0 ]; then
+	diffserv_in_no="$(tc_cake_get_diffserv "$DEVICE_IN")"
+	
+	if [ ! "$diffserv_no" == "$diffserv_in_no" ]; then
```
sumpfralle: `!=` is probably easier to read?

plugins/cake/cake_tin_delay_
```diff
@@ -0,0 +1,377 @@
...
+		print "tin7_in.value " $7
+		print "tin8_in.value " $8
+		}'
+	fi
+	exit 0
+fi
+
+
+exit 1
+
```
sumpfralle: Please remove trailing whitespace.

plugins/cake/cake_tin_delay_
```diff
@@ -0,0 +1,377 @@
...
+		print "tin5_in.value " $5
+		print "tin6_in.value " $6
+		print "tin7_in.value " $7
+		print "tin8_in.value " $8
+		}'
+	fi
+	exit 0
+fi
+
+
```
sumpfralle: Please add an error message instead of silently failing.

plugins/cake/cake_tin_ecn_
```diff
@@ -0,0 +1,365 @@
+#!/bin/bash
```
sumpfralle: The plugins `cake_tin_throughput_`, `cake_tin_ecn_`, `cake_tin_delay_` and `cake_tin_pps_` seem to share almost all of their code.
I would really like to avoid this code duplication.
How could you merge them? Maybe by using a multigraph plugin? :)

plugins/cake/cake_
tohojo: Not in all cases. You could check if the sch_cake module is loaded, but if it's compiled as a built-in module that won't catch it. Also, the module can be loaded without being active.

The only sure way to see if cake is active is to do a `tc qdisc` and loop through all interfaces to try to find it. You could probably do `tc -j qdisc` and construct a query with jq that will spit out all cake instances...

sumpfralle: How about this one?
```shell
tc -j qdisc | jq -r '.[] | select(.kind == "cake").dev' | sort | uniq
```

tohojo:
Lars Kruse <notifications@github.com> writes:

> sumpfralle commented on this pull request.
>
>
>
>> +	echo "$dev" | sed -n 's/.* device \([^ ]*\).*/\1/p' | tr ')' ' ' | awk '{ print $1 }'
> +}
> +
> +tc_cake_sent() {
> +	/sbin/tc -s qdisc show dev "$1" | grep -E "^ Sent" | tr ',' ' ' | tr ')' ' ' 
> +}
> +tc_cake_backlog() {
> +	/sbin/tc -s qdisc show dev "$1" | grep -E "^ backlog" | tr 'p' ' ' 
> +}
> +
> +DEVICE_IN="$(tc_get_ifb_dev)"
> +has_if_in=$?
> +
> +case "$1" in
> +	autoconf)
> +	if [ -r /proc/net/dev ]; then
>
> How about this one?
> ```shell
> tc -j qdisc | jq -r '.[] | select(.kind == "cake").dev' | sort | uniq
> ```

Yup, something like that :)

`sort | uniq` can be replaced with just `sort -u`; but I'm not sure it's
even needed in this case? If we're just checking whether there's an
active interface there's no need to unique-ify the output. And besides,
the only case where the same interface would appear more than once is if
there's a setup with multi-queue, or an HTB structure with more than one
instance of cake. And I don't think the rest of the script handles that
anyway?

plugins/cake/cake_
RubenKelevra: Thanks for the heads up, haven't thought about that.

RubenKelevra: To the regex idea: I'm not firm enought to use it here - I don't use stuff I don't fully understand.

plugins/cake/cake_
RubenKelevra: shellcheck has 255 MB dependencies. I'm sorry, I won't bloat my system with this tool.

plugins/cake/cake_
RubenKelevra: This looks great, thanks!

RubenKelevra:
Hey guys,

I'll come back at a later state and read carefully all comments on the code itself and also probably combine the plugins to one plugin with customization options, but currently, I'm more focused on getting the functionality fully setup and error-free.

If you want to help me on this, feel free to test the functionality, especially with multi-device setups and check I need to check cake_tin_throughput_ for consistency, since this is written into the blue...


Best regards

Ruben

RubenKelevra:
I've created a color scheme for the graphs. I had tested this with some friend, to see if a non-colorblind person, other than me, can distinguish the colors easily and did some changes based on this.

cake_tin_throughput_wlp3s0-day 2

I like to have some feedback on this as well, if you got some time. :)

tohojo:
I'd suggest taking a look at colorbrewer: http://colorbrewer2.org/#type=qualitative&scheme=Accent&n=3

I'm using the second of the qualitative schemes for Flent graphs (https://github.com/tohojo/flent)

sumpfralle:
@tohojo:
> `sort | uniq` can be replaced with just `sort -u`

yes, indeed. My personal preference is the former due to being easy to understand for all readers. Purely a matter of taste :)

> And besides, the only case where the same interface would appear more than once is if there's a setup with multi-queue, or an HTB structure with more than one instance of cake. And I don't think the rest of the script handles that anyway?

I do not know and understand) the output of `tc`, but I just noticed that the wireless interface (running as a client) on my machine exposes four items wih the same `dev` (seemingly only differing in the `parent` field).

@RubenKelevra:
> I'll come back at a later state and read carefully all comments on the code itself and also probably combine the plugins to one plugin with customization options, but currently, I'm more focused on getting the functionality fully setup and error-free.

That sounds perfect to me! Happy hacking!

plugins/cake/cake_
sumpfralle: I *guess* `grep "Egress Redirect to device.*mirrored"` would work here.
(I cannot tell, since my machine does not seem to use cake by default)
Anyway: the current state is also fine.

RubenKelevra:
@tohojo thanks for the advice, but I already started there. I had trouble to identify some lighter colors on 'night mode' on some displays, so I started to modify it.

Afterwards I sent it to some friends, which reported some issues with greenish/bluish color differentiation which I solved by introducing a wider range of colors for the 8 tin graphs.

For the cake_ graph I'll use some standard colors of the new munin standard colors (don't see a point to deviate there)

RubenKelevra:
I've run shellcheck, after I found a custom package avoiding hundreds of megabytes dependencies. I would call those changes bonkers. But since I've quoted "device" in some files on some occasions, I added double quotes everywhere by goodwill. I hope everyone is now fine with this.

plugins/cake/cake_
RubenKelevra: @sumpfralle I doubt that this would work, since the line reads like:

`action order 1: mirred (Egress Redirect to device ifb1) stolen`

mirrored isn't contained ;)

plugins/cake/cake_
RubenKelevra: So... I'll leave it the way it it. If you want to replace this function, feel free to suggest one which covers all features.

Also make sure to add a simple test, if jq is installed with an error-output for autoconf - for example it isn't installed on my system - never had the need for it - so an output about this dependency would be nice.

plugins/cake/cake_
RubenKelevra: Well, shellcheck isn't complaining about this line, what exactly is wrong with it?

integer comparisons AFAIK cannot be double quoted and since it's not an input variable I don't see any point changing this to a string, just to make any sanitizer happy.

plugins/cake/cake_
RubenKelevra: Why do you would want to invert this?

plugins/cake/cake_tin_delay_
RubenKelevra: done

plugins/cake/cake_tin_delay_
RubenKelevra: I feel the other way around for those cases...

"If not a is b" is for me the easiest readable solution.

plugins/cake/cake_tin_delay_
RubenKelevra: This feels quite nitpicky :D

plugins/cake/cake_tin_delay_
RubenKelevra: alright

plugins/cake/cake_tin_ecn_
RubenKelevra: Can I generate multiple level 0 graphs with one multigraph plugin? With different nested level 1 graphs?

plugins/cake/cake_
sumpfralle: > But do we really want to get rid of the filter with packets received?

I am not sure, what this means. How would you detect this situation? I would be happy to suggest a suitable filter.

Regarding the check of for `jq` - this is simply part of "autoconf", too:
```shell
if which jq >/dev/null; then
    echo "no (missing 'jq' executable)"
elif [ -r /proc/net/dev ]; then
...
```

plugins/cake/cake_
sumpfralle: I am not sure, if you meant this serious, but I will just follow your invitation for spoon feeding :)
```shell
grep "mirred.*Egress Redirect to device"
```

Additionally this would be a great place for adding a comment with this expected example output line. This will ease the life of future readers of the code.

And I have to admit, that this line of output looks quite human-centric and thus will *really change* somewhen. `json` would be your friend :)

plugins/cake/cake_
sumpfralle: > Well, shellcheck isn't complaining about this line, what exactly is wrong with it?

Correct: in this case it does not complain, since you are assigning only fixed strings to this variable. Overly simple static parsers (e.g.: humans like me) will repeatedly stumble over this, nevertheless :)

> integer comparisons AFAIK cannot be double quoted

Please rest assured, that the shell parser will *never* forbid you to double-quote a variable :)
(double quotes only prevent word splitting for various expansions - this is usually what you want)

plugins/cake/cake_
sumpfralle: You are using an approach like `early return` here: the conditional secretly applies to all following lines.
Thus a casual contributor, who would add another graph at the end of the plugin would be surprised, that it is not included under some circumstances. Keeping the scope of a conditional small and obvious is usually a good thing :)

plugins/cake/cake_tin_delay_
sumpfralle: Your plugins do work already - thus almost every comment of mine is nitpicking :)

plugins/cake/cake_tin_ecn_
niclan: Yes, you can populate any namespace you like from a multigraph plugin.

RubenKelevra: @niclan thanks, this was very helpful. :)

plugins/cake/cake_
RubenKelevra: It's quite early in the morning, but I don't really get what you want. 🤔

There are currently very small blocks of output conditional, not big chunks. How do you want me to change this?

plugins/cake/cake_
RubenKelevra: ;)

And good idea :)

gamanakis:
@RubenKelevra thanks for this!
I set it up on my router, but some scripts cannot read some of the stats.

cake_tin_delay:
```
bulk.value 0.000039000
besteffort.value
voice.value
```

cake_tin_throughput:
```
bandwidth.value 19000000
bulk.value 30725536
besteffort.value
voice.value

```

plugins/cake/cake_
sumpfralle: Hehe - yes, that works. I am writing way too much shell code for busybox's ash. All the usual features of `dash` are supported there.

plugins/cake/cake_
sumpfralle: Thank you!
Next time I will try to be more precise ...

plugins/cake/cake_
sumpfralle: OK - I see.
Maybe we have different perceptions regarding the purpose of the "autoconf" test.
Here the plugin needs to answer two questions:
* can this plugin run at all? (i.e. are all dependencies satisfied?)
* is this plugin useful for this host?

The "autoconf" call should return "yes" only if both questions are affirmed.

Your current implementation of "autoconf" only cares (partly) for the first question.
Additionally it should check two more details:
* is `tc` available? (see my example for `jq` above)
* is "cake" in use on this system?

The latter can probably be tested like this:
```shell
[ -n "$(tc -j qdisc | jq -r '.[] | select(.kind == "cake").dev')" ]
```

RubenKelevra:
@sumpfralle

> Personally I would really like to use these plugins in the OpenWrt routers of our local wireless community. Thus I would appreciate sh-only plugins.

What do you think about a cut down version in a personal repo as an openwrt feed? This would allow to develop this plugins without any restrictions and the sh-ones to be much smaller - I guess you don't want to run munin-node anway, so a more simple output you can push per http or similar would be enough, right?

RubenKelevra:
@gamanakis

thanks for the feedback, can you please quote the output of `tc -s qdisc show dev YOURDEV` for debug purposes, please? thanks!

sumpfralle:
>> Personally I would really like to use these plugins in the OpenWrt routers of our local wireless community. Thus I would appreciate sh-only plugins.
>
> What do you think about a cut down version in a personal repo as an openwrt feed? This would allow to develop this plugins without any restrictions and the sh-ones to be much smaller [..]

Using ash/dash-compatible features (instead of bash) is in my experience rarely a real restriction.
Only 17 out of 161 of munin's core shell plugins use bash. Thus it does not seem to be a tough challenge :)

> I guess you don't want to run munin-node anway, so a more simple output you can push per http or similar would be enough, right?

There is the excellent `muninlite` available in OpenWrt. I use it a lot.

But I can understand, that you may want to stick to the environment you are comfortable with. In this case I will propose changes (after your plugin is merged), that will make it run with sh-only.

RubenKelevra:
> > > Personally I would really like to use these plugins in the OpenWrt routers of our local wireless community. Thus I would appreciate sh-only plugins.
> > 
> > 
> > What do you think about a cut down version in a personal repo as an openwrt feed? This would allow to develop this plugins without any restrictions and the sh-ones to be much smaller [..]
> 
> Using ash/dash-compatible features (instead of bash) is in my experience rarely a real restriction.
> Only 17 out of 161 of munin's core shell plugins use bash. Thus it does not seem to be a tough challenge :)
> 
> > I guess you don't want to run munin-node anway, so a more simple output you can push per http or similar would be enough, right?
> 
> There is the excellent `muninlite` available in OpenWrt. I use it a lot.
> 
> But I can understand, that you may want to stick to the environment you are comfortable with. In this case I will propose changes (after your plugin is merged), that will make it run with sh-only.

Yeah, I'm comfortable with `sh` it's just that it sometimes looks ugly :D 

This, for example, is from me for sh:
https://github.com/VfN-NRW/offline-ssid/blob/master/gluon-offline-ssid/files/usr/sbin/ff-offline-ssid

RubenKelevra:
> I am runnning sch_cake/tc from repos:
> github.com/dtaht/sch_cake
> github.com/dtaht/tc-adv
> 
> ```
> root@arch:~# tc -s qdisc show dev br-lan
> qdisc cake 800e: root refcnt 2 bandwidth 18Mbit diffserv3 dual-dsthost ingress split-gso rtt 100.0ms noatm overhead 18 mpu 64 
>  Sent 3458265254 bytes 2370831 pkt (dropped 1417, overlimits 4457648 requeues 0) 
>  backlog 0b 0p requeues 0
>  memory used: 582625b of 4Mb
>  capacity estimate: 18Mbit
>  min/max network layer size:           28 /    1500
>  min/max overhead-adjusted size:       64 /    1518
>  average network hdr offset:           14
> 
>                    Bulk  Best Effort        Voice
>   thresh       1125Kbit       18Mbit     4500Kbit
>   target         16.1ms        5.0ms        5.0ms
>   interval      111.1ms      100.0ms      100.0ms
>   pk_delay        7.3ms         88us        253us
>   av_delay        1.6ms         15us         12us
>   sp_delay          2us          2us          2us
>   backlog            0b           0b           0b
>   pkts          2356277        11172         4799
>   bytes      3458013591      2041653       341452
>   way_inds         4425          140            0
>   way_miss         4865         7942           25
>   way_cols            0            0            0
>   drops            1415            2            0
>   marks              32            0            0
>   ack_drop            0            0            0
>   sp_flows           25            3            1
>   bk_flows            2            0            0
>   un_flows            0            0            0
>   max_len         21196        13626         6056
>   quantum           300          549          300
> 
> root@arch:~# ./cake_tin_delay_br-lan 
> bulk.value 0.001600000
> besteffort.value 
> voice.value 
> ```

I don't have a good answer why it doesn't work... well... @sumpfralle do you got an idea?

RubenKelevra:
@gamanakis 
Thanks for your report, found the bug and it should be fixed in 8cc759a0c6a617bb01eec42bfbcbd81849e7de97

RubenKelevra:
@sumpfralle I found some rendering issues after switching to logarithmic scale (I suspect). Maybe you got an idea about that:

cake_tin_pps_wlp3s0-day__
cake_wlp3s0-day__

One direction isn't rendered properly but cut off, while the values below definitely show, that the values are collected correctly.

tohojo:
You can't show negative values on a log scale...

sumpfralle:
How is your progress?
Can I help you somehow?

dtaht:
Yes, I'm rooting for you here, too! Can I help?

sumpfralle:
Ping?

sumpfralle:
I am willing to help, if something is missing.What needs to be done?

RubenKelevra:
@dtaht @sumpfralle sorry guys those notifcations got somehow lost between work. I'm currently working on a small thing for the next 1-2 days, after this I'll complete this pull request :)

RubenKelevra:
But good thing: The plugins still works, they run unchanged fine. This are the latest graphs I got (I disabled munin a while ago)

cake_tin_delay_wlp3s0-week
cake_tin_throughput_wlp3s0-week
cake_wlp3s0-week

dtaht:
Neat. Tracking marks vs drops is becoming interesting in light of L4S.

sumpfralle:
@RubenKelevra: ping?

Do you need help with some finer details?

RubenKelevra:
@sumpfralle thanks for the offering, my workload just shifted massively due to the current crisis. It's on my to-do list, but I just got no time 🙄

dtaht:
There's kind of a related discussion about putting some stats in luci, also, on the openwrt forums: https://forum.openwrt.org/t/sqm-reporting/59960/5

dtaht:
there is kind of a related discussion about cake and luci over here: https://forum.openwrt.org/t/sqm-reporting/59960/5

sumpfralle:
Ping?

dtaht:
Is this project still alive?

PR #945, opened 2018-09-24
Add plugin for graphing zfs user/group space usage

kienanstewart:
It's fairly similar to zfs_usage_, but handles userspace and/or groupspace.

### DIFF
--- a/plugins/zfs/zfs_space_
+++ b/plugins/zfs/zfs_space_
@@ -1,0 +1,142 @@
+#!/usr/bin/env perl
+# -*- perl
+
+=pod
+
+=head1 NAME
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for extending the zfs plugin and for describing the source, this plugin is based upon.

Could you please elaborate, how your new plugin differs from the original one?
I would *really* like to avoid adding code copies (they are a hell of work to maintain over the years). Thus maybe we can find a way to merge your changes into the original plugin?

sumpfralle:
Ping?

sumpfralle:
Ping?

PR #947, opened 2018-09-25
Add plugin for monitoring GPU usage per user

htakano:
It's similar to nvidia_gpu_, but shows each users' GPU usage graph.

### DIFF
--- a/plugins/gpu/nvidia_gpu_by_user
+++ b/plugins/gpu/nvidia_gpu_by_user
@@ -43,8 +43,7 @@ GPLv2
 
 EOF
 
-# . "$MUNIN_LIBDIR/plugins/plugin.sh"
-. /usr/share/munin/plugins/plugin.sh
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
 
 # Get location of nvidia-smi executable or use default
 nvSmiExec=${smiexec:-'/usr/bin/nvidia-smi'}
--- a/plugins/gpu/nvidia_gpu_by_user
+++ b/plugins/gpu/nvidia_gpu_by_user
@@ -6,13 +6,24 @@
 
 =head1 NAME
 
-gpubyuser - Plugin to monitor GPU memory usage by user
+nvidia_gpu_by_user - Plugin to monitor GPU memory usage by user.
 
 =head1 CONFIGURATION
 
 Add this to node configuration file:
  [nvidia_gpu_by_user]
   env.smiexec - Location of nvidia-smi executable.
+  env.gpuusers - List of the username to monitor(space separated).
+
+=head1 USAGE
+
+Example:
+ [nvidia_gpu_by_user]
+  env.smiexec /path/to/nvidia-smi
+  env.gpuusers root hideki
+
+If env.gpuusers is set, graph always shows listed users.
+Otherwise graph shows users that have been using GPUs only.
 
 =head1 AUTHOR
 
@@ -32,10 +43,14 @@ GPLv2
 
 EOF
 
-. "$MUNIN_LIBDIR/plugins/plugin.sh"
+# . "$MUNIN_LIBDIR/plugins/plugin.sh"
+. /usr/share/munin/plugins/plugin.sh
 
 # Get location of nvidia-smi executable or use default
 nvSmiExec=${smiexec:-'/usr/bin/nvidia-smi'}
+
+# Get gpuusers
+gpuUSERS=${gpuusers:-""}
 
 # Check if autoconf was requested
 if [ "$1" = "autoconf" ]; then
@@ -48,8 +63,6 @@ if [ "$1" = "autoconf" ]; then
 		exit 0
 	fi
 fi
-
-gpuUSERS=$(clean_fieldname "$(ls /home)" | tr "\n" " ")
 
 # GPU usage
 smiOutput=$("$nvSmiExec" -q)
@@ -66,6 +79,7 @@ echo "$smiInfo" | \
 BEGIN {
  n=-1;
  split("", gpu);
+ stderr="/dev/stderr"
 }
 
 $0 ~ "^Product Name" {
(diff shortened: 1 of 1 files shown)


plugins/gpu/nvidia_gpu_by_user
```diff
@@ -0,0 +1,240 @@
...
+	if [ -x "$nvSmiExec" ]; then
+		echo yes
+		exit 0
+	else
+		echo "no (nvidia-smi executable not found)"
+		exit 0
+	fi
+fi
+
+gpuUSERS=$(clean_fieldname "$(ls /home)" | tr "\n" " ")
```
sumpfralle: This feels slightly dirty to me, since it uses the non-documented behaviour of `clean_fieldname` with regard to multi-line input. Please use something like the following instead:
```shell
find /home -maxdepth 1 -mindepth 1 -type d | while read -r dname; do clean_fieldname "$(basename "$dname")"; done
```

Btw.: is the underlying assumption (relevant users having a home directory below `/home/` that matches their name) a good choice?
How about the following - in order to get at least around the directory/username match requirement?
```shell
getent passwd | cut -d : -f 1,6 | grep ":/home/" | cut -d : -f 1
```
(I cannot judge its portability - maybe you can comment)

plugins/gpu/nvidia_gpu_by_user
```diff
@@ -0,0 +1,240 @@
...
+}
+
+function getUserCountInGpu(_n) {
+ j = 0;
+ while (gpu["user", _n, j] != "") {
+  j++;
+ }
+ return j;
+}
+
```
sumpfralle: Please remove trailing whitespace.

plugins/gpu/nvidia_gpu_by_user
```diff
@@ -0,0 +1,240 @@
...
+$0 ~ "^Used GPU Memory" {
+ split ($2, tmp, " ");
+ if (gpu["used", n, m] == "") gpu["used", n, m] = tmp[1];
+ else gpu["used", n, m] += tmp[1];
+}
+
+END {
+ if (n < 0) {
+
+  print "No NVIDIA GPUs detected. Exiting."
```
sumpfralle: Please print this to stderr and exit with an error.

plugins/gpu/nvidia_gpu_by_user
```diff
@@ -0,0 +1,240 @@
+#!/bin/sh
+#  -*- sh -*-
+
+: <<EOF #
+ =cut
+
+=head1 NAME
+
+gpubyuser - Plugin to monitor GPU memory usage by user
```
sumpfralle: By convention, this is the full name of the plugin file.

plugins/gpu/nvidia_gpu_by_user
niclan: In some environments "getent passwd" will not return users in LDAP (or even YP/NIS):
```
$ getent passwd | grep niclangf
$ getent passwd niclangf
niclangf:*:7181:10001:niclangf:/home/niclangf:/bin/bash
```
and in some environments "getent passwd" will return many many thousands of users.

What you find in /home is also very dodgy. Home directories can be mounted many other places in large environments.

So getting a "full catalog" of users is not very reliable or scalable.  But if you have either a username or a UID this works

```
$ getent passwd niclangf
niclangf:*:7181:10001:niclangf:/home/niclangf:/bin/bash
$ getent passwd 7181
niclangf:*:7181:10001:niclangf:/home/niclangf:/bin/bash
```

If you want to provide a stable list of users to keep the graphs stable then persist the usernames... but I see I never got around to writing persistence functions for plugin.sh... There are good hints in Munin/Plugin.pm about what environment variables and such to use if you feel wizardly and want to write a file to persist something at least.

plugins/gpu/nvidia_gpu_by_user
niclan: Otherwise excellent plugin!

plugins/gpu/nvidia_gpu_by_user
sumpfralle: So what could be a good and portable approach for requesting "all non-system users" (I think, this is the goal)?
I have the feeling, that we cannot provide a really good and platform neutral implementation with the little effort that is justified for this task in this context. Thus I would be happy with anything that does not feel *too* dirty (maybe even the original proposal).

niclan: There isn't.  Just graph for those that have been using the GPU and somehow persist it.

htakano: Thank you for good comments.
If there isn't suitable implementation for listing all non-system users,
then I think it is still good for users to be configurable.

htakano:
I've fixed some parts based on your feedback.

sumpfralle:
It looks good to me, now!

Just one question:

> Otherwise graph shows users that have been using GPUs only.

Currently `gpuUSERS` defaults to an empty string. This does not match the above description, or?
Since this plugin supports `autoconf` it should work with reasonable defaults without manual intervention. Thus I would suggest one of the following approaches:
* try to find out, which accounts are using the GPU right now (ideally: persist this list below `MUNIN_PLUGSTATE`)
* try to assemble a reasonable list of possible users (e.g. `find /home -mindepth 1 -maxdepth 1 -type d -print0 | xargs -0 -r -n 1 basename`)

htakano:
> Currently gpuUSERS defaults to an empty string. This does not match the above description, or?
How about this comment below? 

If env.gpuusers is set, graph always shows listed users
(root, hideki in example) whether using GPU or not.
Otherwise, graph shows users that are using the GPU right now only.

> try to find out, which accounts are using the GPU right now (ideally: persist this list below MUNIN_PLUGSTATE)
> it should work with reasonable defaults without manual intervention. Thus I would suggest one of the following approaches:

For now, no suitable/portable method is found to list all non-system users automatically,
(find directories right under /home or getent would be at risk of listing thousands of accounts)
so I think it's better to be configurable.

sumpfralle:
> Otherwise, graph shows users that are using the GPU right now only.

Are you sure, that the plugin currently really does this?
As far as I understand the awk code, it looks like it will print no values at all, if the environment variable is empty. But maybe I misunderstand the code?

> For now, no suitable/portable method is found to list all non-system users automatically,
(find directories right under /home or getent would be at risk of listing thousands of accounts)
so I think it's better to be configurable.

That sounds reasonable.
Please just keep in mind, that if the plugin supports `autoconf`, then it should work without a specific configuration (i.e. the graph should not be empty). I am not sure, whether this is currently its behaviour due to my question above.

sumpfralle:
Maybe you want to comment on my last question?

Thank you!

sumpfralle:
@htakano: ping?

sumpfralle:
Ping?

PR #948, opened 2018-09-26
nvme - Munin plugin to monitor the use of NVMe devices

kjetilho:
This is a multigraph plugin which makes three graphs, nvme_usage, nvme_bytes and nvme_writecycles.  See POD below for details.

### DIFF
--- a/plugins/disk/nvme
+++ b/plugins/disk/nvme
@@ -1,0 +1,244 @@
+#! /usr/bin/perl -w
+# -*- perl -*-
+
+=head1 NAME
+
+nvme - Munin plugin to monitor the use of NVMe devices
+
+=head1 CONFIGURATION
+
+The plugin uses nvme(1) from the nvme-cli project to read status from
+the NVMe devices.  This requires root access.
+
+  [nvme]
+     user root
+
+The plugin does not support alerting.
+
+=head1 INTERPRETATION
+
+This is a multigraph plugin which makes three graphs
+
+=head2 nvme_usage
+
+This reports how much of capacity is allocated in each NVMe
+"namespace".  The report is in percent.  This number may not have much
+relation to actual use, e.g., if deleted data areas have not been
+trimmed/discarded.
+
+=head2 nvme_bytes
+
+This reports read and write activity on each NVMe device, in bytes per
+second.  Ideally there should be much more read than write.  If they
+are symmetrical, you are using your NVMe as a very expensive FIFO, and
+if you write more than you read, you should probably look for archival
+storage instead.
+
+It is a good idea to compare these numbers to I/O counters from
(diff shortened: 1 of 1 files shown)


plugins/disk/nvme
```diff
@@ -0,0 +1,244 @@
...
+use strict;
+use Munin::Plugin;
+
+# Check that multigraph is supported
+need_multigraph();
+
+sub run_nvme {
+    my (@cmd) = @_;
+    my @lines;
+    $ENV{'LC_ALL'} = 'C';
```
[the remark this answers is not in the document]
kjetilho: yes, you are right, in lib/Munin/Node/Service.pm:prepare_plugin_environment

plugins/disk/nvme
```diff
@@ -0,0 +1,244 @@
...
+    }
+    return \%info;
+}
+
+use Data::Dumper;
+
+my $mode = ($ARGV[0] or "print");
+
+my $list = nvme_list();
+if ($mode eq 'autoconf') {
```
sumpfralle: Here you need to check for the existence of the nvme executable, or?

plugins/disk/nvme
```diff
@@ -0,0 +1,244 @@
+#! /usr/bin/perl -w
+# -*- perl -*-
+
+=head1 NAME
+
+nvme - Munin plugin to monitor the use of NVMe devices
```
sumpfralle: Maybe add a short explanation for `NVMe`?

plugins/disk/nvme
kjetilho: well ..  I think the current failure mode is OK?

```
  # munin-run nvme autoconf
  Can't exec "nvme": No such file or directory at /etc/munin/plugins/disk/nvme line 89.
  no (no devices to monitor)
```
checking lsmod output is a possibility to give a better suggestion to the admin, but that would tie the plugin even closer to Linux.  is autoconf supposed to be silent on stderr?

niclan: Yes, it's supposed to be silent on stderr.

kjetilho:
okay, I've made it more helpful, I hope.  it will still report on stderr in the case where it's not running as root.
```
  # munin-run --servicedir ~ nvme autoconf
  /dev/nvme1n1: Permission denied
  nvme: probably needs to run as user root
  no (no devices to monitor)
```
(the first two lines are on stderr.)  When run as root it will give the appropriate message without noise on stderr.

kjetilho:
btw, IPC::Cmd has been in Perl's corelist since v5.9.5, so should be ok to add without additional requirements.

sumpfralle:
Sorry for letting this rest for so long ...

Thank you!

Issue #914, opened 2018-03-28
 contrib/plugins/disk/du-2 does not work first time until du.cache file is created manually


Talkless:
It seems that I have to create `/var/lib/munin-node/plugin-state/root/du.cache` file manually, or else plugin does not return anything.

strace show that it fails to open that file and exits:
```
open("/var/lib/munin-node/plugin-state/root/du.cache", O_RDONLY) = -1 ENOENT (No such file or directory)
rt_sigaction(SIGHUP, NULL, {SIG_DFL, [], 0}, 8) = 0
rt_sigaction(SIGINT, NULL, {SIG_DFL, [], 0}, 8) = 0
rt_sigaction(SIGQUIT, NULL, {SIG_DFL, [], 0}, 8) = 0
...
exit_group(1)                           = ?
+++ exited with 1 +++
```
I'm guessing it's because of https://github.com/munin-monitoring/contrib/blob/3f253ebec9a3558e41c84ba3c54221d2ad24d6ca/plugins/disk/du-2#L90 and/or https://github.com/munin-monitoring/contrib/blob/3f253ebec9a3558e41c84ba3c54221d2ad24d6ca/plugins/disk/du-2#L109 statements.

Talkless:
Seems that it has another problems, like "No .label provided":
![munin-du-2](https://user-images.githubusercontent.com/292191/38018767-64a5b9f6-3265-11e8-8228-3a1eda0036eb.png)

For this config:
```
[du]
user root
env.dirs /var/lib/postgresql/10/main/pg_wal/ /var/lib/postgresql/9.6/main/pg_xlog/ /var/lib/postgresql/9.6/secondary/pg_xlog/ 
env.interval 5
timeout 300
```

Any ides how to quick-fix it?

Talkless:
I've provided fix for that "No .label provied":
https://github.com/munin-monitoring/contrib/pull/915

sumpfralle:
Thank you for fixing the first half!

The second half looks reasonable, too:
* fetch: the "daemonize" function should be called before exiting in case of a missing cache file
    * we probably cannot simply call it before the "fetch" section, since this would cause a race condition between writing the new data and reading the old data
* config: the "config" section should not use the cache, but just go through the list of environment items instead
    * otherwise the master would receive an empty config and would not bother to ask for the values (which is necessary for triggering the "daemonize" function)

It is a bit mysterious, that all previous users of that plugin seem to have "solved" the problem manually without talking about it :(

PR #949, opened 2018-10-04
Generate cache file if missing

mikini:
Don't exit with error if cache file is missing, instead refrain from sending values and daemonize() in an attempt to generate one.
Invocation using 'config' before generation is successful will fail, as mentioned in #914, but configuration will get picked up as soon as it is present.


sumpfralle:
Looks good to me - thank you!

PR #950, opened 2018-10-07
Add a group of plugins for unifi (Ubiquity) gear

jtsage:
Plugins available:
<strike>
 * unifi_clients_by_device - Count the number of clients connected on a per-device basis

 * unifi_clients_by_network - Count the number of clients connected to each named network

 * unifi_device_stats - Get basic stats from each unifi device - cpu, memory, load average, and uptime

 * unifi_xfer_by_device - Transfer stats on a per-device basis

 * unifi_xfer_by_network - Transfer stats on a per-network basis

 * unifi_xfer_by_port - Transfer stats per switch port

 * unifi_xfer_by_uplink - Overall transfer with speedtest stats
</strike>
 
 * unifi_api - Monolithic version that provides all of the above.

Example graphs included.

Main repo: https://github.com/jtsage/unifi-munin-plugins


plugins/unifi/unifi_clients_by_device
```diff
@@ -0,0 +1,275 @@
...
+# Log back out. (Should invalidate our cookie file, just in case)
+$curl->setopt($curl->CURLOPT_HTTPGET,1);
+$curl->setopt($curl->CURLOPT_WRITEDATA, \$ignore);
+$curl->setopt($curl->CURLOPT_URL, $APILogout);
+$retcode = $curl->perform;
+
+if ( $retcode != 0 ) {
+	die "FATAL:$me: Unable to logout of API: " . $curl->strerror($retcode) . " " . $curl->errbuf . "\n";
+}
+
```
sumpfralle: All of your plugins share the above ~200 lines of code and comments (with slight differences).
It would be an amazing reduction of the maintenance burden for these upcoming plugins, if you would merge them into one single plugin providing many multigraph.

plugins/unifi/unifi_clients_by_device
```diff
@@ -0,0 +1,275 @@
...
+=encoding utf8
+
+=head1 NAME
+
+unifi_clients_by_device - Munin plugin to display clients connected 
+for each ubiquity / unifi device.
+
+=head1 APPLICABLE SYSTEMS
+
+Currently is cabable of polling the unifi line (Ubiquity) from the unifi
```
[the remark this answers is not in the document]
jtsage: sure, makes sense

plugins/unifi/unifi_xfer_by_port
```diff
@@ -0,0 +1,315 @@
...
+			}
+		}
+	}
+}
+
+# Quick 2 digit zero pad
+sub zPad { return sprintf("%02d", $_[0]); }
+
+# Make field names safe.
+sub make_safe { $_ = lc($_[0]); s/^[^a-z_]+/_/; s/\W/_/g; return ( $_ eq "root" ? "__root" : $_ ); }
```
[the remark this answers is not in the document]
jtsage: I hadn't, as I was concerned about module overhead - then I actually read the source of Munin::Plugin, and realized it's negligible.  Will refactor.

plugins/unifi/unifi_clients_by_device
```diff
@@ -0,0 +1,275 @@
...
+		0 # Not Supported By Munin
+	);
+
+# Somewhat (in)sane defaults for host, pass, etc.
+#
+# Note: SSL checking might not work.  My test rig does not.
+my %APIconfig = (
+	"user"     => (( defined $ENV{'user'} ) ? $ENV{'user'} : "ubnt"),
+	"pass"     => (( defined $ENV{'pass'} ) ? $ENV{'pass'} : "ubnt"),
+	"host"     => (( defined $ENV{'host'} ) ? $ENV{'host'} : "https://localhost:8443"),
```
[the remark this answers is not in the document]
jtsage: Yep.

plugins/unifi/unifi_xfer_by_uplink
```diff
@@ -0,0 +1,245 @@
...
+	print "rx_speed.min 0\n";
+	print "tx_speed.min 0\n";
+	print "rx_bytes.min 0\n";
+	print "tx_bytes.min 0\n";
+	print "rx_speed.graph no\n";
+	print "rx_bytes.graph no\n";
+	print "tx_speed.negative rx_speed\n";
+	print "tx_bytes.negative rx_bytes\n";
+	if ( $ENV{'MUNIN_CAP_DIRTYCONFIG'} ) {
+		print "rx_speed.value " . $netList{"rx_speed"} . "\n";
```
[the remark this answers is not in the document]
jtsage: Take a look at unifi_clients_by_device - I pulled the value sections out into subroutines there - is that better?  Or, rather, when using multigraph, can you "redefine" which graph the following is for? (Can I jump in and out of blocks, and just run the regular "fetch" at the end, or can I only have one "multigraph this_specific_graph.sub_graph" appear once?)

plugins/unifi/unifi_clients_by_device
```diff
@@ -0,0 +1,275 @@
...
+This script uses the multigraph functionality to generate many graphs.
+
+For multigraph operation, the "internal" graphs will be one per device
+with seperate metrics for users and guests. The main graph shows total
+clients.
+
+If you are only interested in a single combined graph, you can set:
+
+  [unifi_clients_by_device]
+    env.multigraph no
```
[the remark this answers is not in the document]
jtsage: I'm of two minds about this - the additional graphs don't impact my system heavily - but my site configuration only features 8 devices, so it's not a ton of data, even extrapolated like this - my concern is, on a site with many, many unifi devices, there would be a *huge* amount of subgraphs. Of particular concern is the unifi_xfer_by_port plugin - if I had even a dozen 48 port switches it goes from a handful of graphs to something like 500+.  Perhaps it can be better named - subgraphs_enabled or something?

plugins/unifi/unifi_clients_by_device
```diff
@@ -0,0 +1,275 @@
...
+
+# Somewhat (in)sane defaults for host, pass, etc.
+#
+# Note: SSL checking might not work.  My test rig does not.
+my %APIconfig = (
+	"user"     => (( defined $ENV{'user'} ) ? $ENV{'user'} : "ubnt"),
+	"pass"     => (( defined $ENV{'pass'} ) ? $ENV{'pass'} : "ubnt"),
+	"host"     => (( defined $ENV{'host'} ) ? $ENV{'host'} : "https://localhost:8443"),
+	"site"     => (( defined $ENV{'site'} ) ? $ENV{'site'} : "default"),
+	"sslh"     => (( defined $ENV{'sslh'} ) ? $ENV{'sslh'} : "yes"),
```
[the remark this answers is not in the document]
jtsage: Yes.

plugins/unifi/unifi_clients_by_device
```diff
@@ -0,0 +1,275 @@
...
+	"sslp"     => (( defined $ENV{'sslp'} ) ? $ENV{'sslp'} : "yes"),
```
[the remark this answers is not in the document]
jtsage: Yes - it's ssl_verfiy_peer.

plugins/unifi/unifi_xfer_by_uplink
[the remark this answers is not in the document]
sumpfralle: Using sub routines is usually a good approach. `do_fetch` and `do_config` are common name patterns.
Regarding the mixing of config and value statements combined with `dirtyconfig`: you are free to use multiple identical `multigraph` statements and spread the information among them. They will be merged by the master.
Thus the following pattern is common:
```
do_config
if dirtyconfig: do_fetch
```

plugins/unifi/unifi_clients_by_device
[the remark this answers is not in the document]
sumpfralle: You are right - the bigger environments would probably be happy about that configuration setting.
How about `enable_detail_graphs`? But maybe you can make up a better name.

plugins/unifi/unifi_clients_by_device
sumpfralle: > [..] but munin-graph, assuming it's not running via CGI, would be painful.

Starting with munin 3.0 only CGI mode (as it is called for 2.0) is available. Thus if your concern only relates to CGI/cron mode, then you may feel free to assume CGI mode to be the default.

Regarding performance / resource usage: turning the one-for-all plugin into a wildcard plugin could be the solution for selecting a specific subset of the plugin's features. But I could imagine, that most users will want to use *all* features of the plugin, or? In this case I would suggest to use a single plugin without the wildcard feature.

> [..] what, if any work has been done on explaining best case scenarios where "suggest" can't work before some basic configuration is done? Is this even a possible workflow?

I am not sure, that I understand your question fully. But I will try to answer it :)
There are two types of wildcard plugins:
* A) a plugin that allows the selection of different topics/aspects (e.g. your plugins)
* B) a plugin that can be applied to different targets (e.g. `if_eth0`)

A) would simply return a static list of keywords during `suggest`.
B) would try to discover the list of acceptable targets dynamically.

Does this answer your question?

sumpfralle:
I forgot to mention before, that your plugins look well organized and neat. And you even provided example graphs - this is excellent!

plugins/unifi/unifi_clients_by_device
jtsage: Cool - that makes sense.  I think what I'll do is go monolithic - in the same manner that I did with unifi_device_stats - everything on be default, but allow individual aspects to be shut off.  It's more config, but as you said, I would imagine the vast majority of users would want all of the bits turned on.

And thats good to hear on the CGI front, that does remove my concern about performance totally - actually, the single plugin approach will actually allow me to reduce the number of API calls, which is also not a bad thing at all - right now, it hits the API 3 times per plugin, and the unifi API is not the fastest thing to query over the network - it's not bad, but it definitely had an impact on the runtime of munin-update.  Thanks for the help, expect to see some of these updates added to this pull request sometime soon - it's kinda been nice, as I hadn't touched perl in *years*, working the brain a bit hard.

jtsage:
Just pushed in the monolithic version - by default, everything renders.  Added a slew of enviromental variables, which I thing are well named to turn off individual graph sections, and the detail graphs of those sections that make sense to have them.

Of your above suggestions, I think I adopted most of them, with the exception of pulling in Munin::Plugin - only because I've also forced field names for these to lower case - mostly because it makes sense to me - it's all device names, and doing something like a pair named WLAN-north and WLAN-NORTH seems like it should be discouraged. 

I also ran some metrics, which were interesting.  My setup, with 8 devices, generates ~60k of data and 77 graphs.  The UBNT demo site generates ~430k of data and 530 graphs (and in this case, does it in around 0.15 seconds - time on my site is around 2 secs, all of which is network, I cached the calls from the demo site)

Thanks!

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1138 @@
...
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see L<http://www.gnu.org/licenses/>.
+
+=head1 VERSION
+
+Most recent version available from 
+
+L<https://github.com/jtsage/unifi-munin-plugins>
```
sumpfralle: Do you plan to maintain of the plugin(s) in your repository in the long term?
In this case I would suggest, that we do not try to maintain a code copy here in this munin contrib repository, since this will increase the maintenance burden with no real benefit.

Of course, we are talking about free software - thus there is no way to stop anyone from maintaining code copies.
Please feel free to follow your own judgement when deciding for a primary location of this code. In any case it would probably be misleading to distribute code that refers to an independent location (with content that may/will vary over time).

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1138 @@
...
+
+if (! eval {require JSON; JSON->import(); 1; } ) {
+	push @errorCode, "JSON module not found";
+}
+if (! eval {require WWW::Curl::Easy; 1;} ) {
+	push @errorCode, "WWW::Curl::Easy module not found";
+}
+
+# Provide autoconf info
+if ( defined($ARGV[0]) && $ARGV[0] eq "autoconf" ) {
```
sumpfralle: It looks like this plugin cannot work out of the box (due the unknown URL of the API provider). In this case you should just omit the `autoconf` capability.
Or is the default `api_url` a good guess, that could work sometimes?

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1138 @@
...
+	exit(0);
+}
+
+# Fail on not found dependancies
+if ( @errorCode != 0 ) {
+	die "FATAL:$me: Perl dependancies not installed (", join(", " => @errorCode), ")\n";
+}
+
+# Multigraph cabability is required for this plugin
+if ( ! $ENV{'MUNIN_CAP_MULTIGRAPH'} ) {
```
sumpfralle: By definition the `munin-node` component is responsible for detecting plugins using `multigraph` (i.e. they are not offered to a client that lacks this capability). Thus you can omit this check here.

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1138 @@
...
+
+#######################
+# SUBROUTINES  CONFIG #
+#######################
+
+sub do_config_clients_by_type {
+	# Provide client count by type - CONFIG
+	if ( !$PluginConfig{'enable_clients_type'} ) { return 0; }
+
+	graph_prolouge(
```
sumpfralle: I think, it should be `prologue` instead.

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1138 @@
...
+
+	$APIJsonResponse{'wlan'} = $jsonOBJ->allow_nonref->utf8->relaxed->decode($APIResponse{'wlan'});
+
+	if ( $APIJsonResponse{'wlan'}->{'meta'}->{'rc'} ne 'ok' ) {
+		die "FATAL:$me: Unable get wlan data from API - it said: " , $APIJsonResponse{'wlan'}->{'meta'}->{'msg'} , "\n";
+	}
+}
+
+# Make field names safe, and lowercase. Sourced originally from Munin::Plugin
+sub make_safe { $_ = lc($_[0]); s/^[^a-z_]+/_/; s/\W/_/g; return ( $_ eq "root" ? "__root" : $_ ); }
```
sumpfralle: Did you decide against `clean_fieldname`?

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1138 @@
...
+---------------------------------------
+Devices          |       8  |     126
+AP's             |       4  |     118
+24xSwitch        |       1  |       5
+8xSwitch         |       2  |       2
+Output Bytes     |  64,262  | 431,434
+Output Lines     |   1,761  |  14,586
+Output Graphs    |      77  |     530
+
+So, just note that the growth in the amount of graphed date can be extream.
```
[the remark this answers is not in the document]
jtsage: thank you.  Totally have the tool.  Didn't use it.

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1138 @@
...
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see L<http://www.gnu.org/licenses/>.
+
+=head1 VERSION
+
+Most recent version available from 
+
+L<https://github.com/jtsage/unifi-munin-plugins>
+
+=head1 MAGICK MARKERS
```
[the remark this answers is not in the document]
jtsage: Huh.  I think I copy-pasta'd.  Will fix.

plugins/ubiquiti/unifi_api
jtsage: Yeah, this repo should be it's home - I tend to develop with a github repo, if for no other reason than to remind myself of what I did - these actually started as PHP scripts, with a 3rd party API include - not the best method, but, should I need to use that for something else, it's in the archive.

I will remove the reference.

plugins/ubiquiti/unifi_api
jtsage: I really did it more as a way to check dependancies - then I backported that check to *always* do it, even when config or fetch are run - I'll drop the autoconf bit - the defaults *could* work, but they really, really shouldn't.  URL might be fine - username/password sure shouldn't be.

plugins/ubiquiti/unifi_api
jtsage: Cool, didn't know that - I'll drop this check.

plugins/ubiquiti/unifi_api
jtsage: Hmm.  indeed.

plugins/ubiquiti/unifi_api
jtsage: Yes, and actually, this isn't the finished version the more I think about this - I was normalizing to all lowercase with this - but the unifi controller software has no stipulation that the names I am using for field names are unique.  So, I'm going to dig a little deeper into what the API is spitting out to find out if there is a uniq suffix I can add to them (I don't want to drop the device name, as it makes finding the RRD in question a *lot* easier) - maybe the MAC address as well

plugins/ubiquiti/unifi_api
niclan: @sumpfralle : sorry I didn't catch this before. Please see https://github.com/munin-monitoring/munin/blob/master/plugins/node.d/snmp__if_multi#L140 and https://github.com/munin-monitoring/munin/blob/master/lib/Munin/Plugin.pm#L613.

On the NODE side: If the plugin is run on a node that understands multigraph then MUNIN_CAP_MULTIGRAPH is set and it's safe for the plugin to run. If the node is too old the environment variable is not set and it's not safe to run the plugin.

On the MASTER side the master signals capabilities with "cap multigraph". If it sends this it will see the multigraph plugins.   If it does not it will not see the multigraph plugins. The node knows to censor the list.

But only a recent enough node.  The whole thing breaks down if the node is too old to know to detect multigraph plugins and hide them from inept masters. Therefore the test is right.  But if something more extensive like what is found in Munin::Plugin is used it also handles autoconfig right and gives a helpful (but empty) graph on the server side.

sumpfralle: Good to know! Thank you for correcting me.

plugins/ubiquiti/unifi_api
sumpfralle: Maybe the `need_multigraph` hint from @niclan helps you to embrace the beauty of reusing existing code? :)

Regarding unique names: it is surely a good idea to use prefixes in situations were independent names may try to squat the same namespace.

jtsage: Agreed.  Testing through what I just did, but I did indeed use clean_fieldname - this function is still there, because I've added a way to toggle off what I consider correct behavior and use the "old" naming method - otherwise, I'm going to need to hand rename a couple hundred RRD files.

jtsage:
Still todo for @jtsage : rename of image files.

jtsage:
Probably in a good place now.  I've removed an option only I needed, and actually cleaned up my RRD files in the process.  Also now using the well known and defined graph categories for the generated graphs, as I figured out how to still run this on the node it was already on, but make it appear in the graphs that it's a unique server (I had used a new "unifi" category, as I was concerned with the graphs getting "lost" in my configuration - kept the option of dropping back to the "unifi" category though)

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1166 @@
...
+	}
+
+	$APIJsonResponse{'wlan'} = $jsonOBJ->allow_nonref->utf8->relaxed->decode($APIResponse{'wlan'});
+
+	if ( $APIJsonResponse{'wlan'}->{'meta'}->{'rc'} ne 'ok' ) {
+		die "FATAL:$me: Unable get wlan data from API - it said: " , $APIJsonResponse{'wlan'}->{'meta'}->{'msg'} , "\n";
+	}
+}
+
+# Make field names safe, and lowercase. Sourced originally from Munin::Plugin
```
sumpfralle: The `sourced` comment is outdated, or?

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1166 @@
...
+
+Unifi controllors with direct API access
+
+Controller version 5+ required (tested with 5.8.x)
+
+WebRTC is not supported at this time
+
+=head1 CONFIGURATION
+
+This script uses the multigraph functionality to generate many graphs.
```
sumpfralle: I think, this line is not relevant for `configuration` (or at all).

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1166 @@
...
+=head1 APPLICABLE SYSTEMS
+
+Unifi controllors with direct API access
+
+Controller version 5+ required (tested with 5.8.x)
+
+WebRTC is not supported at this time
+
+=head1 CONFIGURATION
+
```
sumpfralle: Maybe mention the useful `host_name` option for separating the plugin graphs?

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1166 @@
...
+This script uses the multigraph functionality to generate many graphs.
+
+You will need to supply your API login details:
+
+  [unifi_api]
+    env.user Controller_Username
+    # default is ubnt
```
sumpfralle: I think, it is quite common to put comments *above* the line in question (and not below).
Maybe you want to change this?
(just a matter of taste, I guess)

plugins/ubiquiti/unifi_api
```diff
@@ -0,0 +1,1166 @@
...
+    # Show transfer statistics per named network
+    env.enable_detail_xfer_network yes
+    # Show detailed graphs for each named network
+
+    env.enable_xfer_radio yes
+    # Show transfer statistics per radio
+    env.enable_detail_xfer_radio yes
+    # Show detailed graphs for each radio
+
+    env.use_well_know_categories yes
```
sumpfralle: This setting's name is probably not easy to understand for non-developers.
How about reverting the logic and using the setting name `force_category` (default: unset; overrides the `category` filed of all plugins if non-empty). In your case you would use `env.force_category unify`.

jtsage: yes, agreed.  In point of fact, it didn't actually work before - I wasn't reading the variable correctly.  Now called "force_category", as suggested.

plugins/ubiquiti/unifi_api
jtsage: Aye.  also re-worked the section with some sub headings to be a bit easier to read.

jtsage:
I think I'm good to go now - re: your comment of "Feel free to merge all commits into one and force-push you branch. Keeping the commit history is obviously also acceptable."

Admittedly, in all my time on github, this is probably the first I've done a non-trivial pull request from this side (Merged a fair few into my work, but not the other direction).  And I have no idea how to do this - is it the process being talked about here: https://stackoverflow.com/questions/44663822/how-do-i-squash-all-commit-history-and-push-into-another-remote-repository ?

Thanks!

sumpfralle:
Regarding commit squashing: welcome to the happy world of changing history! :)

The approach mentioned by the link you provided feels a bit dirty to me. I would recommend the following instead:
* determine the oldest commit that you want to manipulate (i.e. your first commit in your branch)
* run `git rebase -i FIRST_COMMIT~` (the tilde refers to its parent - this is the non-modified base for your rebase operation)
* the rebase *dialog* is conducted in your text editor (with instructions below): simple change the first word in every line (except for the first one) from `pick` to `squash`
* afterwards you are asked for a commit message

Please note that this operation changes the git history. You should never do this in a public branch that may be used (i.e. relied upon) by other people. But for most projects it is acceptable as part of a merge/pull request.

In case anything goes wrong: remove your local branch and checkout the remote branch from your repository again.

sumpfralle:
Great - thank you for your patience and for adding a really well-shaped munin plugin to the repository!

PR #952, opened 2018-10-16
Debsecan

Neraud:


Neraud:
File names fixed.
I don't think I've seen this limitation explained on the READ.ME or other docs. If it's indeed not there, it might be a good idea to add it.

sumpfralle:
Thank you!

> I don't think I've seen this limitation explained on the READ.ME or other docs.

There is a link in https://github.com/munin-monitoring/contrib/tree/master/plugins#submit-a-new-plugin pointing at http://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-gallery.html#rules-for-plugin-contributors (one hour ago it still pointed to the old wiki, where you would find a deep link into this section of the guide).
But the existence of such a link is obviously no proof, that it is reachable under real conditions :)
If you have a good idea, where a link could be added: please tell me!

PR #953, opened 2018-10-17
Nginx error: log files may be empty

Slamdunk:
@sumpfralle in the commit https://github.com/munin-monitoring/contrib/commit/0e864944d76b19278d82a602aca850c0d91ac51a forgot the fix merged in https://github.com/munin-monitoring/contrib/pull/813 

The log files may be empty, and without this fix this error occurs:
```
Error output from nginx_error_phpMyAdmin:
    /etc/munin/plugins/nginx_error_phpMyAdmin: line 126: line_counts[$status_code]: bad array subscript
```

### DIFF
--- a/plugins/nginx/nginx_error
+++ b/plugins/nginx/nginx_error
@@ -122,9 +122,13 @@ http_codes[503]='Service Unavailable'
 do_fetch () {
   local count status_code
   declare -A line_counts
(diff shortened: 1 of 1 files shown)


sumpfralle:
Ah - indeed - my bad! :(
Thank you for fixing this!

PR #956, opened 2018-10-19
Added loggrepx_ and service_events

kael-shipman:
This adds two plugins, `loggrepx_` and `service_events`.

`loggrepx_` is a riff on munin's core `loggrep` plugin. While `loggrep` aggregates all matches into a single graph line, `loggrepx_` breaks results out into a line for each logfile being searched. Thus, you can track events across multiple logfiles, and define alarms for each, all in one plugin.

`service_events` is similar (and in fact can be considered a further iteration on the `loggrep` family). Its goal (as stated in the documentation) is to aggregate log events across many different logs, grouped by service, rather than logfile. For example, you might have 4 different webservices running in `/srv/www/`, each with it's own nginx log at `/var/log/nginx/[svcname]-errors.log` and a runtime app log at `/srv/www/[svcname]/logs/app.log`. This plugin would allow you to track events in both of those logs (in a way that's distinct for each type of log) for each service. Kind of complicated, but I had a use case for it :).

Anyway, you can take these or leave them. I just figured they might be useful to someone.


plugins/logevents/loggrepx_
```diff
@@ -0,0 +1,162 @@
...
+title="${title:-Log Matches}"
+
+
+
+
+function config() {
+    echo "graph_title ${title}"
+    echo "graph_args --base 1000 -l 0"
+    echo "graph_vlabel ${title}"
+    echo "graph_category logevents"
```
sumpfralle: Please use a category from the [list of well-known categories](http://guide.munin-monitoring.org/en/latest/reference/graph-category.html#well-known-categories). Probably `appserver` or `other` would be suitable.

plugins/logevents/loggrepx_
```diff
@@ -0,0 +1,162 @@
...
+    echo "graph_info Lists number of times the given regex is matched in the given log files per period"
+
+    local nm logfile lbl
+    while read -u 3 -r logfile; do
+        nm="$(echo "$logfile" | sed -r 's/^[^a-zA-Z]+//g' | sed -r 's/[^a-zA-Z0-9]+/_/g')"
+        lbl="$(eval "echo \$${nm}_label")"
```
sumpfralle: See below for an alternative approach to `eval`.

plugins/logevents/loggrepx_
```diff
@@ -0,0 +1,162 @@
+#!/bin/bash
+
```
sumpfralle: How about `set -e` or even `set -eu` in this plugin? (since you used it in the other one)

plugins/logevents/service_events
```diff
@@ -0,0 +1,361 @@
...
+    >&2 echo "E: You must pass either \$*_logbinding (for each service) or \$services_autoconf"
+    exit 1
+fi
+
+
+# Now go find all log files
+LOGFILES=
+declare -a LOGFILEMAP
+while read -u 3 -r v; do
+    if echo "$v" | egrep -q "_logfiles$"; then
```
sumpfralle: Please use `grep -E` instead of `egrep`.

plugins/logevents/service_events
```diff
@@ -0,0 +1,361 @@
...
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
+
+
+# Now get to the real function definitions
+
+function config() {
+    echo "graph_title ${title}"
+    echo "graph_args --base 1000 -l 0"
+    echo "graph_vlabel ${vlabel}"
+    echo "graph_category logevents"
```
sumpfralle: Please use a well-known category.

plugins/logevents/service_events
```diff
@@ -0,0 +1,361 @@
...
+    echo "graph_info Lists number of matching lines found in various logfiles associated with each service"
+
+    while read -u 3 -r svc; do
+        nm="$(echo "$svc" | sed -r 's/^[^a-zA-Z]+//g' | sed -r 's/[^a-zA-Z0-9]+/_/g')"
```
sumpfralle: What is the purpose of this line?

plugins/logevents/service_events
```diff
@@ -0,0 +1,361 @@
...
+    done 3< <(IFS=$'\n'; echo "${services[*]}")
+}
+
+
+
+
+function fetch() {
+    # Load state
+    touch "$MUNIN_STATEFILE"
+    . "$MUNIN_STATEFILE"
```
sumpfralle: Same here as above: please do not blindly execute external code.

plugins/logevents/service_events
```diff
@@ -0,0 +1,361 @@
...
+            >&2 echo "W: No service associated with log $logfile. Skipping...."
+            continue
+        fi
+
+        # Get shell-compatible names for service and logfile
+        svcnm="$(echo "$service" | sed -r 's/^[^a-zA-Z]+//g' | sed -r 's/[^a-zA-Z0-9]+/_/g')"
+        lognm="$(echo "$logfile" | sed -r 's/^[^a-zA-Z]+//g' | sed -r 's/[^a-zA-Z0-9]+/_/g')"
+
+        # Get previous line count to determine whether or not the file may have been rotated
+        prvlines="$(eval "echo \$${lognm}_lines")"
```
sumpfralle: Maybe use `${!varname}` here, too?

plugins/logevents/loggrepx_
```diff
@@ -0,0 +1,162 @@
...
+    done 3< <(echo "$LOGFILES")
+    return 0
+}
+
+
+
+
+
+case "$1" in
+    autoconf) echo no; exit 0 ;;
```
sumpfralle: Since you did not announce the autoconf capability, you do not have to accept this argument.

plugins/logevents/loggrepx_
```diff
@@ -0,0 +1,162 @@
...
+        nm="$(echo "$logfile" | sed -r 's/^[^a-zA-Z]+//g' | sed -r 's/[^a-zA-Z0-9]+/_/g')"
+
+        # Get running number of lines to determine whether or not the file may have been rotated
+        prvlines="$(eval "echo \$${nm}_lines")"
+        if [ -z "$prvlines" ]; then
+            prvlines=0
+        fi
+
+        # Get the current number of lines in the file
+        curlines="$(wc -l "$logfile" | egrep -o '^[0-9]+')"
```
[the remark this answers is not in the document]
kael-shipman: Veeeeeeeeerry tricky ;).

plugins/logevents/loggrepx_
```diff
@@ -0,0 +1,162 @@
...
+        if ! [ "$curlines" -eq "$curlines" ] &>/dev/null; then
```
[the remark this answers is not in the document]
kael-shipman: It checks to see if it's an integer. If for whatever reason the wc command messed up, then the variable could be blank (or other, though in this case the regex prevents anything other than blank or int), so this just sets it to 0 if it doesn't look valid. Maybe not enough of a risk to justify the three extra lines here, but I tend to be extra defensive in bash ;).

kael-shipman:
Didn't want to lose all the comments and changes, so I didn't move the files into the `logs` directory as suggested just yet. Will make that the final step after approval of changes.

plugins/logevents/loggrepx_
```diff
@@ -0,0 +1,169 @@
...
+    echo "graph_args --base 1000 -l 0"
+    echo "graph_vlabel ${title}"
+    echo "graph_category other"
+    echo "graph_info Lists number of times the given regex is matched in the given log files per period"
+
+    local var_prefix var logfile lbl
+    while read -u 3 -r logfile; do
+        var_prefix="$(echo "$logfile" | sed -r 's/^[^a-zA-Z]+//g' | sed -r 's/[^a-zA-Z0-9]+/_/g')"
+        var="${var_prefix}_label"
+        lbl="${!var:-0}"
```
sumpfralle: Here you probably wanted to do something like the following:

```suggestion
        lbl="${!var:-$logfile}"
```
(this simplifies the line below)

sumpfralle:
Great - thank you for contributing such useful plugins!

kael-shipman:
Oh, one more thing. What's standard procedure here for adding more documentation and screenshots for these plugins? Would hate to make people try to imagine what these do when it would be so easy just to show them.

sumpfralle:
Documentation: the header of each plugin is the best place (to be viewed via `munin-doc PLUGIN`).

Example graphs are very welcome! Please take a look at http://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-gallery.html#rules-for-plugin-contributors. Thank you!

PR #959, opened 2018-11-18
Added extinfo to list affected logs and improved documentation

kael-shipman:
* Improved documentation
* Added example graphs for new plugins
* Added list of affected logs as `extinfo` for `service_events` plugin (this gives visibility into which logs are producing events for a given service)


plugins/logs/service_events
```diff
@@ -330,10 +368,17 @@ function fetch() {
...
 
-        # Aggregate and add to the correct service counter
-        svc_counter="${svcnm}_total"
-        !((matches+=${!svc_counter}))
-        typeset "$svc_counter=$matches"
+        # If there were matches, aggregate them and add this log to the extinfo for the service
+        if [ "$matches" -gt 0 ]; then
+            # Aggregate and add to the correct service counter
+            svc_counter="${svcnm}_total"
+            !((matches+=${!svc_counter}))
```
sumpfralle: What is the purpose of the first exclamation mark here?
Maybe it would increase readability, if you would split this into two lines?
```shell
svc_counter_value=${!svc_counter}
matcher=$((matcher + svc_counter_value))
```

plugins/logs/loggrepx_
```diff
@@ -40,12 +41,13 @@ Available config options include the following:
  env.[field]_critical - Critical level for specific logfile
 
 NOTE: for any variable with [field] in it, [field] is derived from the
-full logfile path by simply replacing all non-alphanumerics with
-underscores. For example, the "warning" field for the logfile
-\`/var/log/nginx/errors.log\` would be \`var_log_nginx_errors_log_warning\`
+full logfile path by simply removing the preceding slash and replacing
+all non-alphanumerics with underscores. For example, the "warning" field
+for the logfile \`/var/log/nginx/errors.log\` would be
```
sumpfralle: Could you verify please, that this looks reasonable in `perldoc FILENAME`? (see https://perldoc.perl.org/perlpod.html#Formatting-Codes)

plugins/logs/service_events
```diff
@@ -306,17 +348,13 @@ function fetch() {
...
-        fi
+        prvlines="${prvlines:-0}"
 
-        # Get the current number of lines in the file
+        # Get the current number of lines in the file (defaulting to 0 on error)
         curlines="$(wc -l < "$logfile")"
-        if ! [ "$curlines" -eq "$curlines" ] &>/dev/null; then
-            curlines=0
-        fi
+        curlines="${curlines:-0}"
```
sumpfralle: Is this conditional supposed to work around a missing `$logfile`?
Wouldn't the above assignment error out anyway in this case?

plugins/logs/service_events
```diff
@@ -330,10 +368,17 @@ function fetch() {
...
         matches="$(tail -n +"$prvlines" "$logfile" | grep -Ec "${!logmatch}" || true)"
 
-        # Aggregate and add to the correct service counter
-        svc_counter="${svcnm}_total"
-        !((matches+=${!svc_counter}))
-        typeset "$svc_counter=$matches"
+        # If there were matches, aggregate them and add this log to the extinfo for the service
+        if [ "$matches" -gt 0 ]; then
+            # Aggregate and add to the correct service counter
+            svc_counter="${svcnm}_total"
```
sumpfralle: Maybe clarify withe name of the variable, that it is a *name* instead of a *value*?

plugins/logs/service_events
kael-shipman: Probably a good idea. The exclamation point is a weird hack that's necessary when the result of the expression equals 1. For whatever reason, in that case (and only that case), the expression returns an exit value of `1`, which, when `set -e` is used, kills the script without explanation. Negating the expression causes bash not to kill script execution, whatever the return value of the expression. Oh, bash....

But you're right. I'll break this out so it doesn't rely on idiomatic hacks.

plugins/logs/service_events
kael-shipman: Ah yes, it would err out, good call. Not sure what the best approach is in this case. Letting it err out and kill the script seems like the best answer, since the output to stderr should be sufficient for debugging purposes.

plugins/logs/service_events
kael-shipman: +1

plugins/logs/service_events
kael-shipman: Or better yet, put this conditional at the top of the loop, where it makes more sense.

sumpfralle: Thinking about it further: munin plugins are supposed to be fun to write - thus feel free to take a shortcut here, that *usually* works :)

plugins/logs/service_events
```diff
@@ -208,8 +209,8 @@ done 3< <(echo "$vars")
 n=0
 allvars=0
 while [ $n -lt "${#reqvars[@]}" ]; do
-    !((allvars+=$(( 2 ** $n ))))
-    !((n++))
+    allvars=$(( allvars + $(( 2 ** $n )) ))
```
sumpfralle: I think, you can cut this down to `allvars=$((allvars + 2 ** n))`?

kael-shipman: Haha, for stuff like this, feel free to just merge it and then fix to your liking. I'm attached to the functionality, not the implementation, so don't worry about offending me by making unsolicited changes. But I think if we go back and forth like this we'll end up with 100 commits just to add 20 lines of code, and we'll both have lost our Sunday ;).

kael-shipman: (That's not to say I'm not learning from this, because I am, and I do value that, even in this relatively trivial example. But I can also learn from what you change post-merge, especially if you comment on why you changed it then.)

sumpfralle: Good point :)

Thus I may conclude: I feel fine the current set of changes, that you are proposing. Thus if you feel fine, too, I will merge it. Have a good remaining Sunday!

kael-shipman: Great, sounds good! And thanks again for working with me. I've already learned 5 concrete new bash/gnutils things just from these two PRs :D.

sumpfralle:

PR #960, opened 2018-11-28
Create plugin timesync_status to monitor systemd ntp.

bertptrs:
This patch adds a plugin that monitors the usual NTP stats for `systemd-timesyncd`. It can do autoconf and requires no configuration on regular systems.

It depends only on a reasonably modern version of Python (3.5+) which has been shipped in the latest (and sometimes previous) LTS releases of most popular distros, save for CentOS.


sumpfralle:
This looks good - thank you!

PR #961, opened 2018-11-28
Add support for configurable thresholds to nullmailer_queue plugin

bobkare:
This commit enables thresholds to be set for the total size of the nullmailer queue.

It uses the print_warning and print_critical functions from plugin.sh, so configuration should be pretty standard.

### DIFF
--- a/plugins/mail/nullmailer_queue
+++ b/plugins/mail/nullmailer_queue
@@ -54,7 +54,7 @@ GPLv2
 
 =cut
 
-. $MUNIN_LIBDIR/plugins/plugin.sh
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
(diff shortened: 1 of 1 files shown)


plugins/mail/nullmailer_queue
```diff
@@ -50,9 +54,13 @@ GPLv2
 
 =cut
 
+. $MUNIN_LIBDIR/plugins/plugin.sh
```
sumpfralle: Please quote the path in order to avoid whitespace issues.

bobkare:
I quoted that path and also ran shellcheck on it to see if it identified any other issues (it didn't).

sumpfralle:
Great - thank you!

PR #962, opened 2018-11-30
Plugins for certificates

ap-wtioit:
* plugin to monitor certificiate lifetime
* plugin to monitor letsencrypt certificate issue limit

### DIFF
--- a/plugins/ssl/certificate_file_expiry
+++ b/plugins/ssl/certificate_file_expiry
@@ -1,0 +1,72 @@
+#!/bin/sh
+: << =cut
+=head1 NAME
+
+certficate_file_expiry - check the certificate validity of your certfificates
+
+= head1 CONFIGURATION
+
+Installing: Add list of your certificates prefixed by the type in munin plugin-conf.d
+
+For openvpn ca.crt and crl.pem
+[certficate_file_expiry]
+user root
+env.CERTS crl:/etc/openvpn/easy-rsa/keys/crl.pem x509:/etc/openvpn/easy-rsa/keys/ca.crt
+
+For letsencrypt certficates
+[certficate_file_expiry]
+user root
+env.CERTS x509:/etc/letsencrypt/live/domain1.example.com/cert.pem x509:/etc/letsencrypt/live/domain2.example.com/cert.pem
+
+Warning and Critical levels can also be configured with env variables like this
+[certficate_file_expiry]
+...
+# warn when certificate will be invalid within 5 days
+env.warning 5:
+# critical when certificate will be invalid within 1 day
+env.critical 1:
+
+=head1 Dependencies
+
+Dependencies: openssl
+
+=head1 AUTHOR
+
+andreas perhab - andreas.perhab@wt-io-it.at
+https://www.wt-io-it.at/
+
+=head1 LICENSE
+
+GPLv2
+
+=cut
+
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
+
+if [ "$1" = "config" ] ; then
+	echo "graph_title Certificate validity"
+	echo "graph_args --logarithmic --base 1000"
+	echo "graph_vlabel certificate validity in days"
+	echo "graph_category security"
+fi
+
+now=$(date +%s)
+warning=${warning:-5:}
+critical=${critical:-1:}
+for cert in ${CERTS}; do
+	cert_type=${cert%:*}
+	cert_file=${cert#*:}
+	cert_name=$(clean_fieldname "$cert_file")
+	if [ "$1" = "config" ] ; then
+		echo "${cert_name}.label ${cert_file}"
+		print_warning "$cert_name"
+		print_critical "$cert_name"
+	elif [ "$1" = "" ] ; then
+		validity=$(/usr/bin/openssl "$cert_type" -text -noout -in "$cert_file" | grep -E '(Next Update|Not After)')
+		validity=${validity#*:}
+		validity=$(date --date="$validity" +%s)
+		validity=$((validity - now))
+		validity=$(echo "$validity" | awk '{ print ($1 / 86400) }')
+		echo "${cert_name}.value $validity"
+	fi
+done
--- a/plugins/ssl/letsencrypt_weekly
+++ b/plugins/ssl/letsencrypt_weekly
@@ -1,0 +1,62 @@
+#!/bin/sh
+: << =cut
+=head1 NAME
+
+letsencrypt_weekly - monitor the number of CSRs by week for /etc/letsencrypt/csr/
+
+see https://letsencrypt.org/docs/rate-limits/
+
+= head1 CONFIGURATION
+
+You can configure the warning and critical limits for this plugin:
+
+[letsencrypt_weekly]
+# warn when more than 40 certificates have been requested in the last week
+env.warning :40
+# critical when more than 50 certificates have been requested in the last week
+env.critical :50
+
+=head1 Dependencies
+
+Dependencies: openssl
+
+=head1 AUTHOR
+
+andreas perhab - andreas.perhab@wt-io-it.at
+https://www.wt-io-it.at/
+
+=head1 LICENSE
+
+GPLv2
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
+
+warning=${warning:-:40}
+critical=${critical:-:50} #letsencrypt doesn't allow more than 50 certificates per week
+# see https://letsencrypt.org/docs/rate-limits/
+
+if [ "$1" = "autoconf" ] ; then
+	test -d /etc/letsencrypt/csr/ && echo "yes" || echo "no (directory /etc/letsencrypt/csr does not exist)"
+elif [ "$1" = "config" ] ; then
+	echo "graph_title Letsencrypt certificate requests during last week"
+	echo "graph_args --base 1000"
+	echo "graph_vlabel Number of certificates"
+	echo "graph_category security"
+	echo "letsencrypt_weekly.label Letsencrypt certificates last week"
+	print_warning "letsencrypt_weekly"
+	print_critical "letsencrypt_weekly"
+elif [ "$1" = "" ] ; then
+	if existing_certs=$(find /etc/letsencrypt/csr/ -mtime -7 -type f 2>/dev/null); then
+		value=$(echo "$existing_certs" | wc -l)
+	else
+		value="U"
+	fi
+	echo "letsencrypt_weekly.value $value"
+fi


shtrom:
There are already two similar plugins in the `ssl` directory.

ap-wtioit:
@shtrom both of them need an open SSL port. In case of the CRL for OpenVPN there is no port where you can get the certificate from, that's why i wrote the plugin that can do it by reading the certificate / CRL from the filesystem. 

Should i move my plugins into the ssl directory and or rewrite the original plugins so they can handle the OpenVPN CRL as well?

shtrom:
Ah, yep, fair enough.

I think it would make sense to move them all in the same directory, yes.

ap-wtioit:
I've updated the pull request and moved the plugins into the ssl directory.

plugins/ssl/cert_validity
```diff
@@ -0,0 +1,65 @@
...
+for cert in ${CERTS[@]}; do
+	cert_type=${cert%:*}
+	cert_file=${cert#*:}
+	cert_name=${cert_file//\//_}
+	cert_name=${cert_name//./_}
+	cert_name=${cert_name//-/_}
+	cert_name=${cert_name#_}
+	if [ "$1" == "config" ] ; then
+		echo "${cert_name}.label ${cert_file}"
+		echo "${cert_name}.warning 5:"
```
sumpfralle: How about the following instead (for configurable thresholds)?
```shell
warning=${warning:-5:} print_warning "$cert_fieldname"
```
(from `clean_fieldname` above)

Same for *critical* ...

plugins/ssl/cert_validity
```diff
@@ -0,0 +1,65 @@
...
+		echo "${cert_name}.critical 1:"
+	elif [ "$1" == "" ] ; then
+		validity=$(/usr/bin/openssl $cert_type -text -noout -in "$cert_file" | grep -E '(Next Update|Not After)')
+		validity=${validity#*:}
+		validity=$(date --date="$validity" +%s)
+		validity=`expr $validity - $now`
```
sumpfralle: How about `validity=$((validity - now))`?

plugins/ssl/cert_validity
```diff
@@ -0,0 +1,65 @@
...
+		if which bc > /dev/null; then
```
sumpfralle: Maybe just use `awk` for float calculations: `echo "$validity" | awk '{ print($1 / 86400) }'`?

(munin plugins may assume, that `awk` is available)

plugins/ssl/cert_validity
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+#
+# check the certificate validity of your certfificates
```
sumpfralle: Please use a perldoc header here.

The [ps_ plugin](https://github.com/munin-monitoring/munin/blob/master/plugins/node.d/ps_) could be a good example.

plugins/ssl/cert_validity
```diff
@@ -0,0 +1,65 @@
...
+#
+# andreas perhab - andreas.perhab@wt-io-it.at
+# https://www.wt-io-it.at/
+#
+# License: GPLv2
+#
+# Installing: Add list of your certificates prefixed by the type in munin plugin-conf.d
+#
+# For openvpn ca.crt and crl.pem
+# [cert_validity]
```
sumpfralle: How about using a name, that indicates the *local* access to certificates (e.g. `certificate_file_expiry`)?

plugins/ssl/cert_validity
```diff
@@ -0,0 +1,65 @@
...
+	echo "graph_args --logarithmic --base 1000"
+	echo "graph_vlabel certificate validity in days"
+	echo "graph_category certificates"
+fi
+
+now=$(date +%s)
+for cert in ${CERTS[@]}; do
+	cert_type=${cert%:*}
+	cert_file=${cert#*:}
+	cert_name=${cert_file//\//_}
```
sumpfralle: Please use `clean_fieldname "$cert_file"` instead.

(you need to add `. "$MUNIN_LIBDIR/plugins/plugin.sh"` before)

plugins/ssl/cert_validity
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
sumpfralle: It would be great, if you could switch to `sh`. This would allow the plugin to be used with OpenWrt or other embedded environments.

I commented the `bash`-only features below.

plugins/ssl/cert_validity
```diff
@@ -0,0 +1,65 @@
...
+
+if [ "$1" == "config" ] ; then
+	echo "graph_title Certificate validity"
+	echo "graph_args --logarithmic --base 1000"
+	echo "graph_vlabel certificate validity in days"
+	echo "graph_category certificates"
+fi
+
+now=$(date +%s)
+for cert in ${CERTS[@]}; do
```
sumpfralle: In case of `sh` (instead of `bash`), `$CERTS` should provide the same result.

plugins/ssl/cert_validity
```diff
@@ -0,0 +1,65 @@
...
+	cert_name=${cert_file//\//_}
+	cert_name=${cert_name//./_}
+	cert_name=${cert_name//-/_}
+	cert_name=${cert_name#_}
+	if [ "$1" == "config" ] ; then
+		echo "${cert_name}.label ${cert_file}"
+		echo "${cert_name}.warning 5:"
+		echo "${cert_name}.critical 1:"
+	elif [ "$1" == "" ] ; then
+		validity=$(/usr/bin/openssl $cert_type -text -noout -in "$cert_file" | grep -E '(Next Update|Not After)')
```
sumpfralle: Please quote `$cert_type` here. `shellcheck` is your friend :)
(here: `shellcheck -s dash`)

plugins/ssl/cert_validity
```diff
@@ -0,0 +1,65 @@
...
+# env.CERTS x509:/etc/letsencrypt/live/domain1.example.com/cert.pem x509:/etc/letsencrypt/live/domain2.example.com/cert.pem
+#
+# Dependencies: openssl
+#
+# Parameters:
+#
+#  config
+
+
+if [ "$1" == "config" ] ; then
```
sumpfralle: In case of switching to `sh`: use `=` instead of `==`.

plugins/ssl/letsencrypt_weekly
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
sumpfralle: Please use `sh`, if possible.

plugins/ssl/letsencrypt_weekly
```diff
@@ -0,0 +1,31 @@
...
+#
+# monitor the number of CSRs by week for /etc/letsencrypt/csr/
```
sumpfralle: Please use a `perldoc` header.

Regarding the name of the plugin: maybe clarify the purpose of the plugin (e.g. *letsencrypt_certificate_request_count*)?

(maybe you want to make the *7 days* configurable later, thus I would not add the period to the plugin name)

plugins/ssl/letsencrypt_weekly
```diff
@@ -0,0 +1,31 @@
...
+# Parameters:
+#
+#  config
+#  autoconfig
+#
+#%# family=auto
+#%# capabilities=autoconf
+
+if [ "$1" == "autoconf" ] ; then
+    find /etc/letsencrypt/csr/ > /dev/null 2>&1 && echo "yes" || echo "no"
```
sumpfralle: Please add a reason to `no` (e.g. `no (no files below /etc/letsncrypt/csr)`).

Additionally you should not let the error message of `find /etc/letsencrypt/csr/` (in case of non-existing directory) slip through to the user. Instead you could simply check, whether the directory exists.

plugins/ssl/letsencrypt_weekly
```diff
@@ -0,0 +1,31 @@
...
+elif [ "$1" == "config" ] ; then
+	echo "graph_title Letsencrypt certificates last week"
+	echo "graph_args --base 1000"
+	echo "graph_vlabel Number of certificates"
+	echo "graph_category certificates"
+	echo "letsencrypt_weekly.label Letsencrypt certificates last week"
+	echo "letsencrypt_weekly.warning :40"
```
sumpfralle: How about `warning=${warning:-:40} print_warning "letsencrypt_weekly"`?

plugins/ssl/letsencrypt_weekly
```diff
@@ -0,0 +1,31 @@
...
+#%# family=auto
+#%# capabilities=autoconf
+
+if [ "$1" == "autoconf" ] ; then
+    find /etc/letsencrypt/csr/ > /dev/null 2>&1 && echo "yes" || echo "no"
+elif [ "$1" == "config" ] ; then
+	echo "graph_title Letsencrypt certificates last week"
+	echo "graph_args --base 1000"
+	echo "graph_vlabel Number of certificates"
+	echo "graph_category certificates"
```
sumpfralle: Please use one of the *well known* category names (see http://guide.munin-monitoring.org/en/latest/reference/graph-category.html#well-known-categories). Here I would suggest `security`.

plugins/ssl/letsencrypt_weekly
```diff
@@ -0,0 +1,31 @@
...
+#  config
+#  autoconfig
+#
+#%# family=auto
+#%# capabilities=autoconf
+
+if [ "$1" == "autoconf" ] ; then
+    find /etc/letsencrypt/csr/ > /dev/null 2>&1 && echo "yes" || echo "no"
+elif [ "$1" == "config" ] ; then
+	echo "graph_title Letsencrypt certificates last week"
```
sumpfralle: How about *Letsencrypt certificate requests during last week*.

plugins/ssl/letsencrypt_weekly
```diff
@@ -0,0 +1,31 @@
...
+	echo "graph_args --base 1000"
+	echo "graph_vlabel Number of certificates"
+	echo "graph_category certificates"
+	echo "letsencrypt_weekly.label Letsencrypt certificates last week"
+	echo "letsencrypt_weekly.warning :40"
+	echo "letsencrypt_weekly.critical :50" #letsencrypt doesn't allow more than 50 certificates per week
+	# see https://letsencrypt.org/docs/rate-limits/
+elif [ "$1" == "" ] ; then
+	echo "letsencrypt_weekly.value $(find /etc/letsencrypt/csr/ -mtime -7 -type f | wc -l)"
```
sumpfralle: If you want to be perfect, you should return `U`, if the directory does not exist.

plugins/ssl/cert_validity
ap-wtioit: i fixed that

plugins/ssl/cert_validity
ap-wtioit: also fixed

plugins/ssl/cert_validity
ap-wtioit: thanks for the suggestion, now the calculation is done with awk

plugins/ssl/cert_validity
ap-wtioit: changed to perldoc

plugins/ssl/cert_validity
ap-wtioit: renamed plugin to certificate_file_expiry

plugins/ssl/cert_validity
ap-wtioit: changed to `clean_fieldname $cert_file`

plugins/ssl/cert_validity
ap-wtioit: applied suggestions for sh instead of bash (could not test sh compatiblity on our servers because /bin/sh is symlinked in debian and ubuntu)

plugins/ssl/cert_validity
ap-wtioit: thanks, is quoted now

plugins/ssl/cert_validity
ap-wtioit: switched to sh and change the `==` to `=`

plugins/ssl/letsencrypt_weekly
ap-wtioit: done

plugins/ssl/letsencrypt_weekly
ap-wtioit: changed to perldoc

did not change the plugin name in this case (the weekly limit is named like this (weekly / per week) in the letsencrypt documentation)

plugins/ssl/letsencrypt_weekly
ap-wtioit: Improved the `no`

The additional error message is suppressed with `> /dev/null 2>&1` already (Did test that no error message is given to the user when the directory does not exist, only the `no ...` message is printed) .

plugins/ssl/letsencrypt_weekly
ap-wtioit: added it here as well

plugins/ssl/letsencrypt_weekly
ap-wtioit: changed category to `security` (even if that feels a bit awkward to me (the plugin does nothing for security) )

plugins/ssl/letsencrypt_weekly
ap-wtioit: Thanks, changed like suggested

plugins/ssl/letsencrypt_weekly
ap-wtioit: added the `U` in case of an error

ap-wtioit:
@sumpfralle thanks for the review, i incorporated most of your suggestions

plugins/ssl/certificate_file_expiry
```diff
@@ -0,0 +1,72 @@
...
+=head1 NAME
+
+certficate_file_expiry - check the certificate validity of your certfificates
+
+= head1 CONFIGURATION
+
+Installing: Add list of your certificates prefixed by the type in munin plugin-conf.d
+
+For openvpn ca.crt and crl.pem
+[cert_validity]
```
sumpfralle: A remainder of the old name is hiding here :)

plugins/ssl/certificate_file_expiry
```diff
@@ -0,0 +1,72 @@
...
+	echo "graph_category security"
+fi
+
+now=$(date +%s)
+warning=${warning:-5:}
+critical=${critical:-1:}
+for cert in ${CERTS}; do
+	cert_type=${cert%:*}
+	cert_file=${cert#*:}
+	cert_name=$(clean_fieldname $cert_file)
```
sumpfralle: For perfection: please quote the variable.

`shellcheck -s dash` is your friend :)

plugins/ssl/certificate_file_expiry
```diff
@@ -0,0 +1,72 @@
...
+	if [ "$1" = "config" ] ; then
+		echo "${cert_name}.label ${cert_file}"
+		print_warning "$cert_name"
+		print_critical "$cert_name"
+	elif [ "$1" = "" ] ; then
+		validity=$(/usr/bin/openssl "$cert_type" -text -noout -in "$cert_file" | grep -E '(Next Update|Not After)')
+		validity=${validity#*:}
+		validity=$(date --date="$validity" +%s)
+		validity=$((validity - now))
+		validity=$(echo $validity | awk '{ print ($1 / 86400) }')
```
sumpfralle: Please quote `validity` here, too.

plugins/ssl/letsencrypt_weekly
```diff
@@ -0,0 +1,50 @@
...
+if [ "$1" = "autoconf" ] ; then
+    find /etc/letsencrypt/csr/ > /dev/null 2>&1 && echo "yes" || echo "no (no files below /etc/letsencrypt/csr)"
+elif [ "$1" = "config" ] ; then
+	echo "graph_title Letsencrypt certificate requests during last week"
+	echo "graph_args --base 1000"
+	echo "graph_vlabel Number of certificates"
+	echo "graph_category security"
+	echo "letsencrypt_weekly.label Letsencrypt certificates last week"
+	print_warning "letsencrypt_weekly"
+	echo "letsencrypt_weekly.critical :50" #letsencrypt doesn't allow more than 50 certificates per week
```
sumpfralle: Regarding the hard limit of `50`: I would suggest, to make this configurable, too. `letsencrypt` will surely change the rules from time to time. It would be nice, if people could adjust to such changes via configuration and not via code.

plugins/ssl/letsencrypt_weekly
```diff
@@ -0,0 +1,50 @@
...
+	# see https://letsencrypt.org/docs/rate-limits/
+elif [ "$1" = "" ] ; then
+	echo "letsencrypt_weekly.value $(find /etc/letsencrypt/csr/ -mtime -7 -type f | wc -l || echo 'U')"
```
sumpfralle: Here `||` would never match, since the exit code of the preceding pipe is determined by `wc` (not `find`). Thus it will always be true.
Maybe use something like the following?
```shell
if existing_certs=$(find /etc/letsencrypt/csr/ -mtime -7 -type f >/dev/null); then
    value=$(echo "$existing_certs" | wc -l)
else
    value="U"
fi
```

```

File: plugins/ssl/letsencrypt_weekly [hunk 23]
```diff
@@ -0,0 +1,50 @@
...
+elif [ "$1" = "config" ] ; then
+	echo "graph_title Letsencrypt certificate requests during last week"
+	echo "graph_args --base 1000"
+	echo "graph_vlabel Number of certificates"
+	echo "graph_category security"
+	echo "letsencrypt_weekly.label Letsencrypt certificates last week"
+	print_warning "letsencrypt_weekly"
+	echo "letsencrypt_weekly.critical :50" #letsencrypt doesn't allow more than 50 certificates per week
+	# see https://letsencrypt.org/docs/rate-limits/
+elif [ "$1" = "" ] ; then
```
sumpfralle: It is quite common to use `[ -z "$1" ]` instead, but this is a matter of taste ...

plugins/ssl/letsencrypt_weekly
```diff
@@ -0,0 +1,50 @@
...
+ #%# capabilities=autoconf
+
+=cut
+
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
+
+warning=${warning:-:40}
+
+if [ "$1" = "autoconf" ] ; then
+    find /etc/letsencrypt/csr/ > /dev/null 2>&1 && echo "yes" || echo "no (no files below /etc/letsencrypt/csr)"
```
sumpfralle: At the moment the reason does not really match the test. Maybe use the following?
```shell
[ -d /etc/letsencrypt/csr/ ] && echo "yes" || echo "no (directory /etc/letsencrypt/csr does not exist)"
```

plugins/ssl/cert_validity
sumpfralle: This should be fine. At least my understanding of `/bin/sh` is, that it is just *the default shell of the system*. Quite recently Debian (and thus Ubuntu) switched from `/bin/sh` being provided by `/bin/bash` to `/bin/dash`. Thus there is no single one implementation. `shellcheck -s dash` is the base reference, we usually use for munin plugins in this repository.

plugins/ssl/certificate_file_expiry
ap-wtioit: thanks for noticing, fixed

plugins/ssl/certificate_file_expiry
ap-wtioit: done

plugins/ssl/certificate_file_expiry
ap-wtioit: done

plugins/ssl/letsencrypt_weekly
ap-wtioit: i added a variable (and config description in the doc)

plugins/ssl/letsencrypt_weekly
ap-wtioit: yes, didn't notice that, thanks

plugins/ssl/letsencrypt_weekly
ap-wtioit: changed to a test of existing directory (better than checking if there is already a CSR)

sumpfralle:
Great - thank you for bearing with my wishes!

PR #963, opened 2018-12-13
Fix jstat* plugins to work again with java 6+7, fix PU reporting for Java8

shuehner:
jstat column output is not stable across JVM versions (technically
version of jstat binary used).
New columns are added not add the end of column list but in the middle
breaking access via constant index.

Following table shows the known columns per version:
Oracle JDK / OpenJDK 1.5 .. 1.7
   S0C    S1C    S0U    S1U      EC       EU        OC         OU       PC     PU    YGC    YGCT   FGC     FGCT    GCT
Openjdk 1.8 .. 10
   S0C    S1C    S0U    S1U      EC       EU        OC         OU       MC     MU    CCSC   CCSU   YGC     YGCT    FGC    FGCT     GCT
Openjdk 11  .. 12
   S0C    S1C    S0U    S1U      EC       EU        OC         OU       MC     MU    CCSC   CCSU   YGC     YGCT    FGC    FGCT     CGC    CGCT     GCT

Earlier commit here added support for java8+ but using version checking
if version == 1.5 > Use pre-java8 column format, else new format.

As fixing java6+7 would mean more version comparison which is ugly this
cmomit changes retrival logic to the following;
a.) Parse first line jstat output to find position for each label
b.) Then use that position to fetch value from 2nd line

That way code auto-adjusts to any ordering change without needing any
java version specific code or checks.

On the way fix 'Permanent Used' value reporting for java8 which was
broken (missing variable rename from PU -> MU when java8 support was
added).

Note:
For gccount + gctime plugins the search by label is probably overkill as just 2 values
But for heap plugin new code is even shorter.
Main reason for changing the logic is to avoid version parsing/comparison.

Tested here locally on:
- Openjdk 7, 8, 10, 11 (jstat from those)
- Partial Test done on Oracle Java SE 1.5 but please recheck

### DIFF
--- a/plugins/jvm/jstat__gccount
+++ b/plugins/jvm/jstat__gccount
@@ -25,6 +25,7 @@
 #       Target Java Virtual Machine to monitor are:
 #         Sun JDK 5.0 (http://java.sun.com/javase/)
 #         Sun JDK 8.0 (http://java.sun.com/javase/)
+#         OpenJDK 1.7 .. 11 (https://openjdk.java.net/)
 #         BEA JRockit 5.0 (http://dev2dev.bea.com/jrockit/)
 #
 # Parameters:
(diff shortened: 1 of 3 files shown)


sumpfralle:
What a pity that these plugins contain so much code duplication :(

Your changes look good to me. Thank you!

PR #966, opened 2019-01-08
Fix spaces in directory names. Glob directories.

MyCryptoCheckout:
Directory names were not being quotes for du.

Also added automatic globbing of directories to enable stuff like /home/*


sumpfralle:
Thank you!

PR #968, opened 2019-01-16
Ubiquity Plugin: login error fix

jtsage:
new controllers require the use of double quoted string to login, thanks to @fongd for catching this.


sumpfralle:
Thank you!

Issue #969, opened 2019-02-03
zfs/zfs_list plugin not working with dash in device name


Sephtex:
When a device name contains a dash (minus sign) in it's name, the graph aren't generated.
Eg: zfspool/device-name
This can be fixed by adding an extra sed with a replace of - to _ for the fsname.

I'll try to create a PR for this

sumpfralle:
Merged in #970 - thank you!

PR #970, opened 2019-02-03
Add support for dash in zfs_list

Sephtex:
#969 zfs/zfs_list: Add support for dash in zfs_list

### DIFF
--- a/plugins/zfs/zfs_list
+++ b/plugins/zfs/zfs_list
@@ -79,7 +79,7 @@ for i in `zfs list -Hp | awk '{print $1}'`; do
 	values=( $(zfs get -p usedbydataset,usedbychildren,usedbysnapshots,usedbyrefreservation,available,quota $i | awk 'BEGIN {total=0;} { if( NR==1 ) next; } !/quota/ {total=total+$3;} {print $3} END{print total;}') )
 	fsname=$(clean_fieldname $(echo "$i" | sed 's/\//__/g'))
 
-echo <<EOF "multigraph zfs_list_$fsname
+	echo <<EOF "multigraph zfs_list_$fsname
 usedbydataset.value ${values[0]}
 usedbysnapshots.value ${values[2]}
 usedbychildren.value ${values[1]}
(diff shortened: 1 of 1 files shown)


sumpfralle:

Your change would indeed fix this specific issue of dashes. But I would be very happy, if you could extend this by using the munin-specific approach:
* add the following line to the top of the plugin (above `need_multigraph`):
``` shell
```
* use `fsname=$(clean_fieldname "$(echo "$i" | sed 's/\//__/g')")` instead of your change proposal

Please note, that the above keeps the current field names unchanged (due to the old `sed` call). In a better world, the original author would have used `clean_fieldname` right from the start to make the thing even simpler.

I lack access to hosts using ZFS - thus I am asking you to change and test the above, if you feel like.

Otherwise I am also willing to just merge your proposal as it is, since it is a real improvement.

What do you think?

Sephtex:
Sure! One of the following days I'll make the extra changes and update the PR so I can extensively test it too.

I suppose the clean_fieldname will filter out (or replace) all the characters which aren't accepted for the multigraph argument? As if that might be the case, I could test it without the sed, as I suppose the original submitter used ´sed 's/\//__/g'´ since slashes wouldn't be allowed either for the argument.

Thanks for your quick response btw!

Sephtex:
I made the extra improvements, like this it all works fine now!

plugins/zfs/zfs_list
```diff
@@ -75,9 +77,9 @@ fi
 
 for i in `zfs list -Hp | awk '{print $1}'`; do
 	values=( $(zfs get -p usedbydataset,usedbychildren,usedbysnapshots,usedbyrefreservation,available,quota $i | awk 'BEGIN {total=0;} { if( NR==1 ) next; } !/quota/ {total=total+$3;} {print $3} END{print total;}') )
-	fsname=`echo $i | sed 's/\//__/g'`
+	fsname=$(clean_fieldname $(echo "$i" | sed 's/\//__/g'))
 
-echo <<EOF "multigraph zfs_list_$fsname
+	echo <<EOF "multigraph zfs_list_$fsname
```
sumpfralle: Did you change the indentation here on purpose?

Sephtex: Yes to align with the same echo as done with the echo in the for loop as done above.
It'll also be easier to read, as it'll be a bit less confusing I suppose for someone who is not a bash expert.

Since this is a bugfix, and if you don't want to see that indentation fix in this bugfix, I could revert and just leave it as it won't change anything besides making it structure wise better.

sumpfralle: OK. So it just looked weird on first sight for me. Thank you for your explanation.

sumpfralle:

PR #971, opened 2019-02-09
Change shebang of systemd/systemd_units from sh to bash

ProgVal:
Running it with sh results in this error:

```
/etc/munin/plugins/systemd_units: 109: /etc/munin/plugins/systemd_units: Bad substitution
```


sumpfralle:
Thank you!

PR #973, opened 2019-03-12
Plugin postfix-rbl-blocked-mails: adds custom RBLs monitoring

GuillaumeSMedia:
With this commit it's now possible to add custom Realtime Blackhole Lists via a env.rbls


sumpfralle:
Thank you!

PR #974, opened 2019-03-18
Fix "TabError: inconsistent use of tabs and spaces in indentation"

wodry:
1. Fixes

```
munin-run tor_bandwidth 
  File "/etc/munin/plugins/tor_bandwidth", line 448
    onr = 0
          ^
TabError: inconsistent use of tabs and spaces in indentation
```
2. Use single quotes consistently also for GeoIP.dat filename

PR #975, opened 2019-03-18
Remove quotes in plugin env parameters

wodry:
Setting
`env.torconnectmethod 'port'`
leads to the error:
```
munin-run tor_bandwidth 
env.torconnectmethod contains an invalid value. Please specify either 'port' or 'socket'.
```
Removing the quotes solves that, because the quotes are part of the variable value.
["There is no need to quote the variable content."](http://guide.munin-monitoring.org/en/stable-2.0/plugin/use.html)

### DIFF
--- a/plugins/tor/tor_
+++ b/plugins/tor/tor_
@@ -27,23 +27,23 @@ The default configuration is below
 [tor_*]
 user toranon # or any other user/group that is running tor
 group toranon
-env.torcachefile 'munin_tor_country_stats.json'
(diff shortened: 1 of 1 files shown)


sumpfralle:
Sorry - I just merged your other pull request (#974) - thus the above commit no longer applies cleanly.
Could you update it, please?

wodry:
Rebased & Merged

sumpfralle:
Thanks!

PR #976, opened 2019-03-19
Exit when retrieving updates fails.

bertptrs:
Previously, when retrieving update counts failed, the plugin reported no updates. This is incorrect. With this patch, the plugin should properly report UNKNOWN.

No additional error handling is needed, since `checkupdates` reports a clear error message when it fails.

### DIFF
--- a/plugins/other/pacman_pending_updates
+++ b/plugins/other/pacman_pending_updates
@@ -54,7 +54,7 @@ EOM
 		;;
 
 	*)
-		updates="$(checkupdates)" || exit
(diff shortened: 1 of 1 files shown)


sumpfralle:
Sorry for letting this fix rest for so long ...

Indeed it would be good, if the plugin reported `unknown`. For munin this means to set a value of `U` (the magic string for "unknown").

Would you mind to change the patch accordingly?

bertptrs:
I had almost forgotten I still had this open. I've made the change.

sumpfralle:
thank you!

PR #977, opened 2019-03-26
DOCSIS status monitoring

heeplr:
MUNIN Plugin to monitor status of Arris TG3442 / TG2492LG-85 and compatible cable modems

### DIFF
--- a/plugins/router/arris-tg3442
+++ b/plugins/router/arris-tg3442
@@ -1,49 +1,65 @@
 #!/usr/bin/env python3
 
 """
-# MUNIN Plugin to monitor status of Arris TG3442 / TG2492LG-85
-# and compatible cable modems
-#
-# Connect to the web-frontend and get current DOCSIS status of upstream and
-# downstream channels. (Signal Power, SNR, Lock Status)
-#
-#
-# Requirements:
-#   - BeautifulSoup
-#   - pycryptodome
-#
-# Configuration:
-#   [arris]
-#   env.url http://192.168.100.1
-#   env.username admin
-#   env.password yourpassword
-#
-# Parameters:
-#   url      - URL to web-frontend
-#   username - defaults to "admin"
-#   password - valid password
-#
-#
-# References: https://www.arris.com/products/touchstone-tg3442-cable-voice-gateway/
-#
-#
-#
-# Copyright (c) 2019 Daniel Hiepler <d-munin@coderdu.de>
-#
-# Permission to use, copy, and modify this software with or without fee
-# is hereby granted, provided that this entire notice is included in
-# all source code copies of any software which is or includes a copy or
-# modification of this software.
-#
-# THIS SOFTWARE IS BEING PROVIDED "AS IS", WITHOUT ANY EXPRESS OR
-# IMPLIED WARRANTY. IN PARTICULAR, NONE OF THE AUTHORS MAKES ANY
-# REPRESENTATION OR WARRANTY OF ANY KIND CONCERNING THE
-# MERCHANTABILITY OF THIS SOFTWARE OR ITS FITNESS FOR ANY PARTICULAR
-# PURPOSE.
-#
-#
-# Magic markers
-#  #%# family=contrib
+=head1 NAME
+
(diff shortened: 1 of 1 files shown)


plugins/router/arris-tg3442
```diff
@@ -0,0 +1,250 @@
...
+            "title": "DOCSIS Downstream signal/noise ratio",
+            "vlabel": "dB",
+            "info": "SNR/MER",
+            "data": downstream,
+            "key": "SNRLevel"
+        }
+    ]
+
+    # configure ?
+    if len(sys.argv) > 1 and "config" in sys.argv[1]:
```
sumpfralle: Here `config == sys.argv[1]` is probably, what you intended?
(instead of `in`)

plugins/router/arris-tg3442
```diff
@@ -0,0 +1,250 @@
...
+        "credential",
+        "eyAidW5pcXVlIjoiMjgwb2FQU0xpRiIsICJmYW1pbHkiOiI4NTIiLCAibW9kZWxuYW1lIjoiV"
+        "EcyNDkyTEctODUiLCAibmFtZSI6InRlY2huaWNpYW4iLCAidGVjaCI6dHJ1ZSwgIm1vY2EiOj"
+        "AsICJ3aWZpIjo1LCAiY29uVHlwZSI6IldBTiIsICJnd1dhbiI6ImYiLCAiRGVmUGFzc3dkQ2h"
+        "hbmdlZCI6IllFUyIgfQ=="
+    )
+
+    # set session
+    r = session.post(f"{url}/php/ajaxSet_Session.php")
+
```
sumpfralle: PEP8 recommends two empty lines between top-level objects.

plugins/router/arris-tg3442
```diff
@@ -0,0 +1,250 @@
+#!/usr/bin/env python3
+
+"""
+# MUNIN Plugin to monitor status of Arris TG3442 / TG2492LG-85
```
sumpfralle: Since you are using a docstring (`"""`), you can remove the leading hashs (`#`).

plugins/router/arris-tg3442
heeplr: It was intended but I admit, it might be a bit too defensive.

plugins/router/arris-tg3442
heeplr: aye

plugins/router/arris-tg3442
heeplr: aye

sumpfralle:
It looks good to me. Thanks!

One last thing, that I overlooked before: you are specifying a static `session.cookies`. Does it contain a secret that is bound to an account?

heeplr:
There is only one default account. It seems to be static. It's base64 that decodes to:
`{ "unique":"280oaPSLiF", "family":"852", "modelname":"TG2492LG-85", "name":"technician", "tech":true, "moca":0, "wifi":5, "conType":"WAN", "gwWan":"f", "DefPasswdChanged":"YES" }`

If it's a problem across models, additional logic + base64 encoding is necessary.

sumpfralle:
In this case it would be good to add this original value (and the conversion command) as a comment to the plugin.

Isn't it weird, that access requires two different user/password combinations? (one given via environment variables and one stored in the cookie)

heeplr:
> In this case it would be good to add this original value (and the conversion command) as a comment to the plugin.

If you think that helps and is legally ok, feel free to add it.

> Isn't it weird, that access requires two different user/password combinations? (one given via environment variables and one stored in the cookie)

Yes. It's questionable design.

sumpfralle:
> If you think that helps and is legally ok, feel free to add it.

Yes, I will do it.
Otherwise I would have the feeling, that the nice plugin contains an evil binary blob and thus dilutes the free software license attached to this plugin.

Besides this: should I merge it now or is there something on your list to be done before?

sumpfralle:
Finalized in 69602b1bbf80c4376ce35cf2b5b4532193379205.

PR #979, opened 2019-04-02
libvirt/kvm_*: Simpler regex to match VM name

benpro:
On Debian Stretch the cmdline is something like:
`qemu-system-x86_64-enable-kvm-nameguest=vmname,debug-threads=on-S-[…]`
Without null characters:
`qemu-system-x86_64 -enable-kvm -name guest=vmname,debug-threads=on[…]`

We need to match only `guest=vmname`, so the regex
`^.*guest=([a-zA-Z0-9.-_-]*).*$` is simpler and match the VM name.
The previous regex `^.*-name\x00([a-zA-Z0-9.-_-]*)\x00\-.*$` was not matching it.

BTW, It seems that `\x00` does not match correctly null characters so I removed it.

### DIFF
--- a/plugins/libvirt/kvm_cpu
+++ b/plugins/libvirt/kvm_cpu
@@ -68,7 +68,7 @@ def find_vm_names(pids):
     result = {}
     for pid in pids:
         cmdline = open("/proc/%s/cmdline" % pid, "r")
-        result[pid] = clean_vm_name(re.sub(r"^.*-name\x00([a-zA-Z0-9.-_-]*)\x00\-.*$",r"\1", cmdline.readline()))
+        result[pid] = clean_vm_name(re.sub(r"^.*guest=([a-zA-Z0-9.-_-]*).*$",r"\1", cmdline.readline()))
     return result
 
 def list_pids():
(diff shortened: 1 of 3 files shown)


sumpfralle:

> BTW, It seems that \x00 does not match correctly null characters so I removed it.

I tried it here, and it seems to work with python and python3:
```shell
python3 -c 'import re; print(re.search(r"o\x00b", "foo\x00bar").span())'
```

Thus maybe you misinterpreted a detail?

Anyway: if you are certain, that this part of the regular expression is not relevant, then we can skip it, if course.

Another detail: your first commit is a merge. You should get rid of it by rebasing, I guess.

benpro:
> Thus maybe you misinterpreted a detail?

Yes maybe.

> Anyway: if you are certain, that this part of the regular expression is not relevant, then we can skip it, if course.

Yes it is not relevant.
What we want if matching `guest=vmname` in a long cmdline like:

```
qemu-system-x86_64-enable-kvm-nameguest=vmname,debug-threads=on-S-objectsecret,id=masterKey0[…]
```

This command line include null character… But maybe we don't care for the pattern to search?  
Parameters are separated by null character so the pattern `^.*guest=([a-zA-Z0-9.-_-]*).*$` should be OK.

> Another detail: your first commit is a merge. You should get rid of it by rebasing, I guess.

I'm gonna check out that.

benpro:
Rebase has been done.

sumpfralle:
great - thank you!

PR #982, opened 2019-05-25
Add ntp_drift.

thermes:
This commit adds a plugin for `ntpd driftfile` .

### DIFF
--- a/plugins/ntp/ntp_drift
+++ b/plugins/ntp/ntp_drift
@@ -1,0 +1,85 @@
+#! /bin/sh
+# -*- sh -*-
+
+: <<=cut
+
+=head1 NAME
+
+ntp_drift - Munin plugin to monitor the NTP drift value.
+
+=head1 APPLICABLE SYSTEMS
+
+Any ntpd host.
+
+=head1 CONFIGURATION
+
+The following configuration parameters are used by this plugin:
+
+ [ntp_drift]
+ env.driftfile - Path to driftfile.
+
+=head2 DEFAULT CONFIGURATION
+
+ [ntp_drift]
+ env.driftfile "/var/lib/ntp/ntp.drift"
+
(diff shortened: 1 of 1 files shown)


sumpfralle:
This plugin also looks good to me!

Please take a look at my comments for #983. Most apply here, too.

btw: do you think it is possible to use a single plugin for both ntp servers, without complicating the code too much?
(maybe the drift file location could be guessed and the format applied based on the content of the file?)

thermes:
@sumpfralle 
Thank you for the review.
The review points have been fixed. I also used `shellcheck`. Please check again.
(Sorry for the late response.)

thermes:
@sumpfralle 
Thank you for your feedback.
I think it is possible to combine plugins for chrony and ntpd driftfile. However, chrony and ntpd speak the same protocol but are different products. If we put plugins together, munin users may get confused.
For that reason, I think it would be better to have separate plugins.

plugins/ntp/ntp_drift
```diff
@@ -0,0 +1,87 @@
...
+        echo "drift.value U"
+        exit 1
+    fi
+}
+
+
+case $1 in
+    autoconf|config|'')
+        do_"$1"
+        exit $?
```
sumpfralle: Since you do not use `return` (or potentially non-successful) statements in your functions, this line will be of no use.

Additionally the exit code of the last executed statement will be the exit code of the script anyway (thus it would be a no-op in this situation).

Thus `exit 0` would be a suitable line (if you want to be explicit). Or you just remove the `exit` line completely.

sumpfralle:
btw: feel free to squash your commits into one, when you are ready.

thermes:
@sumpfralle
Thank you for the review.
Fixed the issue and squashed.

sumpfralle:
Great!

PR #984, opened 2019-05-27
Add plugin to monitor Docker container network interface

mapbuh:


plugins/docker/docker_if_
```diff
@@ -0,0 +1,216 @@
...
+        exit 0
+        ;;
+    config)
+
+        echo "graph_order down up"
+        echo "graph_title Container $CONTAINER traffic"
+        echo 'graph_args --base 1000'
+        echo 'graph_vlabel bits in (-) / out (+) per ${graph_period}'
+        echo 'graph_category network'
+        echo "graph_info This graph shows the traffic of the $CONTAINER containre. Please note that the traffic is shown in bits per second, not bytes. IMPORTANT: On 32-bit systems the data source for this plugin uses 32-bit counters, which makes the plugin unreliable and unsuitable for most 100-Mb/s (or faster) interfaces, where traffic is expected to exceed 50 Mb/s over a 5 minute period.  This means that this plugin is unsuitable for most 32-bit production environments. To avoid this problem, use the ip_ plugin instead.  There should be no problems on 64-bit systems running 64-bit kernels."
```
sumpfralle: typo: `container`

plugins/docker/docker_if_
```diff
@@ -0,0 +1,216 @@
...
+
+        findspeed
+
+        exit 0
+        ;;
+esac
+
+# Escape dots in the interface name (eg. vlans) before using it as a regex
+if [[ -r /sys/class/net/$INTERFACE/statistics/rx_bytes ]]; then
+    echo "down.value $(cat /sys/class/net/$INTERFACE/statistics/rx_bytes)"
```
sumpfralle: You should quote the filename.

plugins/docker/docker_if_
```diff
@@ -0,0 +1,216 @@
...
+        mkdir -p /var/run/netns
+        ln -sf /proc/$pid/ns/net "/var/run/netns/${1}"
+        local index=$(ip netns exec "${1}" ip link show eth0 | head -n1 | sed s/:.*//)
+        let index=index+1
+        z=`ip link show | grep "^${index}:" | sed "s/${index}: \(.*\):.*/\1/"`
+        echo $z | cut -f 1 -d '@'
+        rm -f "/var/run/netns/${1}"
+}
+
+INTERFACE=$(veth_interface_for_container ${CONTAINER})
```
sumpfralle: You should quote the container name.

plugins/docker/docker_if_
```diff
@@ -0,0 +1,216 @@
...
+
+# Who whould have thought it's so much work to determine the
+# maximum speed of a network interface.  Buckle up!
+findspeed_mbps() {
+    # Do not use interface name to guess technology.  Many many
+    # wifi drivers use "eth*" names.
+    IWLIST=$(type -p iwlist)
```
sumpfralle: please use `local`

plugins/docker/docker_if_
```diff
@@ -0,0 +1,216 @@
...
+            *100base*)  echo 100; return ;;
+            *10base*)   echo 10; return ;;
+        esac
+    fi
+
+    echo U
+}
+
+findspeed() {
+    if [[ "$speed" ]]; then
```
sumpfralle: Where could `$speed` come from?
Probably you should use something like `speed=${speed:-}` in the beginning of the script in order to clarify the source of this information.

plugins/docker/docker_if_
```diff
@@ -0,0 +1,216 @@
...
+        else
+                echo " Please install ethtool, wireless-tools, mii-tool or whatever is appropriate for the interface."
+        fi
+
+        return
+    fi
+
+    BPS=$(( $SPEED * 1000 * 1000 ))
+
+    cat <<EOF
```
sumpfralle: I would recommend (for clarity) to rename this function (e.g. `get_interface_bandwidth`) and return only the `BPS` value. The caller should afterwards put together the "config" lines (`u.max`, `u.info` and so on).
This will be easier to maintain in the future.

plugins/docker/docker_if_
```diff
@@ -0,0 +1,216 @@
+#!/bin/bash
+# -*- sh -*-
+
+: << =cut
+
+=head1 NAME
+
+if_ - Wildcard-plugin to monitor traffic on docker container network
```
sumpfralle: probably `docker_if_`?

mapbuh:
Thank you for you detailed and thorough answer. It is indeed a quick hack over if_ plugin.
I'm considering rewriting it from scratch, but I currently cannot find clean way to map container name to network interface.

sumpfralle:
Yes, docker seems to not help a lot in this regard :(
I took a look at https://github.com/moby/moby/issues/14743 - you implemented this approach, as far as I understand. The only missing detail should be something along the following:
```shell
docker exec -it NAME sh -c 'set -e; cd /sys/class/net; for net_if in *; do cmp -s "$net_if/ifindex" "$net_if/iflink" || cat "$net_if/iflink"; done'
```

Your rewrite of the plugin is more than welcome!

PR #985, opened 2019-05-28
[router/freebox] Plugin to monitor Freebox information

shtrom:
Some example graphs

Screenshot from 2019-05-28 21-48-15
Screenshot from 2019-05-28 21-48-22
Screenshot from 2019-05-28 21-48-28


plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+done
+
+fbx_info_url="http://mafreebox.free.fr/pub/fbx_info.txt"
+
+fetch() {
+	# shellcheck disable=SC2154
+	curl -s "$@"
+}
+
+# XXX: testing
```
sumpfralle: This line and the next one can probably be removed?

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+# INFO="$( iconv -f latin1 -t utf8 fbx_info.txt)"
+INFO="$(fetch "${fbx_info_url}" | iconv -f latin1 -t utf8)"
```
sumpfralle: Maybe you could provide an example of this output as a comment?

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+
+eval $(
```
sumpfralle: I am confused: what is the point of this `eval`?

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+		s#^\s*WAN.*\s([0-9]+)\s*ko/s\s*([0-9]+)\s*ko/s#WAN_DOWN="0\1" WAN_UP="0\2"#p;
+		s#^\s*Ethernet.*\s([0-9]+)\s*ko/s\s*([0-9]+)\s*ko/s#ETH_DOWN="0\1" ETH_UP="0\2"#p;
+		s#^\s*USB.*\s([0-9]+)\s*ko/s\s*([0-9]+)\s*ko/s#USB_DOWN="0\1" USB_UP="0\2"#p;
+		s#^\s*Switch.*\s([0-9]+)\s*ko/s\s*([0-9]+)\s*ko/s#SWITCH_DOWN="0\1" SWITCH_UP="0\2"#p;
+	' <<< "${INFO}"
+)
+
+DHCP_CLIENTS=$(sed -nE '/Attributions dhcp/,/Redirections de ports/{s/^\s*([0-9A-F:]{17}).*$/\1/p}' <<< "${INFO}" | wc -l)
+REDIRECT_TCP=$(sed -nE '/Redirections de ports/,/Interfaces réseau/{s/^\s*(TCP).*$/\1/p}' <<< "${INFO}" | wc -l)
+REDIRECT_UDP=$(sed -nE '/Redirections de ports/,/Interfaces réseau/{s/^\s*(UDP).*$/\1/p}' <<< "${INFO}" | wc -l)
```
sumpfralle: The `sed` command `for `REDIRECT_UDP` looks like it would count the sum of the TCP and UDP redirects (since the start line pattern is equal).

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+
+FREEBOX_UPTIME=$(echo "${UPTIME_DAYS}*86400+${UPTIME_HOURS}*3600+${UPTIME_MINUTES}*60" | bc -ql)
```
sumpfralle: Maybe `FREEBOX_UPTIME=$((UPTIME_DAYS * 86400 + UPTIME_HOURS * 3600 + UPTIME_MINUTES * 60))` would be more readable (and not require `bc`)?

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
+#!/bin/bash
+# -*- sh -*-
+
+: << =cut
+
+=head1 NAME
+
+freebox - Plugin to monitor stats of a Freebox
```
sumpfralle: Maybe you could provide a bit of details, what this `Freebox` is?

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+
+The ideal usage is also used as an updated warning limit.
```
sumpfralle: Could you rephrase this, please?
(I do not understand it spontaneously)

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+
+DHCP_CLIENTS=$(sed -nE '/Attributions dhcp/,/Redirections de ports/{s/^\s*([0-9A-F:]{17}).*$/\1/p}' <<< "${INFO}" | wc -l)
+REDIRECT_TCP=$(sed -nE '/Redirections de ports/,/Interfaces réseau/{s/^\s*(TCP).*$/\1/p}' <<< "${INFO}" | wc -l)
+REDIRECT_UDP=$(sed -nE '/Redirections de ports/,/Interfaces réseau/{s/^\s*(UDP).*$/\1/p}' <<< "${INFO}" | wc -l)
+
+FREEBOX_UPTIME=$(echo "${UPTIME_DAYS}*86400+${UPTIME_HOURS}*3600+${UPTIME_MINUTES}*60" | bc -ql)
+LAST_CONNECT=$(sed -nE '/Journal de connexion adsl/,+5{s#^\s*([0-9]{2})/([0-9]{2})/([0-9]{4})[^0-9]*([0-9]+):([0-9]+):([0-9]+)\s+Connexion.*#\3-\2-\1T\4:\5:\6#p}' <<< "${INFO}")
+CONNECTION_UPTIME=na
+if [ -n "${LAST_CONNECT}" ]; then
+	FORMULA="$(date +%s)-$(date -d "${LAST_CONNECT}" +%s)"
```
sumpfralle: I could imagine, that the following is as readable as your suggestion (without requiring `bc`):
```
FORMULA=$(( $(date +%s) - $(date -d "$LAST_CONNECT" + %s) ))

File: plugins/router/freebox [hunk 9]
```diff
@@ -0,0 +1,316 @@
...
+			echo 'redirect_tcp.label TCP redirections'
+			echo 'redirect_udp.label UDP redirections'
+			;;
+		*)
+			echo 'graph_title Freebox connection'
+			echo 'graph_category network'
+			echo 'graph_vlabel bits per second in (+) / out (-)'
+			echo 'graph_order main_atm_down main_atm main_wan_down main_wan'
+
+			# XXX: parameters duplicated from traffic
```
sumpfralle: Why are these duplicated?

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+		.users)
+			echo 'graph_title Network users'
+			echo 'graph_category network'
+			echo 'graph_vlabel count'
+
+			echo 'dhcp.label DHCP leases'
+			echo 'redirect_tcp.label TCP redirections'
+			echo 'redirect_udp.label UDP redirections'
+			;;
+		*)
```
sumpfralle: Is this supposed to be the graph `''` (empty string)? How about using `''` (two single quotes) instead?

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+=cut
+
+# shellcheck disable=SC1090
+. "${MUNIN_LIBDIR}/plugins/plugin.sh"
+
+if [ "${MUNIN_DEBUG}" = 1 ]; then
+    set -x
+fi
+
+for CMD in curl iconv bc; do
```
sumpfralle: Please move this check to `autoconf`. Now it would cause the `autoconf` call to fail. This should not happen.

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+	echo
+}
+
+main() {
+	case $1 in
+		autoconf)
+			if curl -qso /dev/null $fbx_info_url; then
+				echo 'yes'
+			else
+				echo 'no'
```
sumpfralle: Please provide an explanation for `no`, e.g.: `no (failed to retrieve '$fbx_info_url')`

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+
+
+	esac
+	echo
+}
+
+main() {
+	case $1 in
+		autoconf)
+			if curl -qso /dev/null $fbx_info_url; then
```
sumpfralle: Please quote the URL.

plugins/router/freebox
```diff
@@ -0,0 +1,316 @@
...
+			graph_data adsl_errors
+			graph_data traffic
+			graph_data uptime
+			graph_data users
+			;;
+	esac
+
+}
+
+main $1
```
sumpfralle: Please quote `$1`. Maybe you even want to use `main "${1:-}"`.
In this case you should almost be able to enable "error on unset" (via `set -u`).

plugins/router/freebox
shtrom: The `sed` just create a shell snippet that defines all the variables. The `eval` brings them into the context of the current script.

plugins/router/freebox
shtrom: Nope, both are in the same section (`Redirection de ports`), but the `s///p` outputs only lines with matching replacements (so, either `TCP` or `UDP`).

plugins/router/freebox
shtrom: It would be, wouldn't it. I've always assumed that it was a `bash`ishm, but it actually seems to work with `sh`... TIL.

plugins/router/freebox
shtrom: It's a catch-all, I prefer to have that, to always output something, rather than be more limitative. Granted, there is little chance that anything else that `''` will ever get there, but I find it safer.

plugins/router/freebox
shtrom: And I'm using `bash` anyway :man_facepalming:

plugins/router/freebox
shtrom: Ah, this was from a different plugin I took as a base to work on this one. It's not relevant here.

plugins/router/freebox
shtrom: It's a multigraph, so I have those both in the traffic subgraph, and the main graph.

I'm curious to know if there is any other way to do this.

shtrom:
@sumpfralle good for another look. I left the fixup commit there with all the changes for the moment, to ease incremental reviews; I'll squash at the end.

plugins/router/freebox
sumpfralle: Thank you for the clarification.
Personally I feel quite confident with writing shell scripts, but honestly I could only guess, what is happening here (due to the lack of understanding of the input text). This will be very hard for future contributors of this plugin :(
Maybe you can think of a more maintainable approach for this part?

plugins/router/freebox
sumpfralle: OK - thank you.

plugins/router/freebox
sumpfralle: I am more of a python person - thus I like it, when the program informs me when an older assumption is not true anymore. Anyway, it is just a matter of style ...

plugins/router/freebox
sumpfralle: Ah - ok. I did not get, that the duplication is in the main graph. This is indeed fine.

plugins/router/freebox
```diff
@@ -0,0 +1,456 @@
...
+graph_data() {
+	graph=""
+	if [ -n "${1:-}" ]; then
+		graph=".$1"
+	fi
+
+	echo "multigraph freebox${graph}"
+	case "$graph" in
+		.adsl)
+			echo "noise.value ${NOISE_UP:-nan}"
```
sumpfralle: Instead of `nan` you should use `U` (unknown value).

plugins/router/freebox
shtrom: Sadly, no. The intput text (of which I added an example above) is very non machine-parsable, so it required a lot of `sed`-trickeries to get the relevant data out of it. I can add a comment explaining the generic idea, but yeah, essentially `sed` is used to extract the data and put it in the form of a number of variable assignments, for `bash` to evaluate.

shtrom:
@sumpfralle done: I added a comment with an example about the `eval $(sed ...)`, and replaced all `nan`s with `U` (thanks for that, I have been looking for the correct placeholder for a while, but couldn't find it anywhere.

If you're happy to merge, feel free to squash the fixups first.

plugins/router/freebox
sumpfralle: How about the following?
```shell
CRC_DOWN=$(echo "$INFO" | awk '/^\s+CRC\s/ { print $(NF-1); }')
CRC_UP=$(echo "$INFO" | awk '/^\s+CRC\s/ { print $NF; }')
```

Or maybe something even more readable?
```shell
get_line_column() {
    local key="$1"
    local field_index="$2"
    awk '/^\s+'"$key"'/ { print $('"$field_index"'); }'
}
...
CRC_DOWN=$(echo "$INFO" | get_line_column "CRC" "NF-1")
CRC_UP=$(echo "$INFO" | get_line_column "CRC" "NF")
```

Feel free to decide, whether such an approach could be suitable.

shtrom: Ok, this mostly works, but we lose validation of the fields. My `sed` script was ensuring that the extracted data was numeric, while the `awk` way gets whatever is in the column. This means that there is some bad data in case the output file doesn't have any.

For example, with the disconnected USB data:
```USB                    Non connecté```
we now get
```usb.value Non
usb_down.value   USB                    Non connecté 
```

And I still need the `eval $(sed)` to extract the uptime.

It seem to be slightly slower, too (by 100ms or so) [this is with a local dump of the file, rather than the network]:
```
[22:40:49] ~/src/munin-contrib/plugins/router$ for i in `seq 10`; do time MUNIN_LIBDIR=.  ./freebox &> /dev/null; done                                               130 ↵ ✹ router/freebox u+1 
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.14s user 0.08s system 104% cpu 0.214 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.19s user 0.05s system 105% cpu 0.228 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.17s user 0.05s system 105% cpu 0.208 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.17s user 0.07s system 104% cpu 0.229 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.14s user 0.07s system 104% cpu 0.204 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.17s user 0.05s system 105% cpu 0.209 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.17s user 0.07s system 104% cpu 0.224 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.17s user 0.09s system 104% cpu 0.251 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.16s user 0.09s system 103% cpu 0.240 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.16s user 0.06s system 104% cpu 0.201 total
[22:40:54] ~/src/munin-contrib/plugins/router$ git checkout -f HEAD^                                                                                                   
HEAD is now at afb3a6e4 fixup! fixup! [router/freebox] Plugin to monitor Freebox information
[22:40:59] ~/src/munin-contrib/plugins/router$ vim freebox                                                                                                   0s  (remotes/origin/router/freebox)
[22:41:06] ~/src/munin-contrib/plugins/router$ for i in `seq 10`; do time MUNIN_LIBDIR=.  ./freebox &> /dev/null; done                                     6s  ✹ (remotes/origin/router/freebox)
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.05s user 0.02s system 112% cpu 0.056 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.05s user 0.03s system 110% cpu 0.074 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.03s user 0.03s system 112% cpu 0.052 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.06s user 0.01s system 113% cpu 0.067 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.05s user 0.02s system 112% cpu 0.061 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.05s user 0.01s system 112% cpu 0.056 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.05s user 0.02s system 112% cpu 0.065 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.05s user 0.02s system 112% cpu 0.057 total
MUNIN_LIBDIR=. ./freebox &> /dev/null  0.04s user 0.02s system 112% cpu 0.059 total
```

shtrom: I pushed those changes. I'll let it run for a day or so to make sure I still get all the data, and let you know.

sumpfralle: Maybe the following is sufficient for numeric validation and proper handling of *unknown* values?
Bonus: only to be done once in the function :)

```shell
local result
result=$(awk '/^\s+'"$key"'/ { print $('"$field_index"'); }')
if echo "$result" | grep -q '^[0-9]\+$'; then
    echo "$result"
else
    echo "U"
fi
```

Regarding the uptime:
```shell
UPTIME_DAYS=$(echo "$INFO" | get_line_column "Temps depuis la mise en route" "NF-5")
UPTIME_HOURS=$(echo "$INFO" | get_line_column "Temps depuis la mise en route" "NF-3")
UPTIME_MINUTES=$(echo "$INFO" | get_line_column "Temps depuis la mise en route" "NF-1")
```

Maybe I am a bit obsessed about code readability :)

shtrom: Hum, I suppose that could work for the uptime. As for the validation, would there be a more direct way to match the regexp straight within `awk`? The `grep` should work, but it feels like a lot of overhead considering we already parsed the data out.

shtrom: Ok, I found a way. How about now?

sumpfralle: Yes, this should work easily by prepending the `print` call with something like `if ($('"$field_index"') =~ /^[0-9]+$/)`.

sumpfralle: This looks good to me!
Just one thing: maybe you need to add `^` in front of `/[.0-9]+/`? (otherwise `foo4` would match?)

Thank you for bearing with my wishes! I am happy to merge it: just say so.

shtrom: Ah, good catch on the `^...$` in the RE.

It's all fixed now, and squashed, so should be ready to merge :+1: .

sumpfralle:
Thank you for your patience!

PR #986, opened 2019-05-28
Plugin xen-multi: xentop bug workaround

neszt:
There is a bug in xentop, its output is too high in rare cases.

https://lists.xenproject.org/archives/html/xen-users/2019-04/msg00020.html

Workaround to process 3 iterations after the first one (4 itrations) and choose the lowest cpusecs value line.

### DIFF
--- a/plugins/xen/xen-multi
+++ b/plugins/xen/xen-multi
@@ -129,7 +129,12 @@ sub trim_label {
 # Global variables
 my (%domains,$domain,@domainlist,$munindomain,$cpusecs,$cpupercent,$memk,$nettxk,$netrxk,$vbdrd,$vbdwr);
 
-open (XENTOP,"xentop -b -f -i2 |") or die "Could not execute xentop, $!";
+# There is a bug in xentop, its output is too high in rare cases.
+# https://lists.xenproject.org/archives/html/xen-users/2019-04/msg00020.html
+# Workaround to process 3 iterations after the first one (4 iterations) and
+# choose the lowest cpusecs value line.
+
+open (XENTOP,"xentop -b -f -i4 |") or die "Could not execute xentop, $!";
 
 # Now we build a hash of hashes to store information
 while (<XENTOP>) {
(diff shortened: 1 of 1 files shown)


plugins/xen/xen-multi
```diff
@@ -129,7 +129,12 @@ sub trim_label {
 # Global variables
 my (%domains,$domain,@domainlist,$munindomain,$cpusecs,$cpupercent,$memk,$nettxk,$netrxk,$vbdrd,$vbdwr);
 
-open (XENTOP,"xentop -b -f -i2 |") or die "Could not execute xentop, $!";
+# There is a bug in xentop, its output is too high in rare cases.
+# https://lists.xenproject.org/archives/html/xen-users/2019-04/msg00020.html
+# Workaround to process 3 iterations after the first one (4 itrations) and
```
sumpfralle: typo: "itrations"

plugins/xen/xen-multi
```diff
@@ -146,13 +151,16 @@ while (<XENTOP>) {
...
   # We need the remaining data only for a normal run
   if ($ARGV[0] eq "") {
-    $domains{$domain}{'cpusecs'} = $cpusecs;
-    $domains{$domain}{'cpupercent'} = $cpupercent;
-    $domains{$domain}{'mem'} = $memk;
-    $domains{$domain}{'nettx'} = $nettxk;
-    $domains{$domain}{'netrx'} = $netrxk;
-    $domains{$domain}{'vbdrd'} = $vbdrd;
-    $domains{$domain}{'vbdwr'} = $vbdwr;
+    $domains{$domain}{'cnt'}++;
```
sumpfralle: Maybe you should use a separate variable (outside of this hash) in order to keep the code more readable?

plugins/xen/xen-multi
```diff
@@ -146,13 +151,16 @@ while (<XENTOP>) {
...
+    if ( $domains{$domain}{'cnt'} > 1 && ( !defined $domains{$domain}{'cpusecs'} || $cpusecs < $domains{$domain}{'cpusecs'} ) ) {
```
sumpfralle: Please add a comment here, explaining this logic. Or maybe just move the "4 iterations" comment from above down to this line?

plugins/xen/xen-multi
neszt: Fixed

plugins/xen/xen-multi
neszt: Fixed

plugins/xen/xen-multi
neszt: The cnt key must be stored per domain and the $domain hash just has exactly the same data structure, so IMHO it is more readable using the $domain variable than making a new variable for it.

sumpfralle:
Great - thank you!

Would you please mind to squash your commits into one (and force-push the branch)?

sumpfralle:
Thank you!

PR #987, opened 2019-06-23
[file_length_] Generic plugin to monitor the number of lines (or variation thereof) of arbitrary files

shtrom:

### DIFF
--- a/plugins/system/file_length_
+++ b/plugins/system/file_length_
@@ -8,13 +8,17 @@ file_length_ - Plugin to monitor the length of specified files
 
 Useful for things such as lists (white, black, user, ...).
 
+The type of metric defaults to GAUGE, but it can be configured to, say, DERIVE,
+which is useful for monitoring the rate of message in error logs.
+
 =head1 CONFIGURATION
 
     [file_length_IDENTIFIER]
     env.files FILEPATHGLOB1 FILEPATHGLOB2 ...
-    env.category DEFAULTS_TO_system
+    env.category system
     env.title OPTIONAL_TITLE
     env.logarithmic 1
+    env.type GAUGE
 
 =head1 AUTHOR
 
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for contributing to this plugin!

Do you have an idea, *why* it could be useful to use something else, than `GAUGE` for the field type?

shtrom:
Oh yeah, I just added this because I wanted to monitor error log files. In this case, the number of lines is a bit pointless, particularly with log rotation. However, the rate of increase, with `DERIVE` can be indicative of issues if there are spikes and periods of higher error rates than usual.

plugins/system/file_length_
```diff
@@ -15,6 +15,7 @@ Useful for things such as lists (white, black, user, ...).
     env.category DEFAULTS_TO_system
     env.title OPTIONAL_TITLE
     env.logarithmic 1
+    env.type DEFAULT_TO_GAUGE
```
steveschnepp: Make it `GAUGE` instead of `DEFAULT_TO_GAUGE`

steveschnepp:
>  However, the rate of increase, with `DERIVE` can be indicative of issues if there are spikes and periods of higher error rates than usual.

I'd even argue that _indeed_ `DERIVE` might be better default than `GAUGE`...

plugins/system/file_length_
steveschnepp: As I said, `DERIVE` looks like a better default. But I don't really use the plugin, so I don't really vote for any.

sumpfralle:
> I'd even argue that indeed DERIVE might be better default than GAUGE...

But only for error log files, I guess. There can be other uses of this plugin (watching the number of lines in an htpasswd file or similar).
How about adding a line of comment to the documentation, that mentions the use case for `DERIVE`?

shtrom:
I think given the name and the fact that the plugin was already in, we should keep it as `GAUGE`, otherwise we risk breaking some use-cases relying on it being the default.

I'll add the comment (and a bugfix) a bit later on 👍

sumpfralle:
That looks good to me - thank you!

PR #988, opened 2019-06-27
[file_lenght_] Various fixes

shtrom:
* Only show extinfo if stat is the GNU version
* Unify variable references
* Better legend when type is not GAUGE

### DIFF
--- a/plugins/system/file_length_
+++ b/plugins/system/file_length_
@@ -36,20 +36,20 @@ SPDX-License-Identifier: GPL-2.0-only
 
 # needs shellcheck -x /usr/share/munin/plugins/plugin.sh
 # shellcheck source=/usr/share/munin/plugins/plugin.sh
-. "$MUNIN_LIBDIR/plugins/plugin.sh"
+. "${MUNIN_LIBDIR}/plugins/plugin.sh"
 
-NAME=$(echo "$0" | sed 's/.*_//')
-TITLE=${title:-File lengths for $NAME}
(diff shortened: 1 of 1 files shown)


shtrom:
@sumpfralle here are some fixups on yesterday's merge, to make the plugin a bit nicer in DERIVE mode.

sumpfralle:
I am tempted to say, that I prefer the variable references without curly braces (unless they are required) - but that is obviously just a matter of taste :)

Nitpicking: I would have preferred two commits (one for the style - one for the improvements).

But anyway: I would merge it, as it is, if you prefer so.

shtrom:
On Thu 27 Jun 2019 at 14:38:35 -0700, Lars Kruse wrote:
>I am tempted to say, that I prefer the variable references without curly braces (unless they are required) - but that is obviously just a matter of taste :)

Yeah, I think the other way. It's a bit more verbose, but it guarantees 
that your variables are properly delimited. It saved me in a few 
(arguably rare) cases where `$VAR_blah` is not actually the same 
`${VAR}_blah`, so I try to make a habit of always using them.

>Nitpicking: I would have preferred two commits (one for the style - one for the improvements).

Yeah, fair enough. It got bigger than I initially planned as I started 
cleaning up.

>But anyway: I would merge it, as it is, if you prefer so.

Yes, please do, but I'll keep that in mind for next time.

-- 

sumpfralle:
Thank you!

PR #991, opened 2019-07-18
Remove double quote for env.driftfile variable

sflotat:
Information concerning how to use variable env.driftfile is wrong. Specifying these variable as is in plugin-conf.d/00-defaults not work if double quote is used.


sumpfralle:
Indeed - thanks!

PR #993, opened 2019-07-20
[php_fpm_process] Add queue and request information

shtrom:
Breaking changes:
* php_fpm_slowrequests is superseded by php_fpm_requests

### DIFF
--- a/plugins/php/php_fpm_process
+++ b/plugins/php/php_fpm_process
@@ -25,7 +25,7 @@ You have to put this in your plugin.conf.d folder
 
 # If your php process is listening on Unix Socket
 [php_fpm_process]
-   env.sock /var/run/php5-fpm.sock
+   env.sock /var/run/php-fpm.sock
    env.path /status
 
 =head1 MAGIC MARKERS
@@ -35,15 +35,16 @@ You have to put this in your plugin.conf.d folder
 
 =head1 VERSION
 
-  v1.0
+  v2.0.0
 
 =head1 AUTHOR
 
 Minitux
+Olivier Mehani <shtrom+munin@ssji.net>
 
 =head1 LICENSE
 
-GNU General Public License, version 3
+SPDX-License-Identifier: GPL-3.0
 
 =cut
 
@@ -56,6 +57,10 @@ my $body = "";
 my $IDLE = 0;
 my $ACTIVE = 0;
 my $TOTAL = 0;
+my $LISTEN = 0;
+my $MAX = 0;
+my $LEN = 0;
+my $CONNECTIONS = 0;
 my $SLOW_REQUESTS = 0;
 my $PLUGIN_NAME = basename($0);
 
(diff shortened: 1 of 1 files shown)


plugins/php/php_fpm_process
```diff
@@ -122,12 +164,12 @@ if ( defined $ARGV[0] and $ARGV[0] eq "config" )
...
   print <<"EOF";
 multigraph ${PLUGIN_NAME}_process
-graph_title php5-fpm processes for $pool
+graph_title php-fpm processes for $pool
 graph_args --base 1000 -l 0
 graph_vlabel Processes
 graph_scale yes
-graph_category processes
-graph_info This graph shows the php5-fpm process manager status from pool: $pool
+graph_category webserver
```
sumpfralle: I could imagine, that `appserver` would be more appropriate?

plugins/php/php_fpm_process
```diff
@@ -140,19 +182,48 @@ total.label Total processes
 total.type GAUGE
 total.draw LINE2
 total.info The number of idle + active processes
+max.label Max processes
+max.type GAUGE
+max.draw LINE
+max.info The maximum number of processes
```
sumpfralle: This is not very clear to me. Maybe expand it with `... since the start of the FPM process` (if this is what is meant).

plugins/php/php_fpm_process
```diff
@@ -175,10 +246,31 @@ if($body =~ m/total processes: (.*?)\n/) {
         $TOTAL = $1;
   print "total.value ".$TOTAL."\n";
 }
+if($body =~ m/max active processes: (.*?)\n/) {
+        $MAX = $1;
```
sumpfralle: Did you mix whitespace here?
(and in some occurrences below)

plugins/php/php_fpm_process
shtrom: Heh, PHP is so generic that I didn't think about that, but yeah, why not.

plugins/php/php_fpm_process
shtrom: Yes and no. I tried to match the (weird) indent that was already there. I'm more than happy to fix it completely.

shtrom:
Ok, comments addressed. The path to the socket is just in a comment, and there is no default, so this is not actually what breaks, so I don't think there's anything more needed in the commit message.

sumpfralle:
That looks good to me. Do you think, your proposal is ready for squashing/merging?

Regarding the whitespace: this kind of cleanup is always great. I would be even more happy about a separate commit with only whitespace fixes :)
(just wishful thinking for the future ...)

Thanks!

shtrom:
> Do you think, your proposal is ready for squashing/merging?

Yup, I'll squash.

> Regarding the whitespace: this kind of cleanup is always great. I would be even more happy about a separate commit with only whitespace fixes :)

Oh, yeah, I'm the same. This one obviously came as an afterthought (;

sumpfralle:
> Oh, yeah, I'm the same.

perfect! :)

PR #994, opened 2019-07-20
[ssl-certificate-expiry] Add asynchronous update via cron

shtrom:
Also, cleanup the script to be better POSIX sh compatible, and add -u to
the shebang.

### DIFF
--- a/plugins/ssl/ssl-certificate-expiry
+++ b/plugins/ssl/ssl-certificate-expiry
@@ -1,11 +1,11 @@
-#!/bin/sh
+#!/bin/sh -u
 # -*- sh -*-
 
 : << =cut
 
 =head1 NAME
 
-ssl-certificate-expiry - Plugin to monitor CERTificate expiration on multiple services and ports
+ssl-certificate-expiry - Plugin to monitor Certificate expiration on multiple services and ports
 
 =head1 CONFIGURATION
 
@@ -29,13 +29,26 @@ For example:
   ssl-certificate-expiry_192.0.2.42_636
   ssl-certificate-expiry_2001:0DB8::badc:0fee_485
 
-=head1 AUTHOR
+=head2 Cron setup
 
-Pactrick Domack (ssl_)
-Olivier Mehani (ssl-certificate-expiry)
+To avoid having to run the SSL checks during the munin-update, it is possible
+to run it from cron, and save a cachefile to be read during the update, This is
+particularly useful when checking a large number of certificates, or when some
+of the hosts are slow.
 
-Copyright (C) 2013 Patrick Domack <patrickdk@patrickdk.com>
-Copyright (C) 2017 Olivier Mehani <shtrom+munin@ssji.net>
+To do so, add a cron job running the plugin with cron as the argument:
+
+   */5 * * * * <user> /usr/sbin/munin-run/ssl-certificate-expiry cron
+
+<user> should be the user that has write permission to the MUNIN_PLUGSTATE.
+
+=head1 AUTHORS
+
+ * Pactrick Domack (ssl_)
+ * Olivier Mehani (ssl-certificate-expiry)
+
+ * Copyright (C) 2013 Patrick Domack <patrickdk@patrickdk.com>
+ * Copyright (C) 2017, 2019 Olivier Mehani <shtrom+munin@ssji.net>
 
 =head1 LICENSE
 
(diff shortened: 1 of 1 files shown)


sumpfralle:

This is quite a big amount of changes in a single commit :)

I see the following points that could be worth a discussion, if you think, that they could be useful.
* Currently most plugins based on shell (`/bin/sh`) follow the rules of `dash` dialect (as supported by `shellcheck`). Thus I would prefer to keep using `local`.
* The cron job based update sounds like a reasonable idea for spreading the load for a big number of certificates. I would suggest the following adjustments:
    * Maybe keep the cache file after printing instead of removing it. This would allow to use a less frequent update period. In order to avoid stale values from being used forever, you could compare the modify date of the cache file with the current time (`$(( $(date +%s) - $(date --reference "$file" +%s) ))`).
    * Maybe specify a longer time period as a default for the crontab entry in the documentation. This would make it clear for users, that you want to reduce the load effects. Certificates' expiry is usually relevant in days - thus it should be perfectly fine to update the values once per hour.
* Maybe add a comment to `mv` operation in the `cron` section, that mentions your goal of an atomic write operation. At least for me it was not obvious and I was tempted to suggest using a variable instead.

What do you think?

shtrom:
> Currently most plugins based on shell (/bin/sh) follow the rules of dash dialect (as supported by shellcheck). Thus I would prefer to keep using local.

Hum, fair enough... I don't run Linux on all my machines, and some distros don't have dash, so I'm not sure if targetting `dash` is a great idea. I'd rather not add the `local` with `sh` here, but I can move the shebang to `bash`.

> Maybe keep the cache file after printing instead of removing it. This would allow to use a less frequent update period.

Ha, I didn't think about this case. It's not a bad idea. However...

>  you could compare the modify date of the cache file with the current time ($(( $(date +%s) - $(date --reference "$file" +%s) ))).

I wanted to do this at first, but this is difficult to do cross-platform, for example [OpenBSD's `date` doesn't have the `--reference` argument](https://man.openbsd.org/date), and the output of `stat` (which was my initial thought) is completely different from the GNU version, so hard to parse.

Perhaps we can make it configurable? The idea of removing the file is to not keep stale info forever, but it does make sense with longer cron periods.

sumpfralle:
> > Currently most plugins based on shell (/bin/sh) follow the rules of dash dialect (as supported by shellcheck). Thus I would prefer to keep using local.
> 
> Hum, fair enough... I don't run Linux on all my machines, and some distros don't have dash, so I'm not sure if targetting `dash` is a great idea.

Do you have a specific example of a `/bin/sh` implementation, that fails with `local`? Even the minimal `ash` implementation in OpenWrt supports it, thus I would be surprised, if a recent distribution would ship something that supports only the minimal specification.

Many `/bin/sh`-based plugins in the munin and munin-contrib repositories use `local`. Currently I am not aware of issues reported for that.

> I'd rather not add the `local` with `sh` here, but I can move the shebang to `bash`.

Please do not use `bash` without a reason: this will exclude many embedded environment (e.g. OpenWrt).
In this specific case, the `local` was present in this plugin before, thus I would appreciate, if you leave it, unless you solve a problem with this change.


> > you could compare the modify date of the cache file with the current time ($(( $(date +%s) - $(date --reference "$file" +%s) ))).
> 
> I wanted to do this at first, but this is difficult to do cross-platform, for example [OpenBSD's `date` doesn't have the `--reference` argument](https://man.openbsd.org/date), and the output of `stat` (which was my initial thought) is completely different from the GNU version, so hard to parse.

Interesting!

How about `find "$filename" -mtime -1` (or running `find "$filename" -type f -mtime +1 -delete` just before)? Less precision, but probably enough for this use-case ...

shtrom:
> Do you have a specific example of a /bin/sh implementation, that fails with local? Even the minimal ash implementation in OpenWrt supports it, thus I would be surprised, if a recent distribution would ship something that supports only the minimal specification.

Yup, I tested in OpenBSD's sh, and `local` does pass through without an issue. It's weird, as I only removed it because `shellcheck` flagged it. I'll restore that.

> How about find "$filename" -mtime -1 (or running find "$filename" -type f -mtime +1 -delete just before)? Less precision, but probably enough for this use-case ...

Ha! That should work nicely!

shtrom:
Those fixups should address everything (and I'm even using a variable instead of a temporary file now!).

sumpfralle:
Great!

PR #996, opened 2019-07-22
Added Redis lag graph for slave0 node replication

tsaavik:


sumpfralle:
Looks good!

btw: maybe you would like to turn this plugin into a `multigraph` plugin? Then you will not need to manage individual symlinks for all the detail graphs ...

PR #999, opened 2019-07-26
Update ntp_packets

d0m84:
With Debian 10 the command "ntpq -c iostats -c sysstats" produces one empty line.
The additional if condition tackles this.

  File "/etc/munin/plugins/ntp_packets", line 91, in <module>
    stats[line.split(':')[0]] = int(line.split(':')[1])
IndexError: list index out of range

### DIFF
--- a/plugins/ntp/ntp_packets
+++ b/plugins/ntp/ntp_packets
@@ -86,6 +86,8 @@ stats = dict()
 stats_output = subprocess.check_output([cmd, '-c', 'iostats', '-c', 'sysstats'],
                                        universal_newlines=True).splitlines()
 
+# Split the cmd output into key/value pairs
+# Lines that can't be splitted into 2 individual elements by delimiter ':' will be skipped
 for line in stats_output:
     if len(line.split(':')) == 2:
         stats[line.split(':')[0]] = int(line.split(':')[1])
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thanks for improving this plugin!

Regarding the conditional:
* is the splitting not needed anymore with Debian Buster? (after your change it would not be executed anymore)
* could you please add a comment describing the necessity of this conditional? This will allow us to get rid of it in ten years :)

Thanks!

d0m84:
>     * is the splitting not needed anymore with Debian Buster? (after your change it would not be executed anymore)

It is still needed. The condition just makes sure that lines which can't be splitted into 2 parts by delimiter : will be skipped.

>     * could you please add a comment describing the necessity of this conditional? This will allow us to get rid of it in ten years :)

The current version doesn't work with ntpq version of Debian 10. It produces the mentioned Python exception:

```
File "/etc/munin/plugins/ntp_packets", line 91, in
stats[line.split(':')[0]] = int(line.split(':')[1])
IndexError: list index out of range
```

sumpfralle:
Thanks for your explanation!

Could you please add a comment to the code, that explains, under which circumstances the conditional is used (and necessary) and under which circumstances it works without the conditional?

I would like to keep this plugin maintainable by allowing future contributors to understand the current code. Thank you!

sumpfralle:
@d0m84: ping?

d0m84:
@sumpfralle done

sumpfralle:
Thank you!

PR #1003, opened 2019-07-28
[router/freebox] Use DERIVE for error correction subgraph

shtrom:
The error count keeps increasing to silly unreadably high values otherwise, that survive across router reboot. Using a DERIVE here makes it way me readable and useful.


sumpfralle:
Looks good - and thanks for keeping the example graph up to date!

PR #1005, opened 2019-08-09
Fix asterisk ami parsing

lelutin:
This patch set should make the asterisk multi-graph plugin work again on recent versions of Asterisk.

### DIFF
--- a/plugins/asterisk/asterisk
+++ b/plugins/asterisk/asterisk
@@ -77,30 +77,32 @@ use strict;
 use Munin::Plugin;
 use IO::Socket;
 
+# See the following and its subpages for change history in the AMI protocol:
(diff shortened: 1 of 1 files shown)


lelutin:
withtout the two additional commits that fix a series of typoes in "config" output, the plugin does not work at all (e.g. munin-update can't complete polling for that plugin)

sumpfralle:
Hehe - what a nice cleanup - thank you!

lelutin:
oops! I just spotted an issue with my PR: I left two lines of debugging that won't work. I'll submit the fix now to output those only if MUNIN_DEBUG is set.

lelutin:
ok, I've submitted #1006 for the fix

sumpfralle:
No problem - thank you for fixing up the issue!

PR #1006, opened 2019-08-10
asterisk: turn output into debug-only lines

lelutin:
I apparently forgot to clear out two lines of debugging output from the
script before submitting a pull request.

Those lines of output are super useful for debugging whenever things are
not working as expected (e.g. asterisk has changed its output). So it'll
be more productive to keep them around but only output them if
MUNIN_DEBUG is set.

This is a fixup for d66fef2 which was already merged.


sumpfralle:
Thank you!

PR #1007, opened 2019-08-10
Replace meetme with confbridge

lelutin:
This adds a new graph to the multigraph so that we can now have statistics about the ConfBridge application. Since the MeetMe application has been deprecated for a long while, the graph for it is now also disabled by default, but can be brought back with a configuration option.

If the ConfBridge graph is not wanted by users, it can also be disabled by a configuration option.

### DIFF
--- a/plugins/asterisk/asterisk
+++ b/plugins/asterisk/asterisk
@@ -131,8 +131,8 @@ my $confbridge_enabled = $ENV{'confbridge'} || '1';
 
 my $line, my $error;
 my $socket = new IO::Socket::INET(PeerAddr => $peeraddr,
-				  PeerPort => $peerport,
-				  Proto => 'tcp')
+                                  PeerPort => $peerport,
+                                  Proto => 'tcp')
   or $error = "Could not create socket: $!";
 
 if ( $socket ) {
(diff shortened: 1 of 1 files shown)


plugins/asterisk/asterisk
```diff
@@ -27,13 +30,15 @@ This plugin will produce multiple graphs showing:
...
-  env.codecsx  - List of codec IDs (hexadecimal values)
-  env.codecs   - List of codecs names, matching codecsx order
+  env.host       - hostname to connect to
+  env.port       - port number to connect to
+  env.username   - username used for authentication
+  env.secret     - secret used for authentication
+  env.channels   - The channel types to look for
+  env.codecsx    - List of codec IDs (hexadecimal values)
+  env.codecs     - List of codecs names, matching codecsx order
+  env.meetme     - Set to 1 to enable graphs for the MeetMe application
```
sumpfralle: Please use `enable_meetme` or `meetme_enabled` instead. This clarifies the content of the variable.
Same for `confbridge` below.

lelutin:
thanks for your helpful review!

I somehow couldn't find the mixture of tabs and spaces in the spots your pointed to in the review, but since there were mixtures of tabs and spaces elsewhere in the file, I cleared them all out to make sure that there is no confusion.

sumpfralle:
> I somehow couldn't find the mixture of tabs and spaces in the spots your pointed to in the review, [..]

uh - so it was my misunderstanding of github's visualization of the diff: it probably did not refer to a whitespace issue in the line itself, but was related to other lines around.


> [..] but since there were mixtures of tabs and spaces elsewhere in the file, I cleared them all out to make sure that there is no confusion.

Perfect!

Thanks for this thorough cleanup and improvement!

PR #1011, opened 2019-08-14
Convert "w1_" into a plugin for multiple sensors

sumpfralle:
All temperatures are displayed in a single graph.
This should provide more convenience than the previous "suggest"-based
wildcard plugin.

Additionally the label of each sensor can be overridden.


sumpfralle:
Merged manually.

PR #1012, opened 2019-08-16
btrfs_subvol_usage: fix "btrfs qgroup show" output format

rantal:
Switch "btrfs qgroup show" for raw output format, because by default it's human readable form and can't be read later in the script.


sumpfralle:
great!

PR #1016, opened 2019-09-02
Fix typos in certificate_file_expiry

RenWal:
Typos broke the sample configuration


sumpfralle:
Good catch - thank you!

PR #1017, opened 2019-09-06
Add Missing Bracket

ceagan:
Adds the missing bracket `[` to ensure that brackets show up after groups when hosts are grouped together.


sumpfralle:
Looks good - thank you!

PR #1018, opened 2019-09-14
Add support for StartTLS in ssl-certificate-expiry

ruliane:
Added support for StartTLS in _ssl-certificate-expiry_
Use 
`env.services foo.example.net_25_smtp`
to enable StartTLS on a SMTP server.

Please review this PR before commit.

### DIFF
--- a/plugins/ssl/ssl-certificate-expiry
+++ b/plugins/ssl/ssl-certificate-expiry
@@ -11,7 +11,10 @@ ssl-certificate-expiry - Plugin to monitor Certificate expiration on multiple se
 =head1 CONFIGURATION
 
   [ssl-certificate-expiry]
-    env.services www.service.tld blah.example.net_PORT
+    env.services www.service.tld blah.example.net_PORT foo.example.net_PORT_STARTTLS
+
+PORT is the TCP port number
+STARTTLS is passed to openssl as "-starttls" argument. Useful for services like SMTP or IMAP implementing StartTLS.
 
 To set warning and critical levels do like this:
 
@@ -29,6 +32,7 @@ For example:
   ssl-certificate-expiry_www.example.org_443
   ssl-certificate-expiry_192.0.2.42_636
   ssl-certificate-expiry_2001:0DB8::badc:0fee_485
+  ssl-certificate-expiry_mail.example.net_25_smtp
 
 =head2 Cron setup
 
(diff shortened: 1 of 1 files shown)


plugins/ssl/ssl-certificate-expiry
```diff
@@ -11,7 +11,10 @@ ssl-certificate-expiry - Plugin to monitor Certificate expiration on multiple se
 =head1 CONFIGURATION
 
   [ssl-certificate-expiry]
-    env.services www.service.tld blah.example.net_PORT
+    env.services www.service.tld blah.example.net_PORT foo.example.net_PORT_STARTTLS
+
+PORT is the TCP port number
+STARTTLS is passed to openssl as "-starttls" argument. Useful for services like SMTP or IMAP implementing StartTLS.
```
sumpfralle: Please add an example here (e.g. "imap or ldap"). I guess, this would make the usage a bit more obvious.
(`STARTTLS` as a parameter name is quite unspecific - but I fail to imagine something better).

plugins/ssl/ssl-certificate-expiry
```diff
@@ -102,26 +106,34 @@ parse_valid_days_from_certificate() {
...
     local host="$1"
     local port="$2"
+    local starttls="$3"
 
     # Wrap IPv6 addresses in square brackets
     echo "$host" | grep -q ':' && host="[$host]"
 
-    echo "" | openssl s_client -CApath /etc/ssl/certs \
+    if [ -n "$starttls" ]; then
+        echo "" | openssl s_client -CApath /etc/ssl/certs \
```
sumpfralle: I would suggest to add a parameter to the openssl call instead of duplicating the line.

plugins/ssl/ssl-certificate-expiry
ruliane: An example is below.

plugins/ssl/ssl-certificate-expiry
ruliane: openssl "-starttls" argument cannot be empty, so the call MUST be different.

plugins/ssl/ssl-certificate-expiry
sumpfralle: Yes, but just from my personal first attempt of reading this part, I failed to understand the meaning of the `STARTTLS` placeholder (since I do not know the manpage of `s_client` by heart). Just a suggestion, anyway ...

plugins/ssl/ssl-certificate-expiry
sumpfralle: I was a bit too concise, I guess :)

I imagined something like this:
```shell
local s_client_args=
[ -n "$starttls" ] && s_client_args="-starttls $starttls"
```
Afterwards I would add the unquoted `$s_client_args` parameter to the `s_client` call. This approach requires a `shellcheck` warning needs to be silenced (`# shellcheck disable=SC2086`).

What do you think?

ruliane: Good idea, though I don't know what shellcheck is...
I updated the PR with your improvement.

sumpfralle:
Great - thank you!

PR #1020, opened 2019-09-29
shtrom's plugins cleanup

shtrom:
Cleanup some of my plugins (network/upnpc_, isp/internode) a bit.
* Stricter shell syntax
* Add support for host_name option  to allow displaying graph for another host (e.g., the uplink router)
* Add multigraph support for upnpc (backwards compatible)
* Deprecate upnpc_bitrate (backward compatible), and optionally show on upnpc_traffic instead


shtrom:
Just found a bug in `ssl/ssl-certificate-expiry`, sneaking the fix here, too.

plugins/isp/internode_usage
```diff
@@ -68,46 +70,74 @@ fi
...
+		| sed 's/<\([^>]*\)>\([^<]*\)<[^>]*>/\2/;s^N/A^U^'
+	trap - EXIT
 }
 
 xpath_extract_attribute() {
 	xpath="$1"
-	xpath -q -n -e "${xpath}" | \
-		sed 's/.*="\([^"]\+\)".*/\1/'
+	# shellcheck disable=SC2064
+	trap "error extracting attribute at ${xpath} >&2" EXIT
```
sumpfralle: How about using the `curl ... || echo >&2 "..."` approach instead?
(I think, it is easier to read here)

Another detail: I think, you omitted `echo` in the `trap`.

plugins/network/upnpc_
```diff
@@ -37,21 +39,38 @@ SPDX-License-Identifier: GPL-2.0-or-later
 
 =cut
 
+if [ "${MUNIN_DEBUG:-0}" = 1 ]; then
+    set -x
+fi
+
+if ! command -v upnpc >/dev/null; then
```
sumpfralle: This test should be part of the `autoconf` handling.
At the moment a call for `autoconf` results in an invalid response string (`upnpc not found ...`) instead of `no (upnpc not found)`.

Probably you should move the data retrieval (`DATA=$(upnpc -s)`) down to the `config` and `fetch` operation in order to avoid the complexities auf combining `autoconf` and central data collection?

plugins/isp/internode_usage
```diff
@@ -221,14 +221,14 @@ graph_data() {
...
-			echo "hourly_rate.value ${SERVICE_USAGE:-U}"
+			echo "hourly_rate.value ${CURRENT_TIMESTAMP}:${SERVICE_USAGE:-U}"
 			;;
 		.daily)
 			echo "daily_rate.value ${DAILY_TIMESTAMP}:${DAILY_USAGE:-U}"
 			;;
 		*)
-			echo "usage.value ${SERVICE_USAGE:-U}"
-			echo "ideal.value ${IDEAL_USAGE:-U}"
+			echo "usage.value ${CURRENT_TIMESTAMP}:${SERVICE_USAGE:-U}"
```
sumpfralle: Is it really necessary to add the timestamp here? This sounds like a bug in the spool processing to me.

plugins/network/upnpc_
```diff
@@ -20,7 +20,11 @@ Linux systems with upnpc installed (miniupnpc package).
 
 =head1 CONFIGURATION
 
-None needed.
+If you do not want to show the link maximum bitrates, add the following
+plugin-configuration:
+
+  [upnpc*]
```
sumpfralle: Maybe add the final underscore in order to keep it focused?

plugins/network/upnpc_
```diff
@@ -144,6 +147,14 @@ down.graph no
 up.negative down
 EOF
 			;;
+		"multi"|"upnpc")
```
sumpfralle: I think, it would be better to normalize the "MODE" value at the beginning (i.e. `[ "$MODE" = "upnpc" ] && MODE="multi"`). Otherwise the `"multi"|"upnpc"` conditional can be easily misunderstood.

plugins/network/upnpc_
```diff
@@ -144,6 +147,14 @@ down.graph no
...
+			echo "multigraph ${PLUGIN_NAME}"
+			config "traffic"
```
sumpfralle: Is the `traffic` really suitable as a *top level summary* of the other graphs?
Otherwise you may just want to graph all of them at the same level? (just replace the dot with an underscore)

sumpfralle:
Thanks a lot for this good amount of change proposals!

I commented a few of them in detail ...

plugins/isp/internode_usage
shtrom: It is, but there's not real way around it \:

plugins/network/upnpc_
shtrom: No, as the plugin also supports `upnpc` as a shorthand for `upnpc_multi`

plugins/network/upnpc_
shtrom: I think it is. It's a matter of preference though.

That said, we need at one top-level graph. We can't just have subpages.

plugins/isp/internode_usage
shtrom: Good catch on the missing `echo`.

I can't remember why I didn't just do `cmd || echo >&2`. I'll need to rethink that 😬

shtrom: Ah! Yes! That's it, you're right. I want to be able to output the error message, but to let the caller decide whether this is an error that can be tolerated (with `|| true`, which is used a few times).

So, yeah, I think we need to keep it like that.

shtrom:
@sumpfralle Ready for another look. I've fixed what I thought was fixable, and left comments explaining why I didn't think the rest was.

Let me know what you think.

shtrom:
I just snuck in an additional fix for the fresh-backups plugins, too, as the limits where calculated incorrectly.

plugins/network/upnpc_
sumpfralle: Good point!

plugins/isp/internode_usage
sumpfralle: > Ah! Yes! That's it, you're right. [..]

A good example of code with surprising side effects :)

I would appreciate a more obvious approach, but I could live with this one, too.

plugins/isp/internode_usage
sumpfralle: If it is a bug: where is the ticket? (I would like to solve it instead of fostering bad examples in plugin code).

plugins/network/upnpc_
sumpfralle: > That said, we need at one top-level graph. We can't just have subpages.

I wanted to propose to use underscores instead of dots in the multigraph names. In this case there would be no hierarchy and thus just a flat list of graphs at the top level.

Feel free to pick the most suitable approach for this set of information.

sumpfralle:
> Let me know what you think.

Thank you! I did not notice any new details, that I would like to discuss.

> I just snuck in an additional fix for the fresh-backups plugin, too, as the limits were calculated incorrectly.

In general it is fine. In a perfect world I would prefer, if you could separate "rename" and "change" into two commits. At the moment the change within the renamed plugin is hard to spot.

What do you think: are there still things, you would like to improve before I merge the commits?

plugins/network/upnpc_
shtrom: Ah right, yeah. I never thought about it, actually.

That said, one of the reason I (ab)use multigraphs is to not have too many graphs on the front page, while still being able to drill down to more details if I want to. In this respect, I find the traffic most useful as a top-level graph, while the others are less so.

I think I'm going to leave it as is (:

plugins/isp/internode_usage
shtrom: Actually, maybe it's not a bug as such, but a corner case.

This particular plugin outputs data with two different dates: right now, and more or less a day before. Munin is not used to it, and assumes that all the data from a plugin comes from the same time, which is not the case here. Not stating the timestamp explicitely here leads munin to assume that it's the same as the delayed one, which is often lower than the last received statement, which leads Munin to discard fresh values from the current time.

I think it's more of a weird interaction between Munin's expectations, and the data from the API this plugin queries.

plugins/isp/internode_usage
shtrom: Ah! I think I found a better way. What about now?

shtrom:
Couple more fixups. I split the move/change commit from `fresh-backup`, and added caching to the `internode_usage` plugin, as their API is a bit flakey.

I think I also found a `trap`-less way to report errors while letting the caller decide what to do about it.

plugins/isp/internode_usage
```diff
@@ -104,6 +104,25 @@ fetch() {
 		    && false)
 }
 
+get_cached_api_xml() {
+	API_XML=''
+	API_XML_CACHE="${MUNIN_PLUGSTATE}/$(basename "${0}").cache"
```
sumpfralle: I would suggest to move this up closer to the top of the plugin code. This will make it more visible, that cached data is used.

plugins/isp/internode_usage
```diff
@@ -104,6 +104,25 @@ fetch() {
 		    && false)
 }
 
+get_cached_api_xml() {
+	API_XML=''
```
sumpfralle: Probably this variable (especially with such a generic name) deserves to be `local`?

plugins/isp/internode_usage
```diff
@@ -83,31 +83,25 @@ fi
...
 xpath_extract_attribute() {
 	xpath="$1"
-	# use trap to output an error message but let the caller get the status
-	# shellcheck disable=SC2064
-	trap "echo error extracting attribute at ${xpath} >&2" EXIT
-	xpath -q -n -e "${xpath}" \
-		| sed 's/.*="\([^"]\+\)".*/\1/'
-	trap - EXIT
+	(xpath -q -n -e "${xpath}" \
+		| sed 's/.*="\([^"]\+\)".*/\1/') \
```
sumpfralle: I think, here (and above in the other function) you merely test, wether `sed` failed before.

I guess, you will need to store the output of `xpath` in a variable :)

plugins/isp/internode_usage
```diff
@@ -83,31 +83,25 @@ fi
...
 fetch() {
-	# shellcheck disable=SC2064 disable=SC2154
-	# use trap to output an error message but let the caller get the status
-	trap "echo error fetching ${*} for user ${internode_api_login} >&2" EXIT
-	# shellcheck disable=SC2086 disable=SC2154
-	curl -u "${internode_api_login}:${internode_api_password}" -f ${CURL_ARGS} "$@"
-	trap - EXIT
+	# shellcheck disable=SC2154
+	curl -u "${internode_api_login}:${internode_api_password}" -f ${CURL_ARGS} "$@" \
+		|| (echo "error fetching ${*} for user ${internode_api_login}" >&2 \
```
sumpfralle: Just a minor nitpick: `{ ...; }` (instead of `(...)`) would be more efficient here, since it does not require a fork.

sumpfralle:
It looks good to me!

Feel free to respond to my last comments, if they feel useful for you.

As soon as you rebase your `fixup` commits, I will merge your set of changes.
Thank you!

plugins/isp/internode_usage
shtrom: Hum, I prefer the code organised like this, with helper functions together, and data-collecting functions together.

However, I added some details about the caching mechanism in the documentation.

plugins/isp/internode_usage
shtrom: Heh, TIL. I never realised I could use `{ ... ; }`.

The difference is notable
```
$ time sh -c 'for i in `seq 10000`; do true && (false); done'
sh -c 'for i in `seq 10000`; do true && (false); done'  2.66s user 2.47s system 104% cpu 4.886 total
$ time sh -c 'for i in `seq 10000`; do true && { false; }; done'
sh -c 'for i in `seq 10000`; do true && { false; }; done'  0.02s user 0.00s system 98% cpu 0.019 total
```

plugins/isp/internode_usage
shtrom: Ah, you're right...

This is annoying, as I was trying to keep it short and sweet \:

shtrom:
Ok, fixed and squashed.

plugins/isp/internode_usage
sumpfralle: I only meant the definition of the variable (`API_XML_CACHE`), as this filename is part of the interface of the plugin. But anyway: personal preferences vary (which is good) ... :)

plugins/isp/internode_usage
sumpfralle: Nice benchmark!

But to be fair: executing non-builtin statements (e.g. calling `curl`) is quite common in a script. Thus the difference in reality is probably not as noticeable as the above numbers may imply.

sumpfralle:
Thank you for contributing this great amount of neat improvements!

plugins/isp/internode_usage
shtrom: Oh right, sorry. Yeah, it was only one line up, but I see your point. It will happen next time I have improvements to add!

PR #1022, opened 2019-10-21
[hostdenied/fresh-backup] Reverse order of series so new entries are …

shtrom:
…on top


plugins/backup/fresh-backups
```diff
@@ -76,19 +76,22 @@ graph_title Fresh backups
...
-	cat << EOF
-age${AGE}.label older than $((AGE*24))h
-age${AGE}.draw STACK
+for AGE in $(seq "${LIFETIME}" -1 0); do
+	if [ "${AGE}" = 0 ]; then
+		echo "age${AGE}.label today"
+	else
+		echo "age${AGE}.label older than $((AGE*24))h"
+	fi
+	if [ "${AGE}" = "${LIFETIME}" ]; then
```
sumpfralle: You can simplify this with using `AREASTACK` for all fields.

plugins/backup/fresh-backups
```diff
@@ -76,19 +76,22 @@ graph_title Fresh backups
...
+		echo "age${AGE}.draw AREA"
+	else
+		echo "age${AGE}.draw STACK"
+	fi
+		cat << EOF
```
sumpfralle: The indentation can be reduced, I guess.

plugins/backup/fresh-backups
```diff
@@ -76,19 +76,22 @@ graph_title Fresh backups
...
 age${AGE}.colour ${COLOUR[$AGE]}
-age${AGE}.warning :
-age${AGE}.critical :
+age${AGE}.warning
```
sumpfralle: Do these two lines serve a purpose?
How about using `print_warning` instead?

plugins/ssh/hostdenied
```diff
@@ -44,35 +44,38 @@
...
-NAME="$(basename $0)"                                        # component of naming temporary files
+NAME="$(basename "$0")"                                        # component of naming temporary files
 STATEFILE="$MUNIN_PLUGSTATE/$NAME.state"
 COLOUR=(FF0000 DA0024 B60048 91006D 6D0091 4800B6 2400DA 0000FF)  # hot to cold colours
 # ------------------------------------------------------------------------------------------------------
 
 run_autoconf()  {
-        RUN="no"
-        which grep denyhosts basename > /dev/null && RUN="yes"             # only run when grep and denyhost are present
+	RUN="no (denyhosts not found)"
```
sumpfralle: The indentation is mixed.

plugins/ssh/hostdenied
```diff
@@ -44,35 +44,38 @@
...
-age${AGE}.draw STACK
-age${AGE}.colour ${COLOUR[$AGE]}
 EOF
+for AGE in {7..0}; do
+	if [ "${AGE}" = 0 ]; then
+		echo "age${AGE}.label added last 24h"
+	else
+		echo "age${AGE}.label older than $((AGE*24))h"
+	fi
+	if [ "${AGE}" = 7 ]; then
```
sumpfralle: Maybe `AREASTACK`?

plugins/backup/fresh-backups
shtrom: In my early experiment with this, the warning and critical values seemed to be inherited from those for `freshcount` when plotted. This forces no limit to be checked on those series.

`plugin.sh` is not safe to use with `-u` as it doesn't default undefined values (such as warnings and critical limits) to sane values.

shtrom:
@sumpfralle done.

plugins/backup/fresh-backups
sumpfralle: Inheritance: interesting. I am not sure, that this is the way it should work ...

`set -u`: indeed, this needs to be fixed!

sumpfralle:
Thank you!

PR #1028, opened 2019-11-25
Monitor RAM and flash disk usage in Mikrotik routers

alejandroscf:
Two scripts to monitor RAM and flash disk usage in Mikrotik routers via SNMP protocol

### DIFF
--- a/plugins/router/snmp__mikrotik_flash
+++ b/plugins/router/snmp__mikrotik_flash
@@ -1,27 +1,40 @@
 #!/usr/bin/perl -w
-#
-# Copyright (C) 2019 Alejandro Suarez (teconecta.es)
-#
-# Munin plugin to monitor Mikrotik routers Falsh disk utilization.
-# Based on snmp__if_  plugin.
-#
-# This program is free software; you can redistribute it and/or
-# modify it under the terms of the GNU General Public License
-# as published by the Free Software Foundation; version 2 dated June,
-# 1991.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program; if not, write to the Free Software
-# Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 
+=head1 NAME
+
+snmp__mikrotik_flash - Munin plugin to monitor Mikrotik routers Falsh disk utilization.
+
+=head1 APPLICABLE SYSTEMS
+
+Mikrotik Routers
+
+=head1 CONFIGURATION
+
+As a rule SNMP plugins need site specific configuration.  The default
+configuration (shown here) will only work on insecure sites/devices.
+   [snmp_*]
+        env.version 2
+        env.community public
+
+=head1 MAGIC MARKERS
 
 #%# family=snmpauto
 #%# capabilities=snmpconf
+
+=head1 BUGS
+
+None known.
+
+=head1 AUTHOR
+
+Copyright (C) 2020 Alejandro Suarez (teconecta.es)
+Based on snmp__if_  plugin.
+
+=head1 LICENSE
+
+GPLv2.
+
+=cut
 
 use strict;
 use Net::SNMP;
--- a/plugins/router/snmp__mikrotik_flash
+++ b/plugins/router/snmp__mikrotik_flash
@@ -2,7 +2,7 @@
 #
 # Copyright (C) 2019 Alejandro Suarez (teconecta.es)
 #
-# Munin plugin to monitor Mikrotik routers RAM utilization.
+# Munin plugin to monitor Mikrotik routers Falsh disk utilization.
 # Based on snmp__if_  plugin.
 #
 # This program is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


plugins/router/snmp__mikrotik_flash
```diff
@@ -0,0 +1,106 @@
...
+
+
+if ($ARGV[0] and $ARGV[0] eq "config")
+{
+    print "host_name $host\n";
+    if (!defined ($response = $session->get_request($sysRAMTotal)))
+    {
+        die "Croaking: " . $session->error();
+    }
+#    print "graph_args --base 1000 -r --lower-limit 0 --upper-limit 100\n";
```
sumpfralle: Please remove commented code.

plugins/router/snmp__mikrotik_flash
```diff
@@ -0,0 +1,106 @@
...
+    print "sysRAMTotal.draw AREA\n";
+    print "sysRAMUsage.label Used Memory\n";
+    print "sysRAMUsage.draw AREA\n";
+    exit 0;
+}
+
+
+if (defined ($response = $session->get_request(-varbindlist => [$sysRAMUsage, $sysRAMTotal])))
+{
+        print "sysRAMUsage.value ", $response->{$sysRAMUsage}*1024, "\n";
```
sumpfralle: Is this a conversion from bytes to kilobytes? Usually SI base units are preferable in order to let munin/rrd print proper units for higher values.

Another detail: is `RAM` the proper name here? Shouldn't it be about the flash storage?

plugins/router/snmp__mikrotik_flash
```diff
@@ -0,0 +1,106 @@
...
+        print "sysRAMTotal.value ", $response->{$sysRAMTotal}*1024, "\n";
+}
+else
+{
+        print "sysRAMUsage.value U\n";
```
sumpfralle: I guess, you should also emit `sysRAMTotal` as `U`?

plugins/router/snmp__mikrotik_flash
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/perl -w
+#
```
sumpfralle: Please use a perldoc header here - this allows automatic man page generation and better visualization in the plugin gallery.

plugins/router/snmp__mikrotik_ram
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/perl -w
+#
```
sumpfralle: Perldoc header, please.

plugins/router/snmp__mikrotik_ram
```diff
@@ -0,0 +1,106 @@
...
+
+
+if ($ARGV[0] and $ARGV[0] eq "config")
+{
+    print "host_name $host\n";
+    if (!defined ($response = $session->get_request($sysRAMTotal)))
+    {
+        die "Croaking: " . $session->error();
+    }
+#    print "graph_args --base 1000 -r --lower-limit 0 --upper-limit 100\n";
```
sumpfralle: no commented code, please.

plugins/router/snmp__mikrotik_ram
```diff
@@ -0,0 +1,106 @@
...
+    print "sysRAMTotal.draw AREA\n";
+    print "sysRAMUsage.label Used Memory\n";
+    print "sysRAMUsage.draw AREA\n";
+    exit 0;
+}
+
+
+if (defined ($response = $session->get_request(-varbindlist => [$sysRAMUsage, $sysRAMTotal])))
+{
+        print "sysRAMUsage.value ", $response->{$sysRAMUsage}*1024, "\n";
```
sumpfralle: Same as above: is the conversion necessary?

plugins/router/snmp__mikrotik_ram
```diff
@@ -0,0 +1,106 @@
...
+        print "sysRAMTotal.value ", $response->{$sysRAMTotal}*1024, "\n";
+}
+else
+{
+        print "sysRAMUsage.value U\n";
```
sumpfralle: Same here: missing `total`?

plugins/router/snmp__mikrotik_flash
alejandroscf: No, the conversion is from kilobytes to bytes. SNMP is returning it in kilobytes so I convert it to bytes in order to had it in SI baseunits. 

Yes, that should be "FLASH" or something like that.

plugins/router/snmp__mikrotik_flash
alejandroscf: Sure

plugins/router/snmp__mikrotik_flash
alejandroscf: Yes, it should.

plugins/router/snmp__mikrotik_flash
alejandroscf: I'm not sure how to use perldoc. It's like the headers in other plugins?

alejandroscf:
Thank for you comments. I'll try to fix that.
I made it in two different files so you can choose what to monitor in each situation. For example I use this for work and only monitor flash usage in some routers, but I want to keep an eye in RAM usage in all of them. Also the snmp_zyxel_zywall__ plugins also had it in different files.

plugins/router/snmp__mikrotik_flash
sumpfralle: Yes. Just look for `=cut` in other plugins.

sumpfralle:
> For example I use this for work and only monitor flash usage in some routers, but I want to keep an eye in RAM usage in all of them.

Specifc graphs can also be excluded by configuration. Maybe this would be a better solution?

> Also the snmp_zyxel_zywall__ plugins also had it in different files.

Yes, but I would like to prevent bad examples (code copies) from spreading :)

Regarding: the SNMP OIDs: do you think, they are applicable for other SNMP devices? Thus maybe the plugin name could be more generic?

sumpfralle:
Ping?

plugins/router/snmp__mikrotik_flash
alejandroscf: I've added the header, I hope it's ok.

alejandroscf:
I fixed the misspelling and the Perl Doc header. I'm sorry but I have no time to solve the multigraph thing, maybe we could merge now?

sumpfralle:
Thank you!

I really do not like to introduce code duplication into the repository. Thus I took the liberty to migrate the plugins to the multigraph approach myself: #1047.

Please report back whether the unified plugin works for you.

alejandroscf:
Thank you very much! 

I've just tested it and I left you a suggestion in the other PR so I'm closing this one.

PR #1029, opened 2019-12-08
add plugin for Technicolor TC8715D cable modem stats

kenyon:


kenyon:
Eh, I formatted this code with [black](https://github.com/psf/black). E203 is not PEP8 compliant, so that can be ignored. The two line length issues should be ignored because IMO breaking those lines makes it harder to read.

plugins/router/technicolor_tc8715d
```diff
@@ -0,0 +1,195 @@
+#!/usr/bin/env python3
+
+# Munin plugin for graphing statistics from Technicolor TC8715D cable modems.
```
sumpfralle: A perldoc header would be nice.

plugins/router/technicolor_tc8715d
```diff
@@ -0,0 +1,195 @@
...
+            + 8 * self.downstream_channels
+            + 7 * self.upstream_channels : 21
+            + 9 * self.downstream_channels
+            + 7 * self.upstream_channels
+        ]
+
+
+if len(sys.argv) != 2 or sys.argv[1] != "config":
+    print(
+        "Error: plugin designed for the dirtyconfig protocol, must be run with the config argument"
```
sumpfralle: I guess, this is the first plugin, that *requires* the use of the dirtyconfig capability.

@steveschnepp: what do you think - is this a suitable for plugins?

plugins/router/technicolor_tc8715d
```diff
@@ -0,0 +1,195 @@
...
+    )
+    sys.exit(1)
+
+parser = TechnicolorHTMLParser()
+
+for line in urllib.request.urlopen("http://192.168.100.1/vendor_network.asp"):
```
sumpfralle: It would be great, if the IP would be configurable.

plugins/router/technicolor_tc8715d
kenyon: It's certainly not the first plugin, because one of my other cable modem monitoring plugins is the same: https://github.com/munin-monitoring/contrib/blob/master/plugins/router/motorola_sb6141

plugins/router/technicolor_tc8715d
kenyon: It doesn't need to be configurable because the cable modem is always at this special address. It cannot be changed.

plugins/router/technicolor_tc8715d
kenyon: Done.

sumpfralle:
Thank you!

PR #1030, opened 2019-12-08
multi_tcp_ping: enable service_check by default

hannibal218bc:
By default, Net::Ping will report a good attempt even when the remote end returns "connection refused". By enabling service_check, the connection must have been established.
See https://perldoc.perl.org/Net/Ping.html#Functions

### DIFF
--- a/plugins/ping/multi_tcp_ping
+++ b/plugins/ping/multi_tcp_ping
@@ -105,6 +105,7 @@ sub ping_host {
 
     $p=Net::Ping->new("tcp", $defaults{timeout});
     $p->hires();
+    $p->service_check(1);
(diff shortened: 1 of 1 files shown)


sumpfralle:

Since I do not know the details of this module: what exactly would change now?

hannibal218bc:
At the moment, any response to the SYN packet is valid and the timing used as measurement.
e.g. if a firewall between probe and target rejects the connection with a RSET packet, the time to receive this RSET will be measured despite the fact that no connection has been established.

With service_check enabled, only a successful established connection will report a timing. If the connection is rejected, the same result will be returned as when the target does not respond in any way.

sumpfralle:
Thank you for your explanation!

PR #1032, opened 2019-12-18
allow database name with dot

Lokutos:


plugins/mysql/mysql_size_all
```diff
@@ -111,6 +111,7 @@ foreach my $db (getDBList()) {
 
     $total_size = $datas+$indexes;
   }
+  $db =~ s/\./-/g;
```
sumpfralle: The dash is not an allowed character, too. Maybe it works, but by specification it is forbidden. It surely causes problems if used in `cdef` evaluation.

Instead you should do the following:
* add `use Munin::Plugin;` at the top of the script
* replace all illegal characters: `$db = clean_fieldname($db);`

(I hope, I guessed the purpose of your change correctly)

Lokutos: Hi, Thanks for reply.
Didn't insert the dash but i will check and change the lines from the original developer too.

sumpfralle: Indeed I meant `hyphen` instead of `dash`. Thanks for fixing this issue!

sumpfralle:
Did you already take a look? Maybe introduce `clean_fieldname` instead of manual repair?

sumpfralle:
@Lokutos: ping?

sumpfralle:
I created a separate commit (e641c8028469e940578c460060a23a9eeb27657e) with changes similar to the ones suggested by you.
Thanks!

PR #1033, opened 2019-12-19
Fix state saving in loggrepx_ plugin

lucc:
The pattern under which the data was saved was different than the pattern under the data was loaded on the next run. The reason being an undefined variable that evaluates to the empty string. The result was that the number of matching lines reported was for the whole logfile instead of the time since the last run.


sumpfralle:
Thank you!

PR #1034, opened 2019-12-31
create new plugin for graphing PuppetDB's usage of JVM Heapspace

lelutin:
PuppetDB requires a pretty high amount of memory allocated to it. If you
don't give enough memory space to the JVM, PuppetDB might crash when it
has fully used the available max memory. This plugin should be useful
for better tuning up your instance(s) of PuppetDB.

Currently, it's only compatible with PuppetDB 6.x since it's requesting
data from the API as described for that version:

https://puppet.com/docs/puppetdb/6.0/api/metrics/v1/mbeans.html#jvm-metrics

To add support for more versions, one would need to consult the same
documentation for different versions and to adapt URLs used and the code
that's obtaining the data from JSON if the layout of the responses is
different.


plugins/puppetdb/puppetdb
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/python3
+#
+# This code is licensed under GPLv3+
```
sumpfralle: Please use a perldoc formatted documentation instead. This will render the plugin nice in the gallery and for `munin-get`.

plugins/puppetdb/puppetdb
```diff
@@ -0,0 +1,122 @@
...
+#  * pdburl -- Set the URL to your PuppetDB instance. This url should point to
+#              the mbeans endpoint. By default this has a value of
+#              http://localhost:8080/metrics/v1/mbeans
+#
+#  * timeout -- Time in seconds (int) to wait for a result when querying the REST
+#               API. By default, wait for 2 seconds
+#
+import os
+import sys
+import requests
```
sumpfralle: Please consider using python's `urllib` and `json` module instead, since they are available in every python installation.
Your GET request looks simple enough for an easy switch.
(personally I consider `requests` to be useful only for interactive usage - see [requests#4842](https://github.com/psf/requests/issues/4842) for an example of non-deterministic behavior depending on locally installed modules)

plugins/puppetdb/puppetdb
```diff
@@ -0,0 +1,122 @@
...
+
```
sumpfralle: Just guessing: did you check, whether this plugin follows PEP8? (e.g. run `python3 -m flake8 FILE`)
I think, there should be two empty lines after the imports.

plugins/puppetdb/puppetdb
```diff
@@ -0,0 +1,122 @@
...
+def config():
+    """Print all graph configuration for munin."""
+    print("graph_title PuppetDB JVM Memory usage")
+    print("graph_args --base 1024") # TODO not sure about unit
+    print("graph_vlabel Bytes")
+    print("graph_info This graph shows how much memory from the JVM Heapspace is being used by PuppetDB")
+    # This is super weird, but currently puppet graphs are in this
+    # category so let's place puppetdb graphs in the same category
+    print("graph_category devel")
```
sumpfralle: I think, you should use the proper category here. Feel free to correct the category of the other `puppet` plugin.

plugins/puppetdb/puppetdb
```diff
@@ -0,0 +1,122 @@
...
+
+
+def fetch_field_values(mbeans_url, timeout):
+    """Get values from PuppetDB and print them out."""
+    memory_url = f"{mbeans_url}/java.lang:type=Memory"
+
+    try:
+        mem_req = rest_request(memory_url, timeout)
+    except Exception as e:
+        print(f"HTTP Request did not complete successfully: {e}")
```
sumpfralle: I think, you should write to `file=sys.stderr` here.

plugins/puppetdb/puppetdb
duritong: `urllib` on the other hand does not become trivial, as soon as you want to add certificate verification and client cert & key, which is what you need once the puppetdb is only available through SSL. Using `requests, this stays really simple.

sumpfralle: Indeed the certificate validation was problematic in older versions. But since Python 3.4 (see [release notes](https://docs.python.org/3/whatsnew/3.4.html#pep-476-enabling-certificate-verification-by-default-for-stdlib-http-clients)) certificates are verified by default (no additional arguments are required). Since you are using Python's new format strings, you may assume, that a Python version with proper certificate handling is in use.
The users of this munin plugin would surely appreciate, that the plugin works out of the box (without installing an additional python package).
Anyway: feel free to decide, whether you want to stick to `requests` or switch to `urllib`.

btw.: do you think, that it is reasonable to assume that the puppeddb API is configured with a valid certificate (in case of HTTPS)? Otherwise you may want to introduce a configuration setting for disabling certificate checks.

plugins/puppetdb/puppetdb
lelutin: I couldn't find any examples of this in other python plugins in the munin-contrib repository. can you point me to an example of what this would look like?

lelutin: ok nevermind this question. I found an example in the solr plugin. I've reformatted the top comment to make it a perldoc-style docstring. I'll send this in soon

plugins/puppetdb/puppetdb
lelutin: I took a look a the "well-known categories" (http://guide.munin-monitoring.org/en/latest/reference/graph-category.html#well-known-categories) and couldn't find something that would fit better.

should we create a category like "configuration_management" or should they rather be in a "puppet" category ?

plugins/puppetdb/puppetdb
lelutin: I did not implement it because my use case is checking a local puppetdb system, bu I do think that verifying certificates is an important feature for this plugin since many folks have their puppetdb systems on a different server than their puppetmaster

plugins/puppetdb/puppetdb
sumpfralle: I would suggest `system` or `other` for now. The list of categories is surely showing its age, but let us handle this separately.

plugins/puppetdb/puppetdb
sumpfralle: OK. So we assume, that people using https for connecting to the puppetdb use a proper certificate. Good.

sumpfralle:
Thank you for your further improvements.
I think, only the `category` issue remains (but could be kept as it is, if you prefer).
Feel free to squash your commits into one, if you prefer a shorter history.

I will merge it, as soon as you notify me. Thank you!

lelutin:
Hi, I've just squashed all commits and added a change of category for both puppet plugins on top and force-pushed.

I've decided to not switch to urllib for now since it seems like it might complexify adding verifying https connections with an arbitrary CA -- which I believe might be something that folks do with puppetdb and the puppetmaster's CA

sumpfralle:
Great - thank you!

PR #1037, opened 2020-01-20
puppetdb: implement cert validation and client cert auth

lelutin:
It is very common for PuppetDB installs to run on a different host than
the puppetmaster. In such cases, a certificate file is normally used to
establish an encrypted communication to the server.

The most common setup for this server certificate is to use a
certificate that was signed by the puppetmaster's CA, so one would want
to verify the server cert against this same CA (it should be present on
puppet clients).

Moreover, to ensure that *only* puppet clients can communicate with
PuppetDB, a pair of client certificat/key files are usually used to
authenticate clients.

### DIFF
--- a/plugins/puppet/puppetdb
+++ b/plugins/puppet/puppetdb
@@ -21,11 +21,33 @@ Plugin configuration parameters:
     Time in seconds (int) to wait for a result when querying the REST API. By
     default, wait for 2 seconds
 
+  * ca:
+    Path to the Certificate Authority used for verifying a cert received from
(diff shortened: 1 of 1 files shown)


sumpfralle:

The style checks reported the following issue:
```
plugins/puppet/puppetdb:183:100: E501 line too long (119 > 99 characters)
```

Could you please fix this and force-push your updated commit?

Thank you!

lelutin:
Hi! thanks for the quick review. I've changed the line to split the string in two smaller pieces and ammended/force-pushed the change

sumpfralle:
Thank you!

PR #1040, opened 2020-02-02
Fix prosody plugin by encoding telnetlib commands

### LINKED ISSUE #954: Doesnt show informations about client presence
As the title suggest I don't get information about the clients' presence. Using Prosody 0.10

codeurimpulsif:
- Encode telnet commands because [telnetlib only accept bytes](https://docs.python.org/3/library/telnetlib.html#telnetlib.Telnet.write)
- Resolved #954 with proposed regex

### DIFF
--- a/plugins/prosody/prosody_
+++ b/plugins/prosody/prosody_
@@ -56,20 +56,20 @@ def main():
         else:
             connection_count_re = re.compile(r"Total:\s(\d+)\s")
             telnet = telnetlib.Telnet(host, port)
-            telnet.write("c2s:show_secure()\n")
-            telnet_response = telnet.read_until("secure client connections", 5)
-            parsed_info = connection_count_re.findall(telnet_response)
+            telnet.write(b"c2s:show_secure()\n")
+            telnet_response = telnet.read_until(b"secure client connections", 5)
+            parsed_info = connection_count_re.findall(telnet_response.decode('ascii'))
             secure_client_connections = int(parsed_info[0])
             print("secure_client_connections.value %s" % secure_client_connections)
 
-            telnet.write("c2s:show_insecure()\n")
-            telnet_response = telnet.read_until("insecure client connections", 5)
-            parsed_info = connection_count_re.findall(telnet_response)
+            telnet.write(b"c2s:show_insecure()\n")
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you!

msebald:
@sumpfralle can I pull this fix already? My prosody_ still has the timestamp from February 3rd and I get this error:

```
munin-run prosody_c2s
  File "/etc/munin/plugins/prosody_c2s", line 180, in <module>
    main()
  File "/etc/munin/plugins/prosody_c2s", line 61, in main
    parsed_info = connection_count_re.findall(telnet_response.decode('ascii'))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 14629: ordinal not in range(128)
```

sumpfralle:
The `prosody_` plugin in the repository seems to be newer than your version: https://github.com/munin-monitoring/contrib/commits/master/plugins/prosody/prosody_

Do you have an idea, why?

msebald:
@sumpfralle I just doublechecked and downloaded the (raw file of the) version which is here: https://github.com/munin-monitoring/contrib/blob/master/plugins/prosody/prosody_

This version seems to be from August 2020. And I still get the error mentioned above.

sumpfralle:
Maybe prosody sends utf-8 instead of ascii? Maybe try to change the line above to `.decode('utf8')` and see, if it works, afterwards?

msebald:
@sumpfralle Thanks, that was the right hint! I changed several `.decode('ascii')` to `.decode('utf8')` in prosody_ and the plugin is producing numbers again:

```
munin-run prosody_c2s
secure_client_connections.value 1985
insecure_client_connections.value 4
all_client_connections.value 1989
```

```
munin-run prosody_s2s
outgoing_connections.value 1308
incoming_connections.value 1276
```

FYI, I am using Prosody 0.11 Nightly (1nightly125-1~buster), so this should be a pretty common version. I would say it makes sense to change prosody_ in the repository as it might not work on many newer installations anymore.

sumpfralle:
Thank you for testing this change!

> I would say it makes sense to change prosody_ in the repository as it might not work on many newer installations anymore.

Of course :)

It is now fixed in c5daa06d.

PR #1041, opened 2020-02-03
twemproxy: add python3 support

pboesch:
`iteritems() -> items()`


sumpfralle:
Thank you!

I guess, you verified that the plugin works with python3? In this case it would be great if you would also change the shebang to `python3`. And if you cannot stop yourself: `flake8`-cleanliness and a well-formatted perldoc header would be great :)

Of course, you may feel free to reject my long wish list. In this case I will gladly merge your current contribution. Thanks!

sumpfralle:
Merging-as-is - thank you!

PR #1042, opened 2020-02-05
Plugin wordpress2: add PHP7 supports

kozakana:
wordpress2 plugin is not working on PHP7.

* Fix the following error using PHP7 (mysql -> mysqli)
```
PHP Fatal error:  Uncaught Error: Call to undefined function mysql_connect() in /usr/share/munin/plugins/wordpress2:112
Stack trace:
#0 {main}
  thrown in /usr/share/munin/plugins/wordpress2 on line 112
```
* The previous version is still supported in this fix
```
if(function_exists('mysql_connect')) {
  // less than PHP7
}else{
  // PHP7
}
```

### DIFF
--- a/plugins/wordpress/wordpress2
+++ b/plugins/wordpress/wordpress2
@@ -109,13 +109,26 @@ $comments  = 0;
 $pingbacks = 0;
 
 // GET DATA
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you!

PR #1043, opened 2020-02-07
Update mongodb plugins

NoxInmortus:
1. Add documentation for following plugins :
- mongo_btree
- mongo_conn
- mongo_lock
2. Python3 support, feat commun munin-node conf, and small fixes for following plugins :
- mongo_collection_
- mongo_lag
- mongo_mem
- mongo_ops
- mongodb_conn
3. Add mongodb_docs plugin


plugins/mongodb/mongo_btree
```diff
@@ -1,6 +1,22 @@
-#!/usr/bin/python
+#!/usr/bin/env python
+"""
+=head1 NAME
+    MongoDB btree Plugin
 
-## GENERATED FILE - DO NOT EDIT
+=head1 APPLICABLE SYSTEMS
+
+    Works until MongoDB 2.7. The "indexCounters" field was removed in 2.8 version.
```
sumpfralle: Wouldn't it be possible to exclude this field in case it is missing?

plugins/mongodb/mongo_collection_
```diff
@@ -94,25 +94,25 @@ def getCollstats(graphtype):
...
                     continue
                 db = con[dbname]
-                for coll in db.collection_names():
+                for coll in db.list_collection_names():
                         if coll.startswith('system.'):
                                 continue
                         stats = db.command("collstats", coll)
                         collname = dbname + "_" + coll.replace('.', '_')
-                        if not stats_tmp.has_key(collname):
+                        if not collname in stats_tmp:
```
sumpfralle: `flake8` advises to use `... not in ...` instead of `not ... in ...` (avoiding the ambiguity of precedence).

plugins/mongodb/mongo_conn
```diff
@@ -1,6 +1,22 @@
-#!/usr/bin/python
+#!/usr/bin/env python
+"""
+=head1 NAME
+    MongoDB connections Plugin
 
-## GENERATED FILE - DO NOT EDIT
+=head1 APPLICABLE SYSTEMS
+
+    Works until MongoDB 3.6. The httpinterface was later removed.
```
sumpfralle: Is there a way to retrieve the values for this plugin in a different way for never versions of mongodb?

plugins/mongodb/mongodb_conn
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
sumpfralle: Are you sure, that the current plugin does not work with python3?

plugins/mongodb/mongodb_docs
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python
```
sumpfralle: The plugin looks like it should work with python3.

NoxInmortus:
Hello @sumpfralle 
- About : mongo_btree , I dont have the use for this plugin and as It works only on very old mongodb version, I dont see the point of maintain this plugin. I added a piece of documentation to avoid others wasting time trying to understand why it doesnt work. The "indexCounters" cannot be excluded as the plugin is based on this single field.
- About : mongo_collection_ , I will look onto it
- About mongo_conn : Quite the same thing than mongo_btree, but I wrote before the plugin mongodb_conn to exactly replace this one :)

- About your comments on python3 (mongodb_conn and mongodb_docs), and as to say, to every other plugins I edited there, the changes I made was exactly to support python3.

    mongo_collection_
    mongo_lag
    mongo_mem
    mongo_ops
    mongodb_conn
    mongodb_docs

All theses plugins now works on python2 and python3. 
To answer to the use of `#!/usr/bin/env python3` rather than `#!/usr/bin/env python` as you advise, I have to disagree on this matter. As they are compatible python2 and python3, fixing `#!/usr/bin/env python3` will, de-facto, breaks python2 compat. And I dont wish that.

sumpfralle:
> As they are compatible python2 and python3, fixing #!/usr/bin/env python3 will, de-facto, breaks python2 compat.

At least for Debian it means that the plugin will fail to work with the next release (*Bullseye* - to be released around 2021). Python2 will not be part of the release anymore and `/usr/bin/python` will deliberately be missing. This approach prevents accidental mis-usage of python3 for python2-only scripts/programs. Maybe the Debian release after *Bullseye* will provide `/usr/bin/python` again (pointing at python3), but at least for the time in between the plugin would fail to work.

The [migration notes in PEP394](https://www.python.org/dev/peps/pep-0394/#id29) are in line with this approach. They recommend to explicitly state the python version to be used in the shebang.

Thus we (maintainers of the munin plugins in this repository) also aim to change all python-related shebangs to an explicit version in order to avoid any confusion. This seems to be perfectly reasonable to me - especially now, that python2 is already *really finally* end-of-life.

NoxInmortus:
@sumpfralle Hey. What about now ?

plugins/mongodb/mongo_collection_
```diff
@@ -29,8 +29,8 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-#%# family=auto
-#%# capabilities=suggest autoconf
+# %# family=auto
+# %# capabilities=suggest autoconf
```
sumpfralle: `#%#` is a special marker. In this case you would need to use `# #%#` in order to satisfy the python style requirement, as well as munin's marker format.

sumpfralle:
It looks good to me now - thank you!

Just fix the last nitpick regarding munin's markers and I will merge your contributions. Thanks!

plugins/mongodb/mongo_collection_
NoxInmortus: here we are

sumpfralle:
Merged. Thank you!

NoxInmortus:
@sumpfralle thanks you for your time. Cheers

PR #1044, opened 2020-02-08
Rewrite docker plugin

### LINKED ISSUE #1027: docker plugins fails in CentOS 7, these can not found the counters files
The plugins docker_cpu and docker_memory try retrieve the information from files with this path

Example for docker_cpu

`/sys/fs/cgroup/cpuacct/docker/$id/cpuacct.usage_percpu`

But in CentOS 7 and others linux the path is something like 

`/sys/fs/cgroup/cpuacct/system.slice/docker-$id.scope/cpuacct.usage_percpu`

codeurimpulsif:
Based on `docker` Python package to talk to the Docker socket.

Fix:
- CPU
- Memory

Add:
- Containers number
- Containers status
- Images number
- Volumes number

Fix #1027

### DIFF
--- a/plugins/docker/docker_
+++ b/plugins/docker/docker_
@@ -1,0 +1,190 @@
+#!/usr/bin/env python3
+"""
+=head1 NAME
+
+docker_ - Docker wildcard-plugin to monitor a L<Docker|https://www.docker.com> host.
+
+This wildcard plugin provides at the moment only the suffixes C<containers>, C<images>, C<status>,
+C<volumes>, C<cpu> and C<memory>.
+
+=head1 INSTALLATION
+
+- Copy this plugin in your munin plugins directory
+- Install Python3 "docker" package
+
+=over 2
+
+    ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_containers
+    ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_cpu
+    ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_images
+    ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_memory
+    ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_status
+    ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_volumes
+
+=back
+
+After the installation you need to restart your munin-node:
+
+=over 2
+
+    systemctl restart munin-node
+
+=back
+
+=head1 CONFIGURATION
+
+This plugin need to run as root, you need to create a file named docker placed in the
+directory /etc/munin/plugin-conf.d/ with the following config (you can also use
+Docker environment variables here as described in
+https://docs.docker.com/compose/reference/envvars/):
+
+=over 2
+
+    [docker_*]
+    user root
+    env.DOCKER_HOST unix://var/run/docker.sock
+
+=back
+"""
+
+import os
+import sys
+import docker
+
+
+def print_containers_status(client):
+    running = 0
+    paused = 0
+    created = 0
+    restarting = 0
+    removing = 0
+    exited = 0
+    dead = 0
+    for container in client.containers.list():
+        if container.status == 'running':
+            running += 1
+        elif container.status == 'paused':
+            paused += 1
+        elif container.status == 'created':
+            created += 1
+        elif container.status == 'restarting':
+            restarting += 1
+        elif container.status == 'removing':
+            removing += 1
+        elif container.status == 'exited':
+            exited += 1
+        elif container.status == 'dead':
+            dead += 1
+    print('running.value', running)
+    print('paused.value', paused)
+    print('created.value', created)
+    print('restarting.value', restarting)
+    print('removing.value', removing)
+    print('exited.value', exited)
+    print('dead.value', dead)
+
+
+def print_containers_cpu(client):
+    for container in client.containers.list():
+        stats = container.stats(stream=False)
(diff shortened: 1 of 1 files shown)


plugins/docker/docker_
```diff
@@ -0,0 +1,198 @@
...
+            print("graph_vlabel containers")
+            print("graph_category virtualization")
+            print("running.label RUNNING")
+            print("paused.label PAUSED")
+            print("created.label CREATED")
+            print("restarting.label RESTARTING")
+            print("removing.label REMOVING")
+            print("exited.label EXITED")
+            print("dead.label DEAD")
+            sys.exit(0)
```
sumpfralle: I think, it would improve the readability of the flow control, if you simply remove the `exit` calls.

At the moment `exit` is called after every configuration branch, but not after printing the values. There is not difference in practice, but it feels a bit confusing for me.

plugins/docker/docker_
```diff
@@ -0,0 +1,198 @@
...
+            getContainersCpu(client)
+    elif wildcard == "memory":
+        if mode == "config":
+            print("graph_title Docker containers memory usage")
+            print("graph_args --base 1024 -l 0")
+            print("graph_vlabel Bytes")
+            print("graph_category virtualization")
+            print("graph_info This graph shows docker container memory usage.")
+            for container in client.containers.list():
+                print("{}.label {}".format(container.name, container.name))
```
sumpfralle: How about using a `AREASTACK` for graphing the memory consumption?

plugins/docker/docker_
```diff
@@ -0,0 +1,198 @@
...
+
+=back
+"""
+
+import os
+import sys
+import docker
+
+
+def getContainersStatus(client):
```
sumpfralle: Maybe it is just my obsession with naming conventions: maybe you want to change the names of the functions from `get...` to `print...`?

btw: PEP8 recommends the underscore name style for functions (not camel case). But this is just a matter of taste ...

plugins/docker/docker_
```diff
@@ -0,0 +1,198 @@
...
+=head1 CONFIGURATION
+
+When you want to change the default host (unix://var/run/docker.sock) or any other things
+(see https://docs.docker.com/compose/reference/envvars/) do it in a file named docker placed
+in the directory /etc/munin/plugin-conf.d/ with a config like this:
+
+=over 2
+
+    [docker_*]
+    user root
```
sumpfralle: I guess, the `user root` setting is *always* necessary in order to gain access to the control socket?

plugins/docker/docker_
```diff
@@ -0,0 +1,198 @@
...
+C<volumes>, C<cpu> and C<memory>.
+
+=head1 INSTALLATION
+
+- Copy this plugin in your munin plugins directory
+- Install Python3 "docker" package
+
+=over 2
+
+    chmod 755 docker_
```
sumpfralle: The executable should be set, anyway. Thus the user should not have to do this.

plugins/docker/docker_
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python3
+"""
+=head1 NAME
+
+docker_ - Docker wildcard-plugin to monitor a L<Docker|https://www.docker.com> host.
+
+This wildcard plugin provides at the moment only the suffixes C<containers>, C<images>, C<status>,
```
sumpfralle: Just for the inspiration: maybe it would be nice to switch this plugin from the wildcard style to a multigraph plugin? In this case the plugin would output the configuration (or values) of all sections and just add `multigraph docker_cpu` (and so on) lines in front of every block.

This would simplify the plugin installation (multiple symlinks are not required). But if it is common to use only a few of these plugins, then a wildcard plugin could be more suitable.

plugins/docker/README.md
```diff
@@ -0,0 +1,43 @@
+# Docker
```
sumpfralle: I think it is sufficient to keep all the documentation in the perldoc header of the plugin. Otherwise it could be confusing for people that the README in the directory describes only one of the docker-related plugins in this directory.

plugins/docker/docker_
codeurimpulsif: Of course, I just removed that!

plugins/docker/docker_
codeurimpulsif: You're right, I'm a camel case addict actually trying to change that :D

plugins/docker/docker_
codeurimpulsif: Yes, I wrote a precision for this in documentation.

plugins/docker/README.md
codeurimpulsif: Removed :)

plugins/docker/docker_
codeurimpulsif: I never used multigraph plugins but it can be a good thing. I will take a look on this type of plugin.

sumpfralle:
Thank you for your quick adoption of changes!

Please ping me again, as soon as you like the state of your plugin.

Feel free to squash and force-push your commits (or just keep the full history).

NoxInmortus:
Hey, nice to see this pull request, I was just looking at the current docker plugins. Can't wait to try those.
Cheers.

codeurimpulsif:
I think it's ok for me, maybe someone could change somethings in the future in this plugin but I don't have much time and I think it can be merge if it's ok for you too :-)

sumpfralle:
> I think it's ok for me, maybe someone could change somethings in the future in this plugin but I don't have much time and I think it can be merge if it's ok for you too :-)

Great - thank you!

PR #1046, opened 2020-02-28
typo: retransmited -> retransmitted

corbolais:
* quick fix
* if this "typo" was intentional ignore this PR

Signed-off-by: corbolais <corbolais@gmail.com>


sumpfralle:
Maybe the wrong spelling was correct at some time - right now it seems to be wrong, indeed.

Thanks!

PR #1047, opened 2020-03-26
Merge Mikrotik plugins

sumpfralle:
The two SNMP plugins for Mikrotik routers are very similar. Thus a `multigraph` should be easier to maintain and extend in the long run.

Based on #1028 by @alejandroscf.


plugins/router/snmp__mikrotik
```diff
@@ -0,0 +1,145 @@
...
+        die "Croaking: $error";
+}
+
+
+if ($ARGV[0] and $ARGV[0] eq "config")
+{
+    print "host_name $host\n";
+    $response = $session->get_request($sysFlashTotalOID);
+    if (defined $response) {
+        print "multigraph flash";
```
alejandroscf: I tested it and I think we should add "\n" here.

plugins/router/snmp__mikrotik
```diff
@@ -0,0 +1,145 @@
...
+        print "graph_vlabel bytes\n";
+        print "sysFlashTotal.label Total Memory\n";
+        print "sysFlashTotal.draw AREA\n";
+        print "sysFlashUsage.label Used Memory\n";
+        print "sysFlashUsage.draw AREA\n";
+    }
+
+    $response = $session->get_request($sysRAMTotalOID);
+    if (defined $response) {
+        print "multigraph ram";
```
alejandroscf: Also here

plugins/router/snmp__mikrotik
```diff
@@ -0,0 +1,145 @@
...
+        print "sysRAMTotal.label Total Memory\n";
+        print "sysRAMTotal.draw AREA\n";
+        print "sysRAMUsage.label Used Memory\n";
+        print "sysRAMUsage.draw AREA\n";
+    }
+    exit 0 unless (($ENV{MUNIN_CAP_DIRTYCONFIG} || 0) == 1);
+}
+
+
+print "multigraph flash";
```
alejandroscf: And here

plugins/router/snmp__mikrotik
```diff
@@ -0,0 +1,145 @@
...
+$response = $session->get_request(-varbindlist => [$sysFlashUsageOID, $sysFlashTotalOID]);
+if (defined $response) {
+        print "sysFlashUsage.value ", $response->{$sysFlashUsageOID}*1024, "\n";
+        print "sysFlashTotal.value ", $response->{$sysFlashTotalOID}*1024, "\n";
+} else {
+        print "sysFlashUsage.value U\n";
+        print "sysFlashTotal.value U\n";
+}
+
+print "multigraph ram";
```
alejandroscf: And finally here.

plugins/router/snmp__mikrotik
alejandroscf: I tested it and I think we should add "\n" here.

sumpfralle:
Thank you for testing my changes and spotting the missing bits!

I fixed these now and merged your contribution. Thank you!

sumpfralle:
Just in case you did not know: since munin v2.0.52 you can use `munin-get` to install and maintain plugins from the `contrib` repository (and other places).

PR #1049, opened 2020-03-27
Added monitoring of Mikrotik router temperature

alejandroscf:
Monitor temperature reported by Mikrotik routers via SNMP protocol.

PS: The multigraph approach really make easier add functionality.

### DIFF
--- a/plugins/router/snmp__mikrotik
+++ b/plugins/router/snmp__mikrotik
@@ -49,6 +49,7 @@ my $sysFlashUsageOID = "1.3.6.1.2.1.25.2.3.1.6.131072";
 my $sysFlashTotalOID = "1.3.6.1.2.1.25.2.3.1.5.131072";
 my $sysRAMUsageOID = "1.3.6.1.2.1.25.2.3.1.6.65536";
 my $sysRAMTotalOID = "1.3.6.1.2.1.25.2.3.1.5.65536";
+my $sysTempOID = "1.3.6.1.4.1.14988.1.1.3.10.0";
 
 my $response;
 
@@ -57,6 +58,7 @@ if (defined $ARGV[0] and $ARGV[0] eq "snmpconf") {
     print "require $sysFlashTotalOID\n";
     print "require $sysRAMUsageOID\n";
     print "require $sysRAMTotalOID\n";
+    print "require $sysTempOID\n";
     exit 0;
 }
 
@@ -112,6 +114,17 @@ if ($ARGV[0] and $ARGV[0] eq "config") {
         print "sysRAMUsage.label Used Memory\n";
         print "sysRAMUsage.draw AREA\n";
     }
+
+    $response = $session->get_request($sysTempOID);
+    if (defined $response) {
+        print "multigraph temp\n";
+        print "graph_vlabel degC \n";
+        print "graph_title Temperature\n";
+        print "graph_category sensors\n";
+        print "graph_info This graph shows the router's temperature.\n";
+        print "sysTemp.label Temperature\n";
+        print "sysTemp.type GAUGE\n";
+    }
     exit 0 unless (($ENV{MUNIN_CAP_DIRTYCONFIG} || 0) == 1);
 }
 
@@ -134,6 +147,14 @@ if (defined $response) {
 } else {
         print "sysRAMUsage.value U\n";
         print "sysRAMTotal.value U\n";
+}
+
+print "multigraph temp\n";
+$response = $session->get_request(-varbindlist => [$sysTempOID]);
+if (defined $response) {
+        print "sysTemp.value ", $response->{$sysTempOID}/10, "\n";
+} else {
+        print "sysTemp.value U\n";
 }
 
 # vim:syntax=perl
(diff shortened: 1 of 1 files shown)


plugins/router/snmp__mikrotik
```diff
@@ -57,6 +58,7 @@ if (defined $ARGV[0] and $ARGV[0] eq "snmpconf") {
     print "require $sysFlashTotalOID\n";
     print "require $sysRAMUsageOID\n";
     print "require $sysRAMTotalOID\n";
+    print "require $sysTempOID\n";
```
alejandroscf: I'm not sure about this line as some routers didn't report temperature and I don't know if it could affect how the snmpconf parameter works for that routers.

sumpfralle: Probably you are right (I do not know SNMP plugins too well). Thus maybe just remove it and add a comment above, that only attributes available on all routers should be listed here as *required*.

At least this is my understanding.

sumpfralle:
> PS: The multigraph approach really make easier add functionality.

I am glad, you like it :)

You changes look good to me.

plugins/router/snmp__mikrotik
alejandroscf: You are right, I found it in the file [SNMPConfig.pm](https://github.com/munin-monitoring/munin/blob/master/lib/Munin/Node/SNMPConfig.pm). All OIDs are checked and if any fails then  the plugins is not suggested, so the it should be removed. I'll do it.

But I found out that the requirements for the other OIDs are not working as expected. Just now the 'snmp__mikrotik' plugin is being recommended for any device, even if the OID is not present (at least in my system).
That happen because the `require {OID}` could have a second parameter like `require {OID} [{filter}]` and without a filter the plugin is suggested regardless of the OID. I've tested it against devices from other vendors (that they don't have any of the requested OIDs) and it get suggested anyway.

The lines should be something like:

`print "require $OID ^\\d\n";`

But I don't know if I should fix that in this PR or in another.

sumpfralle: The filter should be optional (at least from reading the code). But maybe the implementation of `_snmp_check_require` in munin is broken?

Just for testing a theory - maybe you could change the following line:
```perl
my $response = $session->get_request($oid);
```
into the following:
```perl
my $response = $session->get_request(-varbindlist => [ $oid ],);
```

At least this seems to be documented way for calling `get_request`. Does this fix the issue for you?

(I do not have an SNMP-capable device around, thus I cannot test it easily)

alejandroscf: I've tested adding the '-varbindlist' parameter and it has no effect. I've also tested if any error was returned when trying to get an inaccessible oid and no error was reported. I found this [bug](https://rt.cpan.org/Public/Bug/Display.html?id=131496) that it's not exactly the same but has a similar effect.
I noticed that other plugins like snmp__load or snmp__print_supplies just use the filter parameter.

sumpfralle: Thank you for trying it!
What a pity, that it did not work :(
I will try to debug this tomorrow at work - there are some SNMP switches around.

> I noticed that other plugins like snmp__load or snmp__print_supplies just use the filter parameter.

Anyway this would be a bug in munin (if I understand it correctly). Thus I would really like to fix it at its root instead of adding workarounds everywhere :)

sumpfralle: How do we proceed with this plugin now?

alejandroscf: I've been testing a perl script with only Net::SNMP and I found what looks like a bug in that perl library that didn't throw an error when trying to get an inaccessible oid when working with version 2c of snmp. I'm going to report it, maybe the cause is the same that the other bug.

We can wait until this is solved (but Net::SNMP doesn't looks very active) or I can upload the version I have with the filter parameter and merge it.

alejandroscf: Here is the [bur report](https://rt.cpan.org/Public/Bug/Display.html?id=132323) to Net::SNMP if you want to test it or try to found a solution.

sumpfralle: Thank you for digging further into this issue!

I would be glad, if you would take the following steps:
1. open an issue against munin describing the current problem with the `required` OID setting (I would have a hard time putting the details together, sorry)
2. apply your *filter* workaround to the plugin and reference the issue above (in order to remove the workaround somewhen in the future)

Thank you a lot for your attention to the details!

sumpfralle:
@alejandroscf: did you already find the time to look into the filter issue? Do you want to prepare the plugin with the added *filter* workaround for now?

(just curious - no pressure intended ...)

alejandroscf:
I mostly work in this the weekends and last one I had to work , but tomorrow is holiday and I just posted the issue #1073

plugins/router/snmp__mikrotik
alejandroscf: Just done both things!

sumpfralle:
Great - thank you!

PR #1050, opened 2020-03-29
Remove of deprecated api calls as of 0.9.7

Sephtex:
Aligning rtorrent rtrom_ modules same way as rtom_allsessions_ which was done in munin-monitoring/contrib#921:
In version 0.9.7 deprecated xmlrpc calls were dropped. The added files contain the needed changes for version 0.9.7+ of rtorrent.

Fixes: https://github.com/rakshasa/rtorrent/issues/757

### DIFF
--- a/plugins/rtorrent/rtom_allsessions_mem
+++ b/plugins/rtorrent/rtom_allsessions_mem
@@ -66,7 +66,7 @@ my @ports       = split /,/, $ENV{"port"} || "";
 # detect rtorrent version
 use version;
 my %rtorrent_version;
-sub get_socket_version {
+sub get_rtorrent_version {
 	my $version;
 	my $line_version= "<?xml version=\"1.0\" encoding=\"utf-8\"?><methodCall><methodName>system.client_version</methodName></methodCall>";
 	my $llen	= length $line_version;
@@ -91,7 +91,7 @@ sub rtorrent_version_lower_than {
 		  {
 		    socket( SOCK, PF_UNIX, SOCK_STREAM, 0 ) or die;
 		    connect( SOCK, sockaddr_un( $socket ) ) or die $!;
-		    get_socket_version $port;
+		    get_rtorrent_version $port;
 		    close (SOCK);
 		  }
 		} else {
@@ -99,7 +99,7 @@ sub rtorrent_version_lower_than {
 		  {
 		    socket( SOCK, PF_INET, SOCK_STREAM, getprotobyname( "tcp" ) );
 		    connect( SOCK, sockaddr_in( $port, inet_aton( $ip ) ) );
-		    get_socket_version $port;
+		    get_rtorrent_version $port;
 		    close (SOCK);
 		  }
 		}
--- a/plugins/rtorrent/rtom_allsessions_mem
+++ b/plugins/rtorrent/rtom_allsessions_mem
@@ -27,7 +27,6 @@
 #       ip              rTorrent's ip address                   - using scgi_port       - needed, when "src" is NOT set to "socket"
 #       port            rTorrent's scgi port (scgi_port)        - using scgi_port       - needed, when "src" is NOT set to "socket"
 #       category        Change graph category
-#       api             use "pre09" (pre 0.9.0) or "current" (0.9.0+, the default) API calls
 #
 # Configuration example
 #
@@ -36,7 +35,6 @@
 #       env.src socket
 #       env.socket /home/user/torrent/.socket/rpc.socket,/home/user/torrent/.socket/rpc.socket
 #       env.category Category
-#       env.api current
 #
 #       [rtom_allsessions_*]
 #       user username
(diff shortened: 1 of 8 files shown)


Sephtex:
After I used those changes locally with my rtorrent, my graphs where back! :)

sumpfralle:
Thanks for taking care of this issue!

I took a look at the changes and I have the feeling, that this change can also be detected automatically?
(right now users would need to configure this change manually)

I would be happy, if you could add automatic detection, instead. What do you think?

Sephtex:
Yes sure! I had that same idea, although I never wrote any perl so I just duplicated the existing piece of code. Anyway, since I rather had seen it automatically detected too, so I'll check on it whenever I get some time next week :)

Sephtex:
I got already a piece to to get the version and replace the if `if ($api =~ /pre09/) {` with that check. Later on (at latest next week) I'll create a function for it and propagate it to all modules :)

sumpfralle:
> I got already a piece to to get the version and replace the if `if ($api =~ /pre09/) {` with that check.

Cool! I am glad, that you are smoothing this transition for the users of the Munin plugin ...

(btw.: you can easily install the plugin via `munin-get` - available since Munin v2.0.52)

Sephtex:
Cool, nice to see you can even specify your own repository!
That is a nice feature to add in my automation as I cloned this repository to symlink to a module.

I also didn't tested all my changes yet, only one before I copied all the changes and aligned other modules. I'll test this weekend if no one else did it before I have done it.

I'll keep you posted!

sumpfralle:
> Cool, nice to see you can even specify your own repository!
> That is a nice feature to add in my automation as I cloned this repository to symlink to a module.

I am glad, that you like it!
Indeed I was also suffering from having a hard time to use `contrib` plugins via ansible. Feel free to suggest additional features, if you stumble upon inconveniences.

Sephtex:
I needed to update the _vol modules as it seemed that the SOCK was reused. So I execute the function to get the version first, so the next executions it'll just use rtorrent_version veriable.

The rtrom_{mem,peers,spdd,vol} are working, but for some reason, the rtrom_allsessions_* modules aren't returning anything with munin-run. And I can't seem to see why :/, I tried:
```
# perl -c rtom_allsessions_mem
rtom_allsessions_mem syntax OK
```
And when I use munin-run with the original one, before the addition of the version function, it works without issues.

Regarding munin-get @sumpfralle , it's nice, although it doesn't add any benefit for me, as I use anisble git to clone the repo and symlink to the file in the cloned repo which is faster. I thought it would get the module directly from the git repository, with maybe `git archive`. Maybe it could be a nice enhancement to that module, eg: when you use a flag --no-cache that it'll use git archive to get the single file (module) directly?

Sephtex:
the rtom_allsessions_ are using loops, and socker/port is an array, I'm fixing it :)

Sephtex:
For both:
At current commit mem and spdd is ok
peers and vol needs to be fixed

Sephtex:
Ok fixed them all!

allsessions are giving me the same value as the modules for single instance!
And with all the changes, it'll detect the version of every instance.

I'm not fond of how I added all the changes, but it works! :)

sumpfralle:
> I thought it would get the module directly from the git repository, with maybe git archive.

As I understand `git archive`, it requires a local working copy (not a repository reference). But maybe I am wrong. Maybe you could add some more details? I would happily like to improve `munin-get`.

> I'm not fond of how I added all the changes, but it works! :)

My perl fu is surely limited, but from my point of view it looks good.

The only detail that bothers me a bit is the code duplication.

So if you are in the mood for a bit more improvements, maybe you would like to turn both plugins (or even the others in the `rtorrent` directory) into a single `multigraph` plugin?

The resulting plugin would emit the following data:
```
multigraph  rtom_allsessions_mem
graph_title ...
...

multigraph  rtom_allsessions_peers
graph_title ...
...
```

Thus we could replace the multiple files by just one.

What do you think?

Otherwise just give me a hint, if you want to get your changes merged as-is.

Sephtex:
`git archive` apparently doesn't seem to be working with github. But FYI it doesn't require a local copy, it'll fetch a directory-tree/file from a specific commit/branch/head (so in general, what you exactly want) as archive format to stdout by default.
But since it doesn't seem to work for github, it'll be not much of use in this case and wget/curl would even be better. So I think that how munin-get works is perfect!

My perl knowledge started on the day I worked on this, so I think my perl knowledge will be less than you :)

The duplication is the function I wrote to get the version right?
I like your multigraph suggestion, but I have no idea how to do that. Btw, wouldn't that imply creating new graphs? And probably also creating a new module? As I suppose converting it to a module, which creates a multigraph might reset all the collected data right?

I would actually go ahead and merge this one as it'll already improve for anyone using those modules, or especially those who wants to use those, so they wouldn't need to specify the api version anymore on the allsessions, and it'll work out of the box for older and newer usage.

sumpfralle:
> But FYI it doesn't require a local copy, [..]

Interesting - I will take a look at this.

> I don't like that either, so I like your multigraph suggestion, but I have no idea how to do that. Btw, wouldn't that imply creating new graphs?

Not necessarily: if the new `multigraph` names are picked properly (identical to the previous plugin names), then all data is re-used. I would be happy, if you could take a stab at this. Your contribution will be very welcome!

> I would actually go ahead and merge this one [..]

Sounds good. Many thanks for keeping up with my quests for perfection! :)

PR #1051, opened 2020-03-30
Simple plugin to monitor coturn turn/stun server ...

brknkfr:
... via telnet

### DIFF
--- a/plugins/coturn/coturn_sessions
+++ b/plugins/coturn/coturn_sessions
@@ -1,0 +1,109 @@
+#!/bin/sh
+
+set -e
+
+: << =cut
+
+=head1 NAME
+
+coturn - Simple plugin to monitor sessions on coturn server
+
+=head1 CONFIGURATION
+
+This plugin requires a password set in /etc/turnserver.conf for telnet 
+cli access. A telnet client has to be installed. Optionally you can 
+specify the port of the cli interface.
+
+[coturn_sessions]
+env.clipassword VERYSECUREPASSWORD
+env.port PORT
+
+=head1 AUTHOR
+
+Copyright (C) 2020 Sebastian L. <sebastiab@momou.ch>
+
+=head1 LICENSE
+
+GPLv2
+
+=head1 MAGIC MARKERS
+
+ #%# family=manual
+ #%# capabilities=autoconf
+
+=cut
+
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
+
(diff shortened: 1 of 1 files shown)


plugins/coturn/coturn
```diff
@@ -0,0 +1,107 @@
...
+	for i in udp tcp dtls tls
+	do
+		echo "pu $i"
+		sleep 1
+	done
+}
+
+case $1 in
+    autoconf)
+        if [[ -x /usr/bin/turnserver ]]; then
```
sumpfralle: I think, this is the only *bashism* you are using in this script. But you could easily use single brackets instead in order to a work with a simple `sh`, too.

plugins/coturn/coturn
```diff
@@ -0,0 +1,107 @@
...
+            echo yes
```
sumpfralle: Since `telnet` is not such a common tool anymore, I would suggest to also check for its existence (`[ -z "$(which telnet)" ]`).

plugins/coturn/coturn
```diff
@@ -0,0 +1,107 @@
...
+            exit 0
+        fi
+        ;;
+   config)
+
+		echo "graph_title Sessions on coturn server"
+		echo "graph_vlabel turn sessions by protocols"
+		echo "graph_info This graph shows active sessions sorted by different protocols on the coturn server."
+		echo "graph_args --base 1000 --lower-limit 0"
+		echo "udp_sessions.label udp sessions"
```
sumpfralle: Maybe you want to use a stacked visualization for these connections? Simply add `.draw AREASTACK` for each field.

brknkfr:
> I added some comments below. Most of them are just nitpicking ...
> 
This is my first munin-plugin, so thank you very much for your inputs.

sumpfralle:
> This is my first munin-plugin, [..]

... and (despite the number of my nitpicks) it looks very well structured and carefully written to me.

There is just one small detail left (visible in the CI testing failure): `shellcheck` complains about the first use of the unreferenced variable `clipassword`. I would recommend to add something like the following near to the definition of `cliport`: `clipassword=${clipassword:-}`. This would also serve nicely as a kind of plugin interface (i.e. all expected environment variables are handled in the beginning).

Thank you!

sumpfralle:
btw: feel free to squash (and force-push) your commits into one, if you like.

sumpfralle:
Thank you for this fine new plugin!

PR #1052, opened 2020-03-30
Simple jitsi-videobridge plugin

brknkfr:
... taken from / credits to https://gitlab.gwdg.de/bwegman1/Munin-Plugins/commit/a3c79badaec88d3c665b61db3da09e149d1b788f

### DIFF
--- a/plugins/jitsi/jitsi-videobridge
+++ b/plugins/jitsi/jitsi-videobridge
@@ -1,0 +1,67 @@
+#!/bin/bash
+# -*- sh -*-
+
+: <<=cut
+
+=head1 NAME
+
+jitsi-videobridge-conferences - Jitsi conferences
+
+=head1 APPLICABLE SYSTEMS
+
+Any system with jitsi-videobridge installed and enabled colibri 
+statistics
+
+=head1 CONFIGURATION
+
+No configuration necessary.
+
+=head1 INTERPRETATION
+
+This plugin collects statistics from the jitsi-videobridge HTTP-API.
+
+=head1 MAGIC MARKERS
+
+  #%# family=manual
+  #%# capabilities=autoconf
+
+=head1 BUGS
+
+None known
+
+=head1 AUTHOR
+
+benedikt.wegmann@gwdg.de
+
+=head1 LICENSE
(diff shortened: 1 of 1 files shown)


brknkfr:
Ah, do this again, cleaner.

sumpfralle:
Thanks for this contribution!

A first spontaneous thought: maybe the plugin deserves a split into multiple graphs (e.g. for sessions, users and so on). This could improve the usability of the visualization by comparing only comparable metrics. A simple approach for this would be the `multigraph` plugin style (see [router/snmp__mikrotik](https://github.com/munin-monitoring/contrib/blob/master/plugins/router/snmp__mikrotik) for an example.

Regarding the source of the plugin: I could imagine, that the maintainer of that [repository](https://gitlab.gwdg.de/bwegman1/Munin-Plugins) is actively maintaining this specific plugin. In this case it could be more fruitful to engage him and maybe contribute your changes there. I am a bit reluctant to add plugins to the `contrib` repository, which are actively maintained in another place. This usually leads to duplicated efforts in the long run. Thus I would be happy, if you could discuss this with the current maintainer.

(even though I would be very eager to add a jitsi plugin to the `contrib` repository)

brknkfr:
> A first spontaneous thought: maybe the plugin deserves a split into multiple graphs (e.g. for sessions, users and so on). This could improve the usability of the visualization by comparing only comparable metrics. A simple approach for this would be the `multigraph` plugin style (see [router/snmp__mikrotik](https://github.com/munin-monitoring/contrib/blob/master/plugins/router/snmp__mikrotik) for an example.
> 
That's what I thought aswell.

> Regarding the source of the plugin: I could imagine, that the maintainer of that [repository](https://gitlab.gwdg.de/bwegman1/Munin-Plugins) is actively maintaining this specific plugin. In this case it could be more fruitful to engage him and maybe contribute your changes there. I am a bit reluctant to add plugins to the `contrib` repository, which are actively maintained in another place. This usually leads to duplicated efforts in the long run. Thus I would be happy, if you could discuss this with the current maintainer.
> 
Ok, you're right.

PR #1053, opened 2020-04-03
Add jitsi-videobridge plugin

brknkfr:
Simple multigraph plugin with one graph for current session infos and one
graph for total conferences created.

### DIFF
--- a/plugins/jitsi_videobridge/jitsi_videobridge
+++ b/plugins/jitsi_videobridge/jitsi_videobridge
@@ -1,0 +1,135 @@
+#!/bin/sh
+
+set -e
+
+: << =cut
+
+=head1 NAME
+
+jitsi_videobridge - Monitor sessions and conferences 
+on a jitsi-videobridge
+
+=head1 APPLICABLE SYSTEMS
+
+Jitsi-Videobridge instances
+
+=head1 CONFIGURATION
+
+Requires enabled colibri statistics on jitsi-videobridge and an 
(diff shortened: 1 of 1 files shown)


plugins/jitsi-videobridge/jitsi-videobridge
```diff
@@ -0,0 +1,132 @@
...
+
+    autoconf)
+        if [ -x /usr/bin/curl ]; then
+            if [ -x /usr/bin/jq ]; then
+                echo "yes"
+                exit 0
+            else
+                echo "no (jq not found)"
+                exit 0
+            fi
```
sumpfralle, line 58: For a perfectly self-configuring plugin you should add a test for an empty result of `curl "$COLIBRI_URL"` here. In this case you can also use the family `auto`.

btw: the `/stats` part of the URL path is probably based on the API of the jitsi service. Maybe you want to let the configurable part of the URL only be `http://localhost:8080/colibri` and add the `/stats` path in the plugin. This would allow the plugin to re-adjust the requested path, in case of changes in the jitsi API.

But this is a matter of taste, I guess.

sumpfralle:
Great - thanks!

One last thing: if you do not mind, then I would like to rename `plugins/jitsi_videobridge/jitsi_videobridge` to `plugins/jitsi/jitsi_videobridge`. Or do you think, plugins for the videobridge component should be kept separete from other jitsi-related plugins?

PR #1055, opened 2020-04-04
Jitsi: Rename plugin folder, add colours, check for json

brknkfr:


plugins/jitsi/jitsi_videobridge
```diff
@@ -50,7 +50,7 @@ case $1 in
     autoconf)
         if [ -x /usr/bin/curl ]; then
             if [ -x /usr/bin/jq ]; then
-	        curl -s -f -m 2 -I "$COLIBRI_URL" > /dev/null && echo "yes" && exit 0 || echo "no (no valid env.url)" && exit 0
+	        curl -s -f -m 2 -I "$COLIBRI_URL" | grep "Content-Type: application/json" > /dev/null && echo "yes" && exit 0 || echo "no (no valid env.url)" && exit 0
```
sumpfralle, line 53: That is a good approach!

Regarding `grep` and the check for existence: you can use `grep -q` without any redirects for this purpose.

btw: in the context of autoconf I would suggest to change the wording of *no valid env.url* to something like *invalid or empty response from colibri API ($COLIBRI_URL)*

plugins/jitsi/jitsi_videobridge
```diff
@@ -68,23 +68,28 @@ case $1 in
         echo "jitsi_videostreams.info current number of videostreams"
         echo "jitsi_videostreams.min 0"
         echo "jitsi_videostreams.draw AREA"
+        echo "jitsi_videostreams.colour 80c9ff"
```
sumpfralle, line 71: Personally I would prefer to stick to the default palette, unless there are specific good reasons (e.g. using *red* for the *bad* events). This allows users to specify their own global color palette, if they want to.

But again: this is a matter of taste - feel free to ignore it.

plugins/jitsi/jitsi_videobridge
brknkfr: Good suggestion ... Thank you.

plugins/jitsi/jitsi_videobridge
brknkfr: It _is_ a matter of taste. Do I find somewhere how munin automatically/automagically chooses a palette? Because in my case, the participants and conferences line weren't really readable ...

sumpfralle: > Because in my case, the participants and conferences line weren't really readable [..]

Just wild guessing: maybe your color perception differs from the color perception of other people? Personally my color reception for red is significantly weaker compared to the average population, thus I force myself to refrain from any discussion regarding visual compositions.

Here are some references regarding munin's color palettes:
* [configuring a custom color palette](http://guide.munin-monitoring.org/en/latest/reference/munin.conf.html?highlight=palette#cmdoption-munin-conf-arg-palette)
* [pre-defined color palette used by default](https://github.com/munin-monitoring/munin/blob/stable-2.0/master/lib/Munin/Master/GraphOld.pm#L133)

In the end the specific implementation of the plugin is up to you as its author.
But if you have a strong feeling, that munin's default color palette contains indistinguishable colors, then we should improve these instead of manually overriding colors in every plugin.

niclan: The colors are as you see a fixed palette. The fixed palette was chosen for the distance in the color wheel to optimize contrast.

brknkfr: Now everything is ok with the colours. There must have been some caching issue with generating the graphs.

sumpfralle:
Thanks!

PR #1056, opened 2020-04-04
Add nextcloud_ plugin

brknkfr:
A simple nextcloud multigraph plugin.

### DIFF
--- a/plugins/nextcloud/nextcloud_
+++ b/plugins/nextcloud/nextcloud_
@@ -180,7 +180,7 @@ num_storages_home.min 0
 num_storages_other.label number of other storages
 num_storages_other.info current number of other storages
 num_storages_other.min 0
-multigraph nextcloud_app_$CLEANDOMAIN
+multigraph nextcloud_apps_$CLEANDOMAIN
 graph_title Nextcloud apps on $DOMAIN
 graph_args --base 1000 -l 0
 graph_printf %.0lf
@@ -231,5 +231,5 @@ echo "multigraph nextcloud_dbsize_$CLEANDOMAIN"
 echo "db_size.value $DBSIZE"
 
 # apps
-echo "multigraph nextcloud_app_updates_$CLEANDOMAIN"
+echo "multigraph nextcloud_apps_$CLEANDOMAIN"
 print_json_data "$APPS" num_installed num_updates_available
--- a/plugins/nextcloud/nextcloud_
+++ b/plugins/nextcloud/nextcloud_
@@ -1,0 +1,235 @@
+#!/bin/sh
+
+set -e
+
+: << =cut
+
+=head1 NAME
+
+nextcloud_ - Monitor usage of nextcloud instances
+
+=head1 APPLICABLE SYSTEMS
+
+Nexcloud instances
+
+=head1 CONFIGURATION
+
+Requires installed curl and jq, a command-line json processor.
+
+This is a wildcard plugin. To monitor a nextcloud instance, link
+nextcloud_<nextcloud-domain> to this file. You can even append a port 
+(:8443) to the file if needed. For example,
+
+  ln -s /usr/share/munin/plugins/nextcloud_ \
+        /etc/munin/plugins/nextcloud_cloud.domain.tld
+
+Set username and password in your munin-node configuration
+
+[nextcloud_cloud.domain.tld]
+env.username <nexcloud_user>
+env.password <nextcloud_password>
+
+It's advised to set an app password (for this plugin) in your nextcloud 
+instance and not to use the "real" password of your nextcloud user.
+
+=head1 AUTHOR
+
+Copyright (C) 2020 Sebastian L. (https://momou.ch)
+
+=head1 LICENSE
+
+GPLv2
+
+=head1 MAGIC MARKERS
+
+ #%# family=manual
+ #%# capabilities=autoconf
+
+=cut
+
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
+
+API_PATH="/ocs/v2.php/apps/serverinfo/api/v1/info?format=json"
+DOMAIN="${0##*nextcloud_}"
+CLEANDOMAIN="$(clean_fieldname "$DOMAIN")"
+USERNAME="${username:-}"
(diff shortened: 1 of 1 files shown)


plugins/nextcloud/nextcloud_
```diff
@@ -0,0 +1,230 @@
...
+
+API_PATH="/ocs/v2.php/apps/serverinfo/api/v1/info?format=json"
+DOMAIN="${0##*nextcloud_}"
+CLEANDOMAIN="$(clean_fieldname "$DOMAIN")"
+USERNAME="${username:-}"
+PASSWORD="${password:-}"
+
+print_json_data() {
+    [ -z "$1" ] && exit 0
+    for KEY in "${@:2}"; do
```
sumpfralle, line 60: I think, this is the only place where you use a bash-only feature. Maybe assign `$1` to a local variable and add `shift` in order to void this and reduce the requirements to a simple shell.

sumpfralle: Thanks for changing the shell to `sh`. Now you need to add `shift` here and use the non-indexed `$@` substitution.

plugins/nextcloud/nextcloud_
```diff
@@ -0,0 +1,233 @@
...
+num_shares_room.label room shares
+num_shares_room.info current total of room shares
+num_shares_room.min 0
+num_shares_link_no_password.label link shares without password protection
+num_shares_link_no_password.info current total of link shares without password protection
+num_shares_link_no_password.min 0
+num_fed_shares_sent.label federated shares sent
+num_fed_shares_sent.info current total of federated shares sent
+num_fed_shares_sent.min 0
+num_fed_shares_received.label federated shares recieved
```
sumpfralle, line 149: `codespell` suggests `received` :)

plugins/nextcloud/nextcloud_
```diff
@@ -0,0 +1,230 @@
...
+    [ -z "$1" ] && exit 0
+    for KEY in "${@:2}"; do
+        VALUE=$(echo "$1" | jq -cr ".$KEY")
+        echo "$KEY.value $VALUE"
+   done
+}
+
+test_https() {
+    [ -z "$DOMAIN" ] && exit 0
+    curl -s -f -m 2 -I "https://$DOMAIN" > /dev/null && echo true && exit 0
```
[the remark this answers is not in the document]
brknkfr: Mhm, done that.

plugins/nextcloud/nextcloud_
```diff
@@ -0,0 +1,230 @@
...
+   config)
+
+cat << EOM
+multigraph nextcloud_users_$CLEANDOMAIN
+graph_title Nextcloud users on $DOMAIN
+graph_args --base 1000 -l 0
+graph_printf %.0lf
+graph_vlabel connected users
+graph_info number of connected user
+graph_category nextcloud
```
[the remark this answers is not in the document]
brknkfr: Check!

plugins/nextcloud/nextcloud_
brknkfr: Nah ... it was not only this. In bash grepping jq outputs was working, in posix shell I had to escape the slashes. Now - I think - it's fine.

plugins/nextcloud/nextcloud_
brknkfr: Awh. ;-)

plugins/nextcloud/nextcloud_
```diff
@@ -0,0 +1,230 @@
...
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
+
+API_PATH="/ocs/v2.php/apps/serverinfo/api/v1/info?format=json"
+DOMAIN="${0##*nextcloud_}"
+CLEANDOMAIN="$(clean_fieldname "$DOMAIN")"
+USERNAME="${username:-}"
+PASSWORD="${password:-}"
+
+print_json_data() {
+    [ -z "$1" ] && exit 0
```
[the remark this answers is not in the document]
brknkfr: I don't know how to specify local variables in functions in posix shell, but I just used a simple (exposed) variable, which works.

plugins/nextcloud/nextcloud_
sumpfralle: > In bash grepping jq outputs was working, in posix shell I had to escape the slashes.

That sounds really strange. I am not aware of differences between the shells regarding pipelined data. I have the vague feeling, that your were stumbling upon something else. Maybe you are curious to investigate that? I would enjoy to hear the details ...

Anyway, I am merging your contribution now. Thank you for going through all the details!

sumpfralle:
Oh - I just noticed, that your commits are a bit mixed with your jitsi plugin commits.

I would recommend, that you run `git rebase -i 87fa84b6b~` and squash the first two commits.

plugins/nextcloud/nextcloud_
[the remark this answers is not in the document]
sumpfralle: > I don't know how to specify local variables in functions in posix shell

I guess, `shellcheck` is confusing you here. Expecting `dash` as the shell implementation is reasonable (at least we do this in munin). Thus you should run `shellcheck -s dash` - this would void any warnings for the simple line `local KEY VALUE`.

> [..] but I just used a simple (exposed) variable, which works.

Yes it works, until it breaks by accident :)

Quoting the python people here (`python3 -m this | grep -i names`):
```
Namespaces are one honking great idea -- let's do more of those!
```

Anyway - since it works, it can be considered a matter of taste ...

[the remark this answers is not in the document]
brknkfr: I like the "taste" ...

plugins/nextcloud/nextcloud_
brknkfr: Yes, it is strange.

I tried a lot of stuff and I really don't know where "sh" is stumbling about where "bash" is not with this jq ... It was more trial'n'error that I found out that escaping the slahses helps ... I'll dig further ...

sumpfralle:
Sorry, I overlooked your most recent commits for a while ...

Thanks for your contribution!

PR #1057, opened 2020-04-09
Split videostreams in separate graph for better readability

brknkfr:


sumpfralle:
That looks like a good change - thank you!

PR #1062, opened 2020-04-17
Improved docker_ performance when getting CPU/Memory stats

rwky:
Added parallel processing using python's muliprocessing module
to fork a new process each for each docker container when gathering
stats.

This vastly improves performance since using a for loop the script
blocks until the stats are returned.

When I ran the docker_cpu plugin without this change it would timeout. I have 43 containers running on the host in question and it would take about 90 seconds. With this improvement it takes 2-3.

### DIFF
--- a/plugins/docker/docker_
+++ b/plugins/docker/docker_
@@ -50,6 +50,7 @@ https://docs.docker.com/compose/reference/envvars/):
 import os
 import sys
 import docker
+from multiprocessing import Process, Queue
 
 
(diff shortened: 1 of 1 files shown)


sumpfralle:
I was first a bit surprised by the approach you take, but the example numbers you provided are a very good argument for your approach.

Thank you for your elegant and helpful contribution!

PS: I think, nowadays I would do this via `async / await`. But both ways are fine ...

PR #1063, opened 2020-04-17
Fixed warnings generated by total_by_process_

rwky:
Fixed the following:

* Argument "config" isn't numeric in numeric eq (==) at line 89.
* Use of uninitialized value $ARGV[1] in string eq at line 35


sumpfralle:
Thanks!

PR #1070, opened 2020-04-22
Fix geowebcache-bandwidth

rmaziere:


sumpfralle:
Thank you for fixing these plugins!

btw: due to the similarity of the fixes and the plugins, you can easily submit these in a single pull request in the future. Just a matter of taste ...

PR #1076, opened 2020-04-26
Fixed autoconf on http_request_time

rwky:
Previously it would error since the $url variable is
the array key isn't the url to check it's a santized version


sumpfralle:
Thank you!

PR #1077, opened 2020-04-29
Change regex to match "prepend" pattern in postfix/policyd-spf-python plugin

codeurimpulsif:
By default `policyd-spf-python` generate this output in `/var/log.mail.info`:
```
Apr 29 14:14:22 localhost policyd-spf[25915]: prepend Received-SPF: Pass (mailfrom) identity=mailfrom; client-ip=REDACTED; helo=example.com; envelope-from=example@example.com; receiver=<UNKNOWN>
```
This regex change can match also when `prepend Received-SPF` is present.

### DIFF
--- a/plugins/postfix/policyd-spf-python
+++ b/plugins/postfix/policyd-spf-python
@@ -69,7 +69,7 @@ fi
 #
 
 function get_log_count() {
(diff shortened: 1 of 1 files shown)


sumpfralle:
Looks good to me - thanks!

PR #1078, opened 2020-05-01
currency/bitcoin: fix graph category name typo

Talkless:
Fix "htc" typo to "btc".

### DIFF
--- a/plugins/currency/bitcoin/bitcoind_
+++ b/plugins/currency/bitcoin/bitcoind_
@@ -125,7 +125,7 @@ def main():
         return True
 
     if command == 'config':
-        print('graph_category htc')
+        print('graph_category btc')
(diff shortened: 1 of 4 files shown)


sumpfralle:
Thank you for caring for this plugin!

Indeed the category `htc` seems to be the proper one - it refers to *High-throughput computing* (see the [list of well-known categories](http://guide.munin-monitoring.org/en/latest/reference/graph-category.html#well-known-categories).

Talkless:
> it refers to High-throughput computing

Thanks for explanation. I thought these where free-form.

sumpfralle:
> I thought these where free-form.

Yes, they are just text fields, but we (the maintainers of the contrib repository) try to encourage people to use a small set of generic categories (instead of a wide range of application-specific ones). This reduces clutter in Munin's web visualization.

PR #1079, opened 2020-05-01
Added php_fpm_status plugin

rwky:
This is a fork of the php5-fpm_status plugin.

Changes from in the fork:

* Supports multiple pools using environment variables to specify urls
* Supports wget and curl
* Added GPLv3 License


plugins/php/php_fpm_status
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
```
sumpfralle, line 1: I think, you are not using any bash-specific features. Thus it would be great, if you could switch to `sh` instead.

plugins/php/php_fpm_status
```diff
@@ -0,0 +1,85 @@
...
+# 	Plugin to monitor php-fpm process manager, php-fpm 5.3.3 is required
```
sumpfralle, line 2: Please reformat the documentation into the perlpod format (look for `=head1` in other plugins).

And please clarify the version requirement: is it *exactly* v5.3.3 or are later versions supported, too?

plugins/php/php_fpm_status
```diff
@@ -0,0 +1,85 @@
...
+#  [php_fpm_status_www]
+#  env.url http://127.0.0.1/fpm-status-www
+#  [php_fpm_status_www2]
+#  env.url http://127.0.0.1/fpm-status-www2
+# 
+#%# family=contrib
+
+url=${url:-'https://127.0.0.1/fpm-status'}
+
+if [ -z "$url" ]
```
sumpfralle, line 41: It would be great, if you could convert this into an `autoconf` test.

plugins/php/php_fpm_status
```diff
@@ -0,0 +1,85 @@
...
+then
+  echo "env.url not specified" >&2
+  exit 1
+fi
+if which wget > /dev/null 2>&1
+then
+  wget -q -O "$MUNIN_STATEFILE" "$url"
```
sumpfralle, line 48: You could easily avoid using a temporary file here, by assigning the output to a variable (`foo=$(wget -O - ...)`).

sumpfralle (commented):

Please take a look at my questions below.

As far as I understand your plugin, it would replace the functionality of the existing `php5-fpm_status` plugin. Is this correct? In this case, I would suggest, that you remove the other plugin.

rwky:
All changes made, @sumpfralle let me know if you want anything else tweaked.

sumpfralle:
It looks good to me - thank you!

PR #1081, opened 2020-05-07
Fix to work with old lsblk version...

rantal:


sumpfralle:
Sorry for letting this rot for so long.

PR #1084, opened 2020-05-25
ehterpadlite_: Monitor number of pads

brknkfr:
Multigraph plugin to monitor number of pads in a etherpad-lite instance.


sumpfralle:
Your plugin looks very good to me. Thank you!

PR #1085, opened 2020-05-26
Plugin Timesync_status fixies

onovy:


plugins/systemd/timesync_status
```diff
@@ -53,6 +53,9 @@ def parse_time(value):
     if ' ' in value:
         return sum(parse_time(x) for x in value.split(' '))
 
+    if value == "0":
```
sumpfralle, line 56: Under which circumstances would this value appear?

Would it be more appropriate to return `None` or raise an exception in this case?

plugins/systemd/timesync_status
```diff
@@ -87,11 +90,14 @@ def retrieve():
...
-    print('jitter.value', parse_time(values['Jitter']))
-    print('poll.value', parse_time(values['Poll interval'].split('(')[0]))
+    if 'Offset' in values:
+        print('offset.value', parse_time(values['Offset']))
+        print('delay.value', parse_time(values['Delay']))
+        print('delay.extinfo', 'Server', values['Server'])
+        print('jitter.value', parse_time(values['Jitter']))
+        print('poll.value', parse_time(values['Poll interval'].split('(')[0]))
+    else:
+        print('delay.value -1')
```
sumpfralle, line 100: Is `-1` a suitable value here? How about using `U` (*unknown*) for *all* fields instead?

(a quick comment, which condition leads to this branch would be perfect!)

sumpfralle (commented):
Sorry for letting your contribution rot for so long :(

Your change looks useful.

Please take a look at my two questions below.

plugins/systemd/timesync_status
onovy: If time is exactly zero (for example Jitter), there is no unit (suffix) and re.match will fail. If systemd says it's zero, it means zero, not errror. :)

plugins/systemd/timesync_status
onovy: right, working on it

plugins/systemd/timesync_status
onovy: example:
```
$ timedatectl timesync-status
       Server: #REDACTED#
Poll interval: 1min 4s (min: 32s; max 34min 8s) 
         Leap: normal
      Version: 4
      Stratum: 2
    Reference: C37190EE
    Precision: 62us (-14)
Root distance: 67.641ms (max: 5s) 
       Offset: +191us
        Delay: 537us
       Jitter: 0
 Packet count: 1
    Frequency: +3.394ppm

sumpfralle:

PR #1086, opened 2020-05-27
[plugins/transmission] Fixups for consistency

shtrom:
Use standard colours for the transmission_activity plot (but keep a
darker one for paused torrents).

Also, remove the ${host} from the graph_title, and add it in the
graph_info, for better groupability.


sumpfralle:
Thank you!

(sorry for letting this rot for so long ...)

PR #1088, opened 2020-05-31
ssl-certificate-expiry: Feature added: checking intermediate certs as well

nitram2342:
The validity of a x509 cert does not only depend on the server certificate, which is usually the last cert in a chain. Instead, the validity of a cert also depends on the validity of each cert in a chain. When an intermediate cert expires, the server cert is also not valid anymore, even if the server cert has some days left. 

When you use the ssl-certificate-expiry plugin and missed to update an intermediate cert in server deployment, the plugin  may calculate a wrong expiry time when an intermediate cert expires. If you run your own CA, missing this is more likely than for a fully scripted let's-encrypt-solution. Hence, I added the feature to also check the other certs that are sent by a server.


nitram2342:
I resolved the merge conflict and the Travis CI seems to be fine now.

plugins/ssl/ssl-certificate-expiry
```diff
@@ -123,10 +126,33 @@ print_expire_days() {
     [ -n "$starttls" ] && s_client_args="-starttls $starttls"
 
     # shellcheck disable=SC2086
```
sumpfralle, line 128: I guess, you need to move this line below your comment - otherwise `shellcheck -s dash PLUGIN_FILENAME` will complain. It may look like a comment, but it is a line interpreted by shellcheck.

plugins/ssl/ssl-certificate-expiry
```diff
@@ -123,10 +126,33 @@ print_expire_days() {
...
+
+    # The original code only extracted the server certificate,
```
sumpfralle, line 130: Feel free to omit the reference to the previous code. The future reader of this plugin's code is probably not aware of its history ...

plugins/ssl/ssl-certificate-expiry
```diff
@@ -123,10 +126,33 @@ print_expire_days() {
...
-            $s_client_args 2>/dev/null \
-        | parse_valid_days_from_certificate
+	-servername "$host" -connect "${host}:${port}" \
+	-showcerts \
+	$s_client_args 2>/dev/null | \
+	awk '{
+  	  if ($0 == "-----BEGIN CERTIFICATE-----") cert=""
+  	  else if ($0 == "-----END CERTIFICATE-----") print cert
+  	  else cert=cert$0
+	  }' | ( \
```
sumpfralle, line 150: I think, you can omit the sub-shell (via `( ... )`) here, since the `while ... done` construct is treated as a single command.

plugins/ssl/ssl-certificate-expiry
```diff
@@ -123,10 +126,33 @@ print_expire_days() {
...
+	while read -r CERT; do
```
sumpfralle, line 151: Personally I prefer to indent continued command chains, thus I would move `while` a bit to the right.

In this specific case you may even just the head of the loop to the line above.

But this is just a matter of taste. Feel free to ignore it.

plugins/ssl/ssl-certificate-expiry
```diff
@@ -123,10 +126,33 @@ print_expire_days() {
...
+	    (echo "-----BEGIN CERTIFICATE-----"; echo ${CERT}; echo "-----END CERTIFICATE-----") | \
```
sumpfralle, line 152: Maybe just use the following?
```shell
printf '-----BEGIN CERTIFICATE-----\n%s\n-----END CERTIFICATE-----' "$CERT"
```

plugins/ssl/ssl-certificate-expiry
nitram2342: You are right. That make sense.

nitram2342:
I adjusted the code according to your recommendations.

sumpfralle:
Thank you!

PR #1089, opened 2020-06-24
[plugins/solar/fronius] Monitor Solar Inverters using the Fronius Sol…

shtrom:
…ar API

This includes caching of the inverter info with a 12h TTL, so the graph
still renders when the inverter has gone to sleep at night.


plugins/solar/fronius
```diff
@@ -0,0 +1,285 @@
...
+# the response.  The first argument is a timeout (in minutes) after which the
+# cache is ignored and a new request is attempted.  If the request fails, the
+# cache is used. If the timeout is 0, the request is always attempted, using
+# the cache as a backup on failure.
+cached() {
+	timeout="${1}"
+	shift
+	fn="${1}"
+	shift
+	args=${*:-}
```
sumpfralle, line 212: Here you are discarding the tokens and risk to create problems due to whitespace.

How about just using `"$@"` below instead of `$args`?

plugins/isp/internode_usage
```diff
@@ -38,7 +38,7 @@ multiple times, by symlinking it as 'internode_usage_SERVICEID'.
 =head1 CACHING
 
 As the API is sometimes flakey, the initial service information is cached
-locally, with a one-hour lifetime, before hitting the base API again. However,
+locally, with a day's lifetime, before hitting the base API again. However,
```
sumpfralle, line 41: I assume, that this (unrelated) change somehow sneaked into your proposal accidentally.

I am fine with merging it here anyway, if you are confident, that it is indeed an improvement.

plugins/solar/fronius
```diff
@@ -0,0 +1,285 @@
...
+		api_data=$(cat "${cachefile}")
+	else
+		# shellcheck disable=SC2086
+		api_data="$("${fn}" ${args} \
+			|| true)"
+
+		if [ -n "${api_data}" ]; then
+			echo "${api_data}" > "${cachefile}"
+		else
+			api_data=$(cat "${cachefile}")
```
sumpfralle, line 228: This may fail, if the cache file was never created, or?

plugins/solar/fronius
```diff
@@ -0,0 +1,285 @@
...
+CURL_ARGS="-s --connect-timeout ${CONNECT_TIMEOUT}"
+fetch() {
+	# shellcheck disable=SC2086
+	curl -f ${CURL_ARGS} "$@" \
+		|| { echo "error fetching ${*}" >&2; false; }
+}
+
+get_inverter_info() {
+	fetch "${INVERTER_BASE_URL}/solar_api/v1/GetInverterInfo.cgi" \
+	| recode html..ascii
```
sumpfralle, line 194: Maybe indent the second line?

plugins/isp/internode_usage
shtrom: It's somewhat related, as I used the caching code as a base, and realised the doc was incorrect, so I fixed it at the same time (:

plugins/solar/fronius
shtrom: Correct, but there's not much we can do then. Do you have a suggestion about a better way to handle this? At the moment, the plugin just dies until it has obtained data at least once.

shtrom:
> regarding shebang 

fixed :+1: 

> and check_deps

similar issue to #1090, in this case the hostname of the inverter needs to be configured before the plugin can work

shtrom:
@sumpfralle this one has been squashed, too, and ready to merge

sumpfralle:
Thank you!

Just as an afterthought (feel free to ignore it): you may want to add some `local`s and maybe improve the formatting of the helpful API examples by using a *here document* (```true <<EOF ... EOF```). Anyway: it is a good plugin!

PR #1092, opened 2020-07-03
Modified http_request_time to allow specifying a user agent

rwky:
Similar to other settings for this plugin, there is a new
env.urlN_agent setting which allows specifying a user agent


sumpfralle:
Thank you!

PR #1093, opened 2020-07-06
murmur_users PHP 7 support

Gussi:
`split` was deprecated in PHP 5.3.0 and removed in PHP 7.0.0, `explode`
is a drop-in replacement.


sumpfralle:
Thanks!

PR #1094, opened 2020-07-07
[plugin/docker_] Improvements to docker_ plugin

shtrom:
* Add support for docker_network
* Actually show non running containers counts
* Show intermediate and dangling images counts
* Add warnings on dangling images and dead containers
* Fully support autoconf and suggest
* Add graph_total to the status (thereby making the containers graph
  redundant), CPU and Memory graphs
* Show most graphs as AREASTACK, this allows to get a clearer view of
  both individual and total use
* Use LINESTACK1 to work around munin-contrib/munin#1343
* Include all containers in CPU/Memory/Network config, so data from
  non-running containers is still displayed
* Sprinkle some info and extinfo
* Add ClientWrapper around docker module, providing caching and sorting
* Reverse-engineer author list from git log

### DIFF
--- a/plugins/docker/docker_
+++ b/plugins/docker/docker_
@@ -76,7 +76,6 @@ This section has been reverse-engineered from git logs
 
 import os
 import sys
-import docker
 import re
 from functools import cached_property
 from multiprocessing import Process, Queue
@@ -109,7 +108,7 @@ class ClientWrapper:
     def __init__(self, client, exclude_re=None):
         self.client = client
         if exclude_re:
-            self.exclude = re.compile(exclude)
+            self.exclude = re.compile(exclude_re)
 
     @cached_property
     @sorted_by_creation_date
@@ -121,7 +120,7 @@ class ClientWrapper:
     def all_containers(self):
         return [c for c in self.client.containers.list(all=True)
                 if not self.exclude
-                or not self.exclude.search(exclude, c.name)]
+                or not self.exclude.search(c.name)]
 
     @cached_property
     @sorted_by_creation_date
--- a/plugins/docker/docker_
+++ b/plugins/docker/docker_
@@ -5,7 +5,7 @@
 docker_ - Docker wildcard-plugin to monitor a L<Docker|https://www.docker.com> host.
 
 This wildcard plugin provides at the moment only the suffixes C<containers>, C<images>, C<status>,
-C<volumes>, C<cpu> and C<memory>.
+C<volumes>, C<cpu>, C<memory> and C<network>.
 
 =head1 INSTALLATION
 
@@ -18,6 +18,7 @@ C<volumes>, C<cpu> and C<memory>.
     ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_cpu
     ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_images
     ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_memory
+    ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_network
     ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_status
     ln -s /usr/share/munin/plugins/docker_ /etc/munin/plugins/docker_volumes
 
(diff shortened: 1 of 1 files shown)


plugins/docker/docker_
```diff
@@ -54,35 +55,60 @@ from multiprocessing import Process, Queue
...
-    dead = 0
-    for container in client.containers.list():
+    running = []
+    paused = []
+    created = []
+    restarting = []
+    removing = []
+    exited = []
+    dead = []
+    for container in client.containers.list(all=True):
```
sumpfralle, line 65: Maybe `sorted()` around the container list would make sure, that the list of hosts always uses the same order?

plugins/docker/docker_
```diff
@@ -54,35 +55,60 @@ from multiprocessing import Process, Queue
...
+    print('created.extinfo', ', '.join(created))
+    print('restarting.value', len(restarting))
+    print('restarting.extinfo', ', '.join(restarting))
+    print('removing.value', len(removing))
+    print('removing.extinfo', ', '.join(removing))
+    print('exited.value', len(exited))
+    print('exited.extinfo', ', '.join(exited))
+    print('dead.value', len(dead))
+    print('dead.extinfo', ', '.join(dead))
+
```
sumpfralle, line 94: I think, `flake8` would appreciate two empty lines here :)

plugins/docker/docker_
```diff
@@ -54,35 +55,60 @@ from multiprocessing import Process, Queue
...
+def str_image(image):
+    return f"{image.short_id} ({round(image.attrs['Size']/1024**2, 2)} MiB)"
+
+def print_images_count(client):
+    images = client.images.list()
+    intermediate = [ i for i in client.images.list(all=True)
```
sumpfralle, line 100: Maybe `sorted` would be nice here, too?

plugins/docker/docker_
```diff
@@ -54,35 +55,60 @@ from multiprocessing import Process, Queue
...
+                     if i not in images ]
+    dangling = client.images.list(filters={'dangling':True})
+    images = [ i for i in images
```
sumpfralle, line 103: As a happy user of sets, I would suggest `images = set(images).difference(set(dangling))`. But this is just a matter of taste :)

plugins/docker/docker_
```diff
@@ -54,35 +55,60 @@ from multiprocessing import Process, Queue
...
+               if i not in dangling ]
+
+    print('intermediate_quantity.value', len(intermediate))
+    print('intermediate_quantity.extinfo', ', '.join([str_image(i) for i in intermediate]))
```
sumpfralle, line 107: Style nitpick: you can remove `[...]` from the list comprehension and thus turning it into a generator expression. Again, just a matter of taste ...

plugins/docker/docker_
```diff
@@ -186,8 +243,10 @@ def main():
             print("graph_vlabel CPU usage (%)")
             print("graph_category virtualization")
             print("graph_info This graph shows docker container CPU usage.")
-            for container in client.containers.list():
+            print("graph_total Total CPU usage")
+            for container in client.containers.list(all=True):
```
sumpfralle, line 247: Maybe `sorted` would be helpful here, too. But maybe the `list` method already produces a stable order - I do not know.

plugins/docker/docker_
```diff
@@ -197,10 +256,34 @@ def main():
             print("graph_vlabel Bytes")
             print("graph_category virtualization")
             print("graph_info This graph shows docker container memory usage.")
-            for container in client.containers.list():
+            print("graph_total Total memory usage")
+            for container in client.containers.list(all=True):
```
sumpfralle, line 260: Maybe add `sorted`.

plugins/docker/docker_
```diff
@@ -197,10 +256,34 @@ def main():
...
             print_containers_memory(client)
+    elif wildcard == "network":
+        if mode == "config":
+            print("graph_title Docker containers network usage")
+            print("graph_args --base 1024 -l 0")
+            print("graph_vlabel bits in (-) / out (+) per ${graph_period}")
+            print("graph_category virtualization")
+            print("graph_info This graph shows docker container network usage.")
+            print("graph_total Total network usage")
+            for container in client.containers.list(all=True):
```
sumpfralle, line 273: `sorted`?

plugins/docker/docker_
shtrom: I ended up doing it in the caching wrapper, using a decorator.

shtrom:
@sumpfralle ready for another look

plugins/docker/docker_
```diff
@@ -107,18 +243,39 @@ def print_containers_cpu(client):
     for container, stats in parallel_container_stats(client):
         cpu_count = len(stats["cpu_stats"]["cpu_usage"]["percpu_usage"])
         cpu_percent = 0.0
-        cpu_delta = float(stats["cpu_stats"]["cpu_usage"]["total_usage"]) \
-            - float(stats["precpu_stats"]["cpu_usage"]["total_usage"])
-        system_delta = float(stats["cpu_stats"]["system_cpu_usage"]) \
-            - float(stats["precpu_stats"]["system_cpu_usage"])
+        cpu_delta = float(stats["cpu_stats"]["cpu_usage"]["total_usage"])
+        - float(stats["precpu_stats"]["cpu_usage"]["total_usage"])
```
sumpfralle, line 247: I think, you need to add parenthesis around the two lines - otherwise the second line is just interpreted as a (negative) number without any impact on the calculation above.

plugins/docker/docker_
```diff
@@ -45,44 +46,179 @@ https://docs.docker.com/compose/reference/envvars/):
...
 import sys
-import docker
+from functools import cached_property
 from multiprocessing import Process, Queue
 
+# Early autoconf-style detection
+try:
+    import docker
+except ImportError as e:
+    print(f'no ({e})')
```
sumpfralle, line 75: I think, you can *translate* the pythonic error message here into a user friendly text: `no (failed to import python module 'docker')`.

plugins/docker/docker_
```diff
@@ -45,44 +46,179 @@ https://docs.docker.com/compose/reference/envvars/):
...
+    sys.exit(1)
```
sumpfralle, line 76: The autoconf specification says, that even in case of `no`, there should be a zero-exitcode.

But the non-zero-exitcode should be emitted in all other (non-`autoconf`) cases, of course.

Thus you should probably just set a `docker_import_success` variable here and use it for a proper `autoconf` later.

plugins/docker/docker_
```diff
@@ -45,44 +46,179 @@ https://docs.docker.com/compose/reference/envvars/):
...
+
+
+def sorted_by_creation_date(func):
```
sumpfralle, line 79: Is this a good order of things?

After a reboot of the hypervisor machine, this would probably change the order of items (and their color) in the graph, or?

Maybe use the name instead?

(but feel free to follow your own judgement here - I cannot really tell)

plugins/docker/docker_
shtrom: Ah, right! Yeah, I thought it was weird that it deindented it... Somehow I lost the `\`, but parens will be less brittle.

plugins/docker/docker_
shtrom: I could, but I went down a different path now, doing all the autoconf checks at once, so it's more difficult to do now.

plugins/docker/docker_
shtrom: The name is generally random. The creation date doesn't change for the lifetime of the container.

The colours will always be a bit unstable, as they will change anytime you delete an old container, but at least, in this way, containers only get added at the end, rather than in between based on their names.

shtrom:
@sumpfralle comments addressed and/or replied to, and rebase on the latest master, fixing conflicts with the new exclusion RE. Ready for re-review!

plugins/docker/docker_
```diff
@@ -174,23 +372,42 @@ def main():
...
-            print('images_quantity.value', len(client.images.list()))
+            print_images_count(client)
     elif wildcard == "volumes":
         if mode == "config":
             print("graph_title Docker volumes")
             print("graph_vlabel volumes")
             print("graph_category virtualization")
             print("volumes_quantity.label Volumes")
+            print("volumes_quantity.draw AREASTACK")
+            print("volumes_quantity.info Unused volumes can be deleted with "
```
sumpfralle, line 406: Here you are using *two* spaces (instead of one) due to your line breaking.

But this is very minor and I would have omitted to comment on it, if there would not be a [style-check failure](https://travis-ci.org/github/munin-monitoring/contrib/builds/720628638) :)

shtrom: Hum, I don't see the added space here. 

However, there were a few other test failures that were valid, so I fixed those (:

sumpfralle:
Thanks for your fixes!

Regarding the whitespace: I am quite confident, that double spaces appear in some places:
```shell
$ printf '%s\n' 'print("foo "' '    " bar")' | python3
foo  bar
```

Anyway: this is such a minor detail, that it does not deserve a change ...

Thanks!

shtrom:
On 26 August 2020 12:19:58 am AEST, Lars Kruse <notifications@github.com> wrote:
>Regarding the whitespace: I am quite confident, that double spaces
>appear in some places:
>```shell
>$ printf '%s\n' 'print("foo "' '    " bar")' | python3
>foo  bar
>```
Yeah, it would in those cases, but I couldn't find the occurence there.
-- 
Sent from my mobile. Please excuse my brevity.

sumpfralle:
> [..] but I couldn't find the occurence there.

Oh sorry - after taking a closer look again, I finally noticed the backticks, that I misinterpreted as whitespace. I would like to blame the font or my eyes - but it was just me :(

Sorry for the noise!

PR #1096, opened 2020-07-16
hwmon: recognize frequencies

antonv6:
Discovered thanks to an RX 570, which can show sclk and mclk via /sys/class/hwmon.


antonv6:
For reference: https://www.kernel.org/doc/html/latest/gpu/amdgpu.html#gpu-power-thermal-controls-and-monitoring

sumpfralle:
Thanks for your contribution!

PR #1097, opened 2020-07-20
Improve kvm_* plugins

zivillian:
This fixes the label (vm names) for vms startet via kvm. This was tested on a host running Proxmox VE. In addition I've removed the draw parameter to unify the graphs and removed the duplicate graph from kvm_io and kvm_net.


sumpfralle:

I could imagine, that the label convention, that you are adding here, is only suitable for proxmox, or?
In this case, I would appreciate, if you would add a very short comment above these lines mentioning proxmox.

Thank you!

zivillian:
done

sumpfralle:
Thank you!

PR #1098, opened 2020-07-22
[plugins/snmp__cisco_] SNMP plugin to query arbitrary SNMP MIBs on Ci…

shtrom:
…sco (and maybe other) equipment

### DIFF
--- a/plugins/router/snmp__cisco_
+++ b/plugins/router/snmp__cisco_
@@ -1,0 +1,241 @@
+#!/usr/bin/perl -w
+# -*- perl -*-
+# vim: ft=perl
+
+=head1 NAME
+
+snmp__cisco_ - Munin plugin to monitor arbitrary SNMP MIBs on Cisco Switches.
+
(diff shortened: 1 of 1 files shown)


sumpfralle:
The new plugin looks good to me - thank you!

Could you please verify, whether these OIDs are standardized? In this case I would prefer a different plugin name.

shtrom:
On 15 August 2020 8:37:36 am AEST, Lars Kruse <notifications@github.com> wrote:
>The new plugin looks good to me - thank you!
>
>Could you please verify, whether these OIDs are standardized? In this
>case I would prefer a different plugin name.

They are not, unfortunately. They are specifically in the Cisco Envmon MIBs. Arguably, though, we could extend to any other, standard or custom, MIBs, but this would require modifying the plugin, so I guess a rename would make sense if/when this happens, but probably not for now.
-- 
Sent from my mobile. Please excuse my brevity.

shtrom:
@sumpfralle @steveschnepp AFAIK, this PR is good to go.

PR #1099, opened 2020-07-23
docker_cpu: fix data for container names starting with numbers

ap-wtioit:
This fixes an issue with docker_cpu when your container names start with integers (e.g. because you found it a good idea to prefix them with the CI_PIPELINE_ID ;-) )

munin-node output
```
155222_110_testyaml_nginx_1.cdef 155222_110_testyaml_nginx_1,1000000000,/
```

gets transfered to when calling rrd
```
...
DEF:a06259ff28d9f69b=/var/lib/munin/redacted/redacted-docker_cpu-_55222_110_testyaml_nginx_1-d.rrd:42:MAX
DEF:i06259ff28d9f69b=/var/lib/munin/redacted/redacted-docker_cpu-_55222_110_testyaml_nginx_1-d.rrd:42:MIN
DEF:g06259ff28d9f69b=/var/lib/munin/redacted/redacted-docker_cpu-_55222_110_testyaml_nginx_1-d.rrd:42:AVERAGE
...
CDEF:a27dc539c2004686=155222_110_testyaml_nginx_1,1000000000,/
CDEF:i27dc539c2004686=155222_110_testyaml_nginx_1,1000000000,/
CDEF:g27dc539c2004686=155222_110_testyaml_nginx_1,1000000000,/
...
```

and causes an error there (logged in munin-graph.log):
```
2020/07/23 15:18:03 [WARNING] Could not draw graph "/var/cache/munin/www/redacted/redacted/docker_cpu-day.png": /var/cache/munin/www/redacted/redacted/docker_cpu-day.png
2020/07/23 15:18:03 [RRD ERROR] Unable to graph /var/cache/munin/www/redacted/redacted/docker_cpu-year.png : don't understand '155262_110_testyaml_nginx_1,1000000000,/'
2020/07/23 15:18:03 [RRD ERROR] rrdtool 'graph' '/var/cache/munin/www/redacted/redacted/docker_cpu-year.png' \
        '--title' \
        'Docker container CPU usage - by year' \
        '--start' \
        '-400d' \
        '--base' \
        '1000' \
        '-r' \
        '--lower-limit' \
```

this looks like a munin(-graph?) bug (not replacing the name in CDEF correctly when it started with a number originally), but i guess it's best to not start the names with numbers (and i have no clue how to fix it in munin).

Info @wt-io-it

### DIFF
--- a/plugins/docker/docker_cpu
+++ b/plugins/docker/docker_cpu
@@ -62,10 +62,13 @@ for my $i (1 .. $#containers)
    my @fields = split / +/, $containers[$i];
    my $id = $fields[0];
    my $name = $fields[$#fields];
+   my $label = $name;
    # manage container name containing arithmetic operators and dots. E.g, my-container.
    $name =~ s/[-\+*\/\.]/_/g;
    # truncate container name with "," character.
    $name =~ s/,.*//g;
+   # prefix if container starts with 0-9
+   $name =~ s/^([0-9])/c$1/;
    if (open(my $file, '<', "/sys/fs/cgroup/cpuacct/docker/$id/cpuacct.usage"))
    {
       my $total_cpu_ns = <$file>;
@@ -75,7 +78,7 @@ for my $i (1 .. $#containers)
       {
          my @ncpu = split / /, <$file>;
          close $file;
-         push @result, {'name'=>$name, 'total_cpu_ns'=>$total_cpu_ns, 'ncpu'=>$#ncpu};
+         push @result, {'name'=>$name, 'label'=>$label, 'total_cpu_ns'=>$total_cpu_ns, 'ncpu'=>$#ncpu};
       }
    }
 }
@@ -100,7 +103,7 @@ if (defined $ARGV[0] and $ARGV[0] eq "config")
 
    foreach(@result)
    {
-      print "$$_{'name'}.label $$_{'name'}\n";
+      print "$$_{'name'}.label $$_{'label'}\n";
       print "$$_{'name'}.draw LINE2\n";
       print "$$_{'name'}.min 0\n";
       print "$$_{'name'}.type DERIVE\n";
--- a/plugins/docker/docker_cpu
+++ b/plugins/docker/docker_cpu
@@ -86,9 +86,9 @@ for my $i (1 .. $#containers)
 if (defined $ARGV[0] and $ARGV[0] eq "config")
 {
    my $nanoSecondsInSecond=1000000000;
-   my $graphlimit = $result[0]{'ncpu'};
+   my $graphlimit = 1;
    foreach(@result){
-      if ($$_{'ncpu'} > $graphlimit){
+      if ($$_{'ncpu'} || 1 > $graphlimit){
          $graphlimit = $$_{'ncpu'};
       }
    }


plugins/docker/docker_cpu
```diff
@@ -100,11 +103,14 @@ if (defined $ARGV[0] and $ARGV[0] eq "config")
 
    foreach(@result)
    {
-      print "$$_{'name'}.label $$_{'name'}\n";
-      print "$$_{'name'}.draw LINE2\n";
-      print "$$_{'name'}.min 0\n";
-      print "$$_{'name'}.type DERIVE\n";
-      print "$$_{'name'}.cdef $$_{'name'},$nanoSecondsInSecond,/\n";
+      if ($$_{'name'})
```
sumpfralle, line 106: How could it happen, that this variable is empty?

ap-wtioit: that was a configuration / system problem (munin-node had no access to the necessary paths (i did not map all the docker paths into my docker-container running munin-node)), i guess it can happen on other systems as well if the process returns an error.

sumpfralle: Hm. I would feel better, if this problem could be detected when it happens, instead of cleaning in up in the output.
But if you have no idea, how this could be done in the plugin, then I would appreciate a comment above the `if`, which describes the problem, that it tries to solve.

Thank you!

ap-wtioit: I'll add the comment (in the code).

For me with munin-node plugins i always try to separate real errors from perl warnings (i find it very hard to diagnose the root cause if an error happens and there is a bunch of perl warnings for missing vars as well). so this cleans up the perl warnings when an error happens. the error is still shown but the warnings are gone. so in case of an error you now only get the error and this makes it easier for me to focus on the error and not on the warnings.

ap-wtioit: ok the warnings are even more anoying, without any error or configuration issue, this happens if there is no docker containers running (without the if):
```
bigbear3001@wt-io-it-bigbear3001:/etc/munin/plugins$ sudo munin-run docker_cpu config
graph_title Docker container CPU usage
graph_args --base 1000 -r --lower-limit 0 --upper-limit 100
graph_vlabel %
graph_scale no
graph_period second
graph_category virtualization
graph_info This graph shows docker container CPU usage.
Use of uninitialized value in concatenation (.) or string at /etc/munin/plugins/docker_cpu line 106.
Use of uninitialized value in concatenation (.) or string at /etc/munin/plugins/docker_cpu line 106.
.label 
Use of uninitialized value in concatenation (.) or string at /etc/munin/plugins/docker_cpu line 107.
.draw LINE2
Use of uninitialized value in concatenation (.) or string at /etc/munin/plugins/docker_cpu line 108.
.min 0
Use of uninitialized value in concatenation (.) or string at /etc/munin/plugins/docker_cpu line 109.
.type DERIVE
Use of uninitialized value in concatenation (.) or string at /etc/munin/plugins/docker_cpu line 110.
Use of uninitialized value in concatenation (.) or string at /etc/munin/plugins/docker_cpu line 110.
.cdef ,1000000000,/
```

ap-wtioit: found a better fix:
```
my $graphlimit = $result[0]{'ncpu'}
```
somehow initialized $result / @ result
and as there was a loop right after that, replacing it with an initial value of 1 should have the same result
```
   my $graphlimit = 1;
   foreach(@result){
      if ($$_{'ncpu'} || 1 > $graphlimit){
         $graphlimit = $$_{'ncpu'};
      }
   }
```

sorry that i didn't catch that right away (perl isn't one of my common languages) and thanks for sticking with the problem. i hope that this solution is more appropriate.

sumpfralle: I guess, I found the culprit:
```
my $graphlimit = $result[0]{'ncpu'};
```

The above line accesses a non-existing array item and turns it into a hash. Perl can be astonishingly weird :)

If you replace the line above with `my $graphlimit = 1;`, then your sanity check can be removed.

sumpfralle:
Thank you!

PR #1100, opened 2020-07-31
[plugins/systemd/systemd_units] Optionally hide extinfo of active services

PoroCYon:
These services are running fine, so the extinfo would mostly just clutter
the webui while making the extinfo of relevant (i.e. down) services
harder to find.

### DIFF
--- a/plugins/systemd/systemd_units
+++ b/plugins/systemd/systemd_units
@@ -31,6 +31,8 @@ C<grep -E>):
 
 An example configuration might be something like this:
 
+=over 2
+
  [systemd_units]
  env.failed_warning 0
  env.failed_critical 5
@@ -38,10 +40,15 @@ An example configuration might be something like this:
  env.inactive_critical 20
  env.exclude boring
  env.inactive_exclude sleepy
+ env.silence_active_extinfo 1
+
+=back
 
 In the example above, we've overridden the default warning and critical levels for failed units, added warning
 and critical levels for inactive units, then filtered out boring units from all results and filtered out sleepy
-units from results for the inactive state.
+units from results for the inactive state. In addition to that, only more extensive info of non-active units
+are displayed in order to quickly see which units are failing and why in the webui. (By default, all extra info
+about all units is displayed.)
 
 =head1 AUTHOR
 
@@ -63,6 +70,7 @@ GPLv2
 
 failed_warning="${failed_warning:-0}"
 failed_critical="${failed_critical:-10}"
+silence_active_extinfo="${silence_active_extinfo:-0}"
 
 states="active \
 	reloading \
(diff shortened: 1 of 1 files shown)


plugins/systemd/systemd_units
```diff
@@ -123,8 +125,12 @@ fetch () {
 		count=$(echo "$state_unit_list" | grep -c "$state$")
 		echo "$state.value $count"
 		extinfo=$(echo "$state_unit_list" | grep "$state$" | cut -d " " -f 1 | tr '\n' ' ')
-		if [ -n "$extinfo" ]; then
-			echo "$state.extinfo" "$extinfo"
+		if [ -z "$silence_active_extinfo" ]; then
```
sumpfralle, line 128: I guess, you should use `[ "${silence_active_extinfo:-0}" = "0" ]` instead. Otherwise you would surprise users by misinterpreting a value of `0`.

Additionally you should set this value in the header of the plugin in order to give a quick overview, which environment variables are parsed.

plugins/systemd/systemd_units
```diff
@@ -123,8 +125,12 @@ fetch () {
...
+			[ -n "$extinfo" ] && echo "$state.extinfo" "$extinfo"
```
sumpfralle, line 129: This short conditional expression (`... && ...`) usually works fine, but should be avoided at the end of complicated expressions (here `if ... fi`). Otherwise the non-zero exitcode (if `extinfo` is empty) would traverse up to the higher layers and maybe even end up as an unexpected exit code of the plugin.

This issue is more relevant for scripts using the *exit on error* setting (`set -e`), but it is applicable here, too.

plugins/systemd/systemd_units
```diff
@@ -38,10 +38,12 @@ An example configuration might be something like this:
  env.inactive_critical 20
  env.exclude boring
  env.inactive_exclude sleepy
+ env.silence_active_extinfo 1
 
 In the example above, we've overridden the default warning and critical levels for failed units, added warning
 and critical levels for inactive units, then filtered out boring units from all results and filtered out sleepy
-units from results for the inactive state.
+units from results for the inactive state. In addition to that, only more extensive info of non-active units
```
sumpfralle, line 45: Please mention the default value of `silence_active_extinfo`.

plugins/systemd/systemd_units
PoroCYon: Done that.

plugins/systemd/systemd_units
PoroCYon: Fixed.

plugins/systemd/systemd_units
PoroCYon: Done.

sumpfralle:
Thank you!

PR #1101, opened 2020-07-31
[plugins/shorewall/shorewall_log] Added plugin for monitoring shorewall blocks

PoroCYon:
Graphs the number of blocks done by shorewall, given a specific rule suffix

### DIFF
--- a/plugins/shorewall/shorewall_log
+++ b/plugins/shorewall/shorewall_log
@@ -1,0 +1,298 @@
+#!/usr/bin/env python3
+# -*- python -*-
+
+"""
+
+=head1 NAME
+
+Plugin to monitor iptables logs configured by shorewall
+
+=head1 CONFIGURATION
+
+  logfile: Path to the iptables log file, or "journald" to use journald.
+           When journalctl exists, the default is "journald", otherwise
+           "/var/log/kern.log".
+  journalctlargs: Arguments passed to journalctl to select the right logs.
+                  The default is "SYSLOG_IDENTIFIER=kernel".
+  taggroups: Space separated list of groups. A group contains a tag if the
+             group is substring of the tag. Tags belonging to the same group
+             will be combined in one graph.
+  tagfilter: Space separated list of filters. When a tag is matched by a
+             filter (i.e. if the filter is a substring of the tag) it is
+             ignored.
+  prefixformat: The format of the prefix configured in iptables, this is the
+                LOGFORMAT option in shorewall.conf. When not set the entire
+                prefix is used.
+  include_ungrouped: when a tag is found that does not belong to a group,
+                     make it it's own group
+
+Example:
+
+Using /var/log/kern.log as logfile:
+
+=over 2
+
+  [shorewall_log]
+  group adm
+  env.logfile /var/log/kern.log
+
+=back
+
+Using journald:
+
+=over 2
+
+  [shorewall_log]
+  group systemd-journal
+
+=back
+
+=head1 HISTORY
+
+2017-11-03: v1.0 Bert Van de Poel <bert@bhack.net>: created
+2020-07-16: v2.0 Vincent Vanlaer <vincenttc@ulyssis.org>: rewrite
+                 - read all tags from iptables config, instead of the last 24h
+                   of logs
+                 - add support for journald
+                 - use cursors for accuracy
+                 - convert to multigraph to reduce load
+
+=head1 USAGE
+
+Parameters understood:
+
+  config   (required)
+  autoconf (optional - used by munin-config)
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+"""
+
+
+import sys
+import os
+
+
+def autoconf() -> str:
+    if sys.version_info < (3, 5):
+        return 'no (This plugin requires python 3.5 or higher)'
+
+    if os.getenv('MUNIN_CAP_MULTIGRAPH', '0') != '1':
+        return 'no (No multigraph support)'
+
+    import shutil
+
+    if not shutil.which('shorewall'):
+        return 'no (No shorewall executable found)'
+
+    if not shutil.which('iptables-save'):
+        return 'no (No iptables-save executable found, required for tag enumeration)'
+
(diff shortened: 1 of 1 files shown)


plugins/shorewall/shorewall_log
```diff
@@ -0,0 +1,292 @@
...
+                prefix = next(args)
+
+                if prefix_format:
+                    tag = prefix_format.match(prefix)[1]
+                else:
+                    tag = prefix.rstrip()
+
+                for ignored in tagfilter:
+                    if ignored in tag:
+                        continue
```
sumpfralle, line 149: I think, this will just leave the inner `for` loop.
How about the following?
```python
if any(ignored in tag for ignored in tagfilter):
    continue
```

plugins/shorewall/shorewall_log
```diff
@@ -0,0 +1,292 @@
...
+
+
+def get_logtags() -> Tuple[Dict[str, Set[str]], Dict[str, Set[str]]]:
+    rules = (run(['iptables-save'], stdout=PIPE, universal_newlines=True).
+             stdout.splitlines())
+    tags = defaultdict(set)
+    groups = defaultdict(set)
+
+    for line in rules:
+        args = iter(shlex.split(line))
```
sumpfralle, line 135: The `break` statements are a bit confusing for me.

How about processing the list instead of using a loop?

```python
if ("-j" in args) and (args[args.index("-j") + 1] == "LOG"):
    if "--log-prefix" in args:
        prefix = args[args.index("--log-prefix") + 1]
        ....
```

Additionally it would be helpful, if you could add an example line as a comment. This would ease the understanding of the code.

plugins/shorewall/shorewall_log
```diff
@@ -0,0 +1,292 @@
...
+        return pickle.dump(state, f)
+
+
+def get_lines_journalctl(state: State) -> Iterator[str]:
+    cursor = state.journal
+
+    def catch_cursor(line: str):
+        cursor_id = '-- cursor: '
+        if line.startswith(cursor_id):
+            save_state(State(line[len(cursor_id):-1], None))
```
sumpfralle, line 187: Is the last character in `line` thrown away on purpose? Maybe it is a line break?

plugins/shorewall/shorewall_log
```diff
@@ -0,0 +1,292 @@
+#!/usr/bin/python3
```
sumpfralle, line 1: Please use `/usr/bin/env python3` in order to support virtualenvs.

plugins/shorewall/shorewall_log
```diff
@@ -0,0 +1,292 @@
...
+# -*- python -*-
+
+"""
+
+=head1 INTRODUCTION
```
sumpfralle, line 6: Please use the section name `NAME` instead. This will improve the visualization in the [plugin gallery](https://gallery.munin-monitoring.org/).

sumpfralle (commented):
Thank you for proposing this plugin!

plugins/shorewall/shorewall_log
PoroCYon: that was a bug, `.splitlines()` already removes the line breaks

plugins/shorewall/shorewall_log
PoroCYon: done

plugins/shorewall/shorewall_log
PoroCYon: done

plugins/shorewall/shorewall_log
```diff
@@ -0,0 +1,292 @@
...
+Parameters understood:
+
+  config   (required)
+  autoconf (optional - used by munin-config)
+
+=head1 MAGIC MARKERS
+
+#%# family=auto
+#%# capabilities=autoconf
+"""
```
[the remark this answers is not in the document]
PoroCYon: done

plugins/shorewall/shorewall_log
PoroCYon: I hope to have clarified it a bit with additional comments, the person who originally wrote the code was somehow a bit vindicative against changing it into your proposed form.

FYI, "line" here is just a normal `iptables` rule.

plugins/shorewall/shorewall_log
```diff
@@ -0,0 +1,298 @@
...
+                                             'SYSLOG_IDENTIFIER=kernel')))
+taggroups = os.getenv('taggroups')
+taggroups = taggroups.split() if taggroups else []
+tagfilter = os.getenv('tagfilter')
+tagfilter = tagfilter.split() if tagfilter else []
+include_ungrouped = (os.getenv('includeungrouped', 'true').lower() == 'true')
+prefix_format = os.getenv('prefixformat')
+if prefix_format:
+    if sys.version_info < (3, 7):
+        prefix_format = re.escape(prefix_format).replace('\\%s', '(.+)').replace('\\%d', '\\d+')
```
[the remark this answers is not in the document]
PoroCYon: `re.escape` returns a string, so this is a regular string replace operation. The grouping/capture is needed for the `match(...)[1]` on line 150.

plugins/shorewall/shorewall_log
```diff
@@ -0,0 +1,298 @@
...
+=head1 USAGE
+
+Parameters understood:
+
+  config   (required)
+  autoconf (optional - used by munin-config)
+
+=head1 MAGIC MARKERS
+
+=cut
```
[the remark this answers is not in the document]
PoroCYon: Oops, yes, fixed.

plugins/shorewall/shorewall_log
PoroCYon: Oops, I changed the `=head1` to `=name` instead of the `INTRODUCTION`, fixed now.

plugins/shorewall/shorewall_log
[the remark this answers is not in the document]
sumpfralle: Initially I was confused by the missing braces around the numeric expression (`\\d+`). But I guess, this number is not relevant (just being non-predictable). I guess, you know what you are doing :)

sumpfralle:

PR #1103, opened 2020-07-31
[plugins/mysql/mysql_disk_by_prefix] Add plugin for DB size by prefix

PoroCYon:
The `mysql_size_*` scripts list the size per database, not per user

### DIFF
--- a/plugins/mysql/mysql_disk_by_prefix
+++ b/plugins/mysql/mysql_disk_by_prefix
@@ -1,0 +1,151 @@
+#!/usr/bin/env python3
+# -*- python -*-
+
+"""
+=head1 INTRODUCTION
+
+Plugin to monitor MySQL database disk usage per prefix. A prefix can be eg.
+'user' for the databases 'user_testdb' and 'user_db2', as can be seen in
+certain shared hosting setups.
+
+=head1 APPLICABLE SYSTEMS
+
+Local access to the database files is required (which are stored in
+/var/lib/mysql by default).
+
+=head1 INSTALLATION
+
+Place in /etc/munin/plugins/ (or link it there using ln -s)
+
+=head1 CONFIGURATION
+
+Add this to your /etc/munin/plugin-conf.d/munin-node:
+
+=over 2
+
+    [mysql_disk_by_prefix]
+    user mysql
+    env.prefixes_with_underscores foo_bar d_ritchie # prefixes that include an underscore
+    env.db_minsize 1024000 # minimum db size in order to report a specific prefix,
+                           # in bytes (defaults to 50M). "0" for none.
+    env.mysql_db_dir /var/lib/mysql # default value
+
+=back
+
+=head1 HISTORY
+
+2019-07-25: v 1.0 pcy <pcy.ulyssis.org>: created
+
+=head1 USAGE
+
+Parameters understood:
+
+  config   (required)
+  autoconf (optional - used by munin-config)
+
+=head1 MAGIC MARKERS
+
+#%# family=auto
+#%# capabilities=autoconf
+"""
+
+
+import os
+import re
+import subprocess as sp
+import sys
+
+
+def weakbool(x):
+    return x.lower().strip() in {'true', 'yes', 'y', 1}
+
+
+LIMIT = os.getenv('db_minsize', str(50000 * 1024))
+try:
+    LIMIT = int(LIMIT)
+except ValueError:
+    LIMIT = 0
+
+exceptions = os.getenv('prefix_with_underscores', '').split(' ')
+if exceptions == ['']:
+    exceptions = []
+
+mysqldir = os.getenv('mysql_db_dir', '/var/lib/mysql')
+
+
+def name_from_path(path):
+    filename = os.path.basename(path)
+    for name in exceptions:
+        if filename.startswith(name):
+            return name
+    name = filename.split('_')[0]
+
+    def decode_byte(m):
+        return bytes.fromhex(m.group(1)).decode('utf-8')
+
+    # Decode MySQL's encoding of non-ascii characters in the table names
+    return re.sub('@00([0-9a-z]{2})', decode_byte, name)
+
+
+def calc_dir_size(directory):
+    total = 0
+
+    for filename in os.listdir(directory):
+        filedir = os.path.join(directory, filename)
+
+        if os.path.islink(filedir):
+            continue
+        if os.path.isfile(filedir):
+            total += os.path.getsize(filedir)
+
+    return total
+
+
+def size_per_subdir(parentdir):
+    for subdir in os.listdir(parentdir):
+        dirpath = os.path.join(parentdir, subdir)
+
+        if os.path.islink(dirpath):
+            continue
+        if os.path.isdir(dirpath):
+            yield calc_dir_size(dirpath), name_from_path(os.path.split(dirpath)[1])
+
+
+def sizes_by_name(limit=None):
+    sizes = {}
+    for size, name in size_per_subdir(mysqldir):
+        sizes[name] = sizes.get(name, 0) + size
+    for name, total_size in sizes.items():
+        if limit <= 0 or limit <= total_size:
+            yield name, total_size
+
+
+def fetch():
+    for name, total_size in sorted(sizes_by_name(limit=LIMIT), key=lambda x: x[0]):
+        print('{}.value {}'.format(name, total_size))
+
+
+def main():
+    if len(sys.argv) == 1:
+        fetch()
+    elif sys.argv[1] == 'config':
+        print('graph_title MySQL disk usage by prefix')
+        print('graph_vlabel bytes')
+        print('graph_category db')
+
+        names = sorted(name for name, _ in sizes_by_name(limit=LIMIT))
+        for name in names:
+            print('{0}.label {0}'.format(name))
+            print('{}.type GAUGE'.format(name))
+            print('{}.draw AREASTACK'.format(name))
+    elif sys.argv[1] == 'autoconf':
+        print('yes' if os.path.isdir(mysqldir)
+              else "no (can't find MySQL's data directory)")
+    else:
+        fetch()
+
+
+if __name__ == "__main__":
+    main()
+
+# flake8: noqa: E265


plugins/mysql/mysql_disk_by_user
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env python3
+# -*- python -*-
+# -*- coding: utf-8 -*-
```
sumpfralle, line 3: This was probably only relevant for python2.

plugins/mysql/mysql_disk_by_user
```diff
@@ -0,0 +1,113 @@
...
+#
+# Plugin to monitor MySQL database disk usage per user
```
sumpfralle, line 5: Please use a perldoc header.

plugins/mysql/mysql_disk_by_user
```diff
@@ -0,0 +1,113 @@
...
+#
+# Usage: Place in /etc/munin/plugins/ (or link it there using ln -s)
+#   Add this to your /etc/munin/plugin-conf.d/munin-node:
+#       [mysql_disk_size]
+#       user mysql
+#       users_with_underscores foo_bar d_ritchie # usernames of users in the
```
sumpfralle, line 11: These need to be prefixed with `env.`, I guess?

Additionally: why is *underscore* relevant here? Does this need to be part of the variable name?

plugins/mysql/mysql_disk_by_user
```diff
@@ -0,0 +1,113 @@
...
+#                                                # system that include an underscore
+#       limit 1024000 # minimum db size in order to report a specific user,
```
sumpfralle, line 13: Maybe use `0` as the *no limit* value. This is quite common practice. Additionally it avoids the ambiguity of interpreting a number as as string under certain circumstances.

Additionally it would be good to use a more precise name: `limit` is a bit too broad for its purpose.

plugins/mysql/mysql_disk_by_user
```diff
@@ -0,0 +1,113 @@
...
+    u = x.upper().strip()
+    if u == "FALSE" or u == "NO" or u == "N":
+        return False
+    try:
+        return bool(int(x))
+    except ValueError:
+        return bool(x)
+
+
+LIMIT = os.getenv('limit', str(50000 * 1024))
```
sumpfralle, line 43: I think, you are using 50M as a default here (instead of 5M advertised in the documentation).

plugins/mysql/mysql_disk_by_user
```diff
@@ -0,0 +1,113 @@
...
+
+    def decode_byte(m):
+        return bytes.fromhex(m.group(1)).decode('utf-8')
+
+    return re.sub('@00([0-9a-z]{2})', decode_byte, name)
+
+
+def sizes_by_name(limit=None):
+    sizes = dict({})
+    output = sp.run(['du', '-B1'], cwd='/var/lib/mysql', stdout=sp.PIPE)
```
sumpfralle, line 69: The path `/var/lib/mysql` should be configurable, I guess.

Additionally: wouldn't it be easier to access the files with python code and check their size without spawning `du`?

plugins/mysql/mysql_disk_by_user
```diff
@@ -0,0 +1,113 @@
...
+    filename = os.path.basename(path)
+    for name in exceptions:
+        if filename.startswith(name):
+            return name
+    name = filename.split('_')[0]
+
+    def decode_byte(m):
+        return bytes.fromhex(m.group(1)).decode('utf-8')
+
+    return re.sub('@00([0-9a-z]{2})', decode_byte, name)
```
sumpfralle, line 64: I am only guessing here: do you assume, that the `user` (owner?) of the database can be identified via the filename prefix?

In this case I would hesitate to call the plugin `mysql_disk_by_user` - maybe `mysql_disk_by_db_prefix` instead?

PoroCYon: Right, it's only the username in our hosting setup. Will rename.

plugins/mysql/mysql_disk_by_user
PoroCYon: Removed.

plugins/mysql/mysql_disk_by_user
PoroCYon: Done.

plugins/mysql/mysql_disk_by_user
PoroCYon: Indeed, oops.

It's relevant (for some value thereof) because in our setup, the prefies (usernames in our case) sometimes may contain an underscore, while the directories are called "<username>_<dbname>". I know, it's a hack, I'll remove it if you prefer.

plugins/mysql/mysql_disk_by_user
PoroCYon: Done and done.

plugins/mysql/mysql_disk_by_user
PoroCYon: Oops, you're right, fixed.

plugins/mysql/mysql_disk_by_user
PoroCYon: Made it configurable.

Rewrote the `du` part in Python, but it's a bit longer now, actually.

plugins/mysql/mysql_disk_by_prefix
```diff
@@ -0,0 +1,146 @@
...
+
+
+def weakbool(x):
+    return x.lower().strip() in {'true', 'yes', 'y', 1}
+
+
+LIMIT = os.getenv('db_minsize', str(50000 * 1024))
+try:
+    LIMIT = int(LIMIT)
+except Exception:
```
sumpfralle, line 59: Please avoid unspecific exceptions. `ValueError` would be suitable here.

plugins/mysql/mysql_disk_by_prefix
```diff
@@ -0,0 +1,146 @@
...
+
+def main():
+    if len(sys.argv) == 1:
+        for name, total_size in sizes_by_name(limit=LIMIT):
+            print('{}.value {}'.format(name, total_size))
+
+    elif sys.argv[1] == 'config':
+        print('graph_title MySQL disk usage by prefix')
+        print('graph_vlabel bytes')
+        print('graph_category mysql')
```
sumpfralle, line 123: Please use `db` instead.

plugins/mysql/mysql_disk_by_prefix
```diff
@@ -0,0 +1,146 @@
...
+        itr = iter(sizes_by_name(limit=LIMIT))
+        name, _ = next(itr)
+        print('{0}.label {0}'.format(name))
+        print('{}.type GAUGE'.format(name))
+        print('{}.draw AREA'.format(name))
```
sumpfralle, line 128: You can use `AREASTACK` in order to avoid the manual `iter` processing.

plugins/mysql/mysql_disk_by_prefix
```diff
@@ -0,0 +1,146 @@
...
+        if os.path.islink(dirpath):
+            continue
+        if os.path.isdir(dirpath):
+            yield calc_dir_size(subdir), name_from_path(os.path.split(subdir)[1])
+
+
+def sizes_by_name(limit=None):
+    sizes = {}
+    for size, name in size_per_subdir(mysqldir):
+        sizes[name] = (sizes[name] + size) if name in sizes else size
```
sumpfralle, line 109: How about `sizes[name] = sizes.get(name, 0)` + size?

plugins/mysql/mysql_disk_by_prefix
```diff
@@ -0,0 +1,146 @@
...
+    for name, total_size in sizes.items():
+        if LIMIT <= 0 or LIMIT <= total_size:
```
sumpfralle, line 111: You probably wanted to use `limit` instead of `LIMIT` here? (or just remove it from the interface of the function)

plugins/mysql/mysql_disk_by_prefix
```diff
@@ -0,0 +1,146 @@
...
+    filename = os.path.basename(path)
+    for name in exceptions:
+        if filename.startswith(name):
+            return name
+    name = filename.split('_')[0]
+
+    def decode_byte(m):
+        return bytes.fromhex(m.group(1)).decode('utf-8')
+
+    return re.sub('@00([0-9a-z]{2})', decode_byte, name)
```
sumpfralle, line 79: I assume, that this is mysql's approach of encoding non-ascii characters in database directory names? Maybe add a short comment mentioning this?

plugins/mysql/mysql_disk_by_prefix
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python3
+# -*- python -*-
+
+"""
+=head1 INTRODUCTION
+
+Plugin to monitor MySQL database disk usage per prefix
+
```
sumpfralle, line 8: Please add a section `APPLICABLE SYSTEMS` (which is very common for munin plugins) and mention that local access to the database directory is required.

plugins/mysql/mysql_disk_by_prefix
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python3
+# -*- python -*-
+
+"""
+=head1 INTRODUCTION
+
+Plugin to monitor MySQL database disk usage per prefix
```
sumpfralle, line 7: I think, it is good idea to add one sentence describing the `prefix` (e.g. *database name prefix*) and maybe mention your example of the convention of prefixing database names with the name of their owner.

plugins/mysql/mysql_disk_by_prefix
```diff
@@ -0,0 +1,146 @@
...
+def main():
+    if len(sys.argv) == 1:
+        for name, total_size in sizes_by_name(limit=LIMIT):
+            print('{}.value {}'.format(name, total_size))
+
+    elif sys.argv[1] == 'config':
+        print('graph_title MySQL disk usage by prefix')
+        print('graph_vlabel bytes')
+        print('graph_category mysql')
+        itr = iter(sizes_by_name(limit=LIMIT))
```
sumpfralle, line 124: At some point you should sort the entries (e.g. by name) in order to achieve a stable order of data fields.

(`os.listdir` is returns the entries in an arbitrary order)

plugins/mysql/mysql_disk_by_user
sumpfralle: Thanks for renaming. Please do not forget to adjust the commit message.

plugins/mysql/mysql_disk_by_user
sumpfralle: I think, I was confused here, since you never explicitly mentioned, that *prefix* refers to the token before an *underscore*. Maybe you want to add this somewhere in the documentation.

Due to this specific meaning of the underscore character, the environment variable name looks fine to me.

plugins/mysql/mysql_disk_by_prefix
PoroCYon: Fixed.

plugins/mysql/mysql_disk_by_prefix
PoroCYon: Done.

plugins/mysql/mysql_disk_by_prefix
PoroCYon: Done.

plugins/mysql/mysql_disk_by_prefix
PoroCYon: Done that, thanks, didn't know about this one.

plugins/mysql/mysql_disk_by_prefix
PoroCYon: Indeed, oops.

plugins/mysql/mysql_disk_by_prefix
PoroCYon: Yep, commented that.

plugins/mysql/mysql_disk_by_prefix
PoroCYon: Done.

plugins/mysql/mysql_disk_by_prefix
PoroCYon: Done.

plugins/mysql/mysql_disk_by_prefix
PoroCYon: Done.

(`os.listdir` uses inode order iirc, which is usually mostly stable between runs, but no guarantees of course.)

sumpfralle:
Thank you for your patience!

PR #1104, opened 2020-07-31
[plugins/mysql/mysql_audit] Added plugin for monitoring the audit log

PoroCYon:
Shows the number of active connections over time

### DIFF
--- a/plugins/mysql/mysql_audit
+++ b/plugins/mysql/mysql_audit
@@ -1,0 +1,189 @@
+#!/usr/bin/env python3
+# -*- python -*-
+
+"""
+=head1 INTRODUCTION
+
+Plugin to monitor the MySQL audit log connection count
+
+=head1 APPLICABLE SYSTEMS
+
+MySQL needs to be configured manually in order to enable the audit log. This
+can be done as follows:
+
+=over 2
+
+      plugin-load = server_audit=server_audit.so
+      server_audit_events = connect
+      server_audit_output_type=file
+      server_audit_file_path = /var/log/mysql/audit.log
+      server_audit_file_rotate_size = 512M
+      server_audit_logging = ON
+      server_audit_file_rotations = 5
+
+=back
+
+=head1 INSTALLATION
+
+Place in /etc/munin/plugins/ (or link it there using ln -s)
+
+=head1 CONFIGURATION
+
+Add this to your /etc/munin/plugin-conf.d/munin-node:
+
+=over 2
+
+      [mysql_audit]
+      user mysql
+      env.logfile /var/log/mysql/audit.log
+      env.sorted 1 # sort output (if not: unsorted)
+      env.toplist 10 # only the top 10 (if unset: all of them). implies sorted
+
+=back
+
+=head1 HISTORY
+
+2017-11-03: v 1.0 Bert Van de Poel <bert@bhack.net>: created
+2020-07-19: v 1.1 pcy <pcy@ulyssis.org>: added config options
+
+=head1 USAGE
+
+Parameters understood:
+
+  config   (required)
+  autoconf (optional - used by munin-config)
+
+=head1 MAGIC MARKERS
+
+#%# family=auto
+#%# capabilities=autoconf
+"""
+
+
+from datetime import datetime, timedelta, timezone
+import operator
+import os
+import struct
+import sys
+
+
+def weakbool(x):
+    return x.lower().strip() in {'true', 'yes', 'y', '1'}
+
+
+logfile = os.getenv('logfile', '/var/log/mysql/audit.log')
+toplist = int(os.getenv('toplist', '0'))
+sortlist = weakbool(os.getenv('sorted', 'N')) or toplist > 0
+
+STATEFILE = os.getenv('MUNIN_STATEFILE')
+
+
+def loadstate():
+    if not os.path.isfile(STATEFILE):
+        return None
+
+    with open(STATEFILE, 'rb') as f:
+        tstamp = struct.unpack('d', f.read())[0]
+        return datetime.fromtimestamp(tstamp, tz=timezone.utc)
+
+
+def savestate(state):
+    with open(STATEFILE, 'wb') as f:
+        f.write(struct.pack('d', state.timestamp()))
+
+
+def reverse_lines(filename, BUFSIZE=4096):
+    with open(filename, "r") as f:
+        f.seek(0, 2)
+        p = f.tell()
+        remainder = ""
+        while True:
+            sz = min(BUFSIZE, p)
+            p -= sz
+            f.seek(p)
+            buf = f.read(sz) + remainder
+            if '\n' not in buf:
+                remainder = buf
+            else:
+                i = buf.index('\n')
+                for L in buf[i + 1:].split("\n")[::-1]:
+                    yield L
+                remainder = buf[:i]
+            if p == 0:
+                break
+        yield remainder
+
+
+def get_data(do_save=True):
+    occurrences = {}
+    begin = datetime.now(timezone.utc)
+    begin_local = datetime.now()
+
+    state = loadstate()
+    if state is None:
+        # need to do something here to prevent reading indefinitely
+        state = begin - timedelta(minutes=5)
+
+    for line in reverse_lines(logfile):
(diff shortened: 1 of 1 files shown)


plugins/mysql/mysql_audit
```diff
@@ -0,0 +1,160 @@
...
+
+=head1 CONFIGURATION
+
+Add this to your /etc/munin/plugin-conf.d/munin-node:
+
+=over 2
+
+      [mysql_audit]
+      user mysql
+      period 3
```
sumpfralle, line 21: I think, you omitted the `env.` prefix here.

plugins/mysql/mysql_audit
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/python
```
sumpfralle, line 1: Probably you meant `python3` or even `/usr/bin/env python3`?

plugins/mysql/mysql_audit
```diff
@@ -0,0 +1,160 @@
...
+
+
+import sys
+import os
+from datetime import datetime, timedelta
+import operator
+
+
+def weakbool(x):
+    return x.lower().strip() in {'true','yes','y','1'}
```
sumpfralle, line 69: PEP8 welcomes spaces between operators (`,`). `flake8` is your friend :)

plugins/mysql/mysql_audit
```diff
@@ -0,0 +1,160 @@
...
+
+
+tdminutes = os.getenv('period', '5')
+logfile = os.getenv('logfile', '/var/log/mysql/audit.log')
+toplist = int(os.getenv('toplist', '0'))
+sortlist = weakbool(os.getenv('sorted', 'N')) or toplist > 0
+
+
+def reverse_lines(filename, BUFSIZE=4096):
+    f = open(filename, "r")
```
sumpfralle, line 79: Not so long ago python introduced the *contexts*, which are great when dealing with files (e.g. `with open(filename, "r") as f:`), since it takes care for cleaning up (closing handles). Maybe you want to change this?

plugins/mysql/mysql_audit
```diff
@@ -0,0 +1,160 @@
...
+    f.seek(0, 2)
+    p = f.tell()
+    remainder = ""
+    while True:
```
sumpfralle, line 83: After reading the function twice, I understood its purpose, I guess ... :)

I think, a statefile would be suitable here. In order to reduce re-parsing of old lines, the statefile could contain the filesize during the previous run (i.e. the already processed content). Whenever the file size decreases you would start again at the beginning. This would would mimic the implementation of `tail_open` in `Munin::Plugins`.

plugins/mysql/mysql_audit
```diff
@@ -0,0 +1,160 @@
...
+                yield L
+            remainder = buf[:i]
+        if p == 0:
+            break
+    yield remainder
+
+
+def get_data():
+    occurrences = dict()
+    date_threshold = (datetime.now() - timedelta(minutes=int(tdminutes)))
```
sumpfralle, line 102: Is `tdminutes` related to munin's retrieval period? In this case a statefile would be nice.

plugins/mysql/mysql_audit
```diff
@@ -0,0 +1,160 @@
...
+    print("no (logfile not found)" if os.path.isfile(logfile) else "yes")
+
+
+def configure():
+    print('graph_title MySQL Audit connect count')
+    print('graph_vlabel Connections/%s minutes' % tdminutes)
+    print('graph_category mysql')
+
+    occurrences = get_data()
+    occitems = sorted(occurrences.items(), key=operator.itemgetter(1), reverse=True) \
```
sumpfralle, line 131: Just for code beautification (feel free to ignore my taste here): maybe just assign the dict items and use a conditional in-place sorting afterwards?
(I was struggling for a few seconds before I understood the simple purpose of this evaluation)

Regarding sorting: in general you should try to produce a stable output order of the fields. Otherwise the colors for the different fields will change randomly in munin's graph. Just sort based on the key, if nothing else is configured.

plugins/mysql/mysql_audit
PoroCYon: Oops, sorry again.

plugins/mysql/mysql_audit
PoroCYon: Indeed.

plugins/mysql/mysql_audit
PoroCYon: Yes. The CI pipeline actually runs `flake8`, which is how I fixed a few of these issues before, but it always choked on the magic markers (cf. [E265](https://www.flake8rules.com/rules/E265.html)), which is why I added [this directive](https://github.com/munin-monitoring/contrib/blob/9ad2de67c790d01d8c872ae70110ee89be979789/plugins/mysql/mysql_audit#L160) at the very end of the file. But it looks like `flake8` misparses that and starts ignoring *everything*...

plugins/mysql/mysql_audit
PoroCYon: Right, will fix.

plugins/mysql/mysql_audit
PoroCYon: It is indeed... will do that I guess. (and also cf. the previous comment)

plugins/mysql/mysql_audit
PoroCYon: I don't think that extra step is needed here, as the `for` loop further down will break after the date cutoff. All it'd do is reduce the amount of seeking a bit, but not the parsing.

It would be nice to have of course, for other plugins, but I don't feel like messing with this function much... (A colleague of mine wrote it for the shorewall plugin, I just took the code. (Not exactly a colleague as what we're doing isn't technically our job, but, you get the idea.))

plugins/mysql/mysql_audit
PoroCYon: Done that, sort of. Always sorting now, but:

* if `sortitems`: sort by value/amount and in descending order (this option is useful for us to find out which user is making too many database connections, which probably means there's something going wrong)
* else: sort by key/name, ascending

plugins/mysql/mysql_audit
sumpfralle: Good catch! :)

Indeed the flake8 documentation says, that files containing the line `# flake8: noqa` are skipped. It seems like a partial match (not the full line) is sufficient for this.

Probably `E265` is not relevant anymore, since you switched to the perldoc header?

I just removed the line from `mysql_disk_by_prefix`. Please do the same for your open pull requests, if applicable.

plugins/mysql/mysql_audit
```diff
@@ -0,0 +1,206 @@
...
+
+logfile = os.getenv('logfile', '/var/log/mysql/audit.log')
+toplist = int(os.getenv('toplist', '0'))
+sortlist = weakbool(os.getenv('sorted', 'N')) or toplist > 0
+
+STATEFILE = os.getenv('MUNIN_STATEFILE')
+STATE = None
+
+
+if os.path.isfile(STATEFILE):
```
sumpfralle, line 81: I think, it would be good to move this to a `load_state` function and only call it, when it is necessary (avoiding `global`).

Additionally it would probably be much simpler to store the file position instead of a timestamp. But this is a matter of taste.

PoroCYon: Moved it to a function.

As I argued [above](https://github.com/munin-monitoring/contrib/pull/1104#discussion_r471213477), I don't think it won't cause unnecessary parsing etc., and because it works right now, I don't want to mess with it much.

sumpfralle:
Thanks for your changes.

I think, there is only one detail left: at the moment `get_data()` updates the timestamp in the statefile. But `get_data()` is not just called in `fetch`, but also during `config`. I guess, you should move the `save_state()` call to the `fetch` function (or add a boolean parameter to `get_data`).

Do I understand this correctly? At the moment, your graphs should be more or less blank, I guess?

PoroCYon:
Wait, does munin run config every time it runs a script, or only once, for configuring? Because I thought it was the latter?

niclan:
Every time. Because dynamic environments.

PoroCYon:
Ah I see. Changed it, then.

PR #1105, opened 2020-07-31
[plugins/user/membyuser] Minor enhancements

PoroCYon:
* Don't use a fixed scale
* Use resident set size as measure for a process' memory usage instead of vsz
* More robust username enumeration

### DIFF
--- a/plugins/user/membyuser
+++ b/plugins/user/membyuser
@@ -2,10 +2,18 @@
 #
 # Plugin to monitor Memory usage inspired by cpubyuser
 #
-# 2012-05-23 Sebastien Campion
+# Usage: Place in /etc/munin/plugins/ (or link it there using ln -s)
+#	Add this to your /etc/munin/plugin-conf.d/munin-node:
+#		[membyuser]
+#		user root # required if /proc can't be read from by any user!
 
-LU=`ps auh | cut -d' ' -f 1  | sort -u`
-USERS=`echo $LU`
+## 2012-05-23 Sebastien Campion
+# changed on 2019-08-30 by pcy <pcy@ulyssis.org>:
+# - change category from 'memory' to 'system' (so it appears next to cpubyuser)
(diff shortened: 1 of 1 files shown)


plugins/user/membyuser
```diff
@@ -2,9 +2,18 @@
...
+#		user root # required if /proc can't be read from by any user!
 
-LU=`ps auh | cut -d' ' -f 1  | sort -u`
+## 2012-05-23 Sebastien Campion
+# changed on 2019-08-30 by pcy <pcy@ulyssis.org>:
+# - change category from 'memory' to 'system' (so it appears next to cpubyuser)
+# - use rss instead of vsz
+# - more robust username enumeration
+
+LU=`ps ax --format uname | tail +2 | sort -u | grep -v -e "^root$"`
```
sumpfralle, line 16: Please use `$(...)` instead of backticks. Feel free to merge this line with the following one.

plugins/user/membyuser
```diff
@@ -17,12 +26,11 @@ if [ "$1" = "autoconf" ]; then
 fi
 
 if [ "$1" = "config" ]; then
-	echo "graph_args --base 1000 -r --lower-limit 0"
+	echo "graph_args --base 1024"
 	echo "graph_title Memory usage, by user"
-	echo "graph_category memory"
+	echo "graph_category system"
```
sumpfralle, line 31: I think, `memory` is suitable here, since the plugins focuses on this property.

PoroCYon: That is fair, the other 'xyzbyuser' scripts seem to be in the 'system' category (iirc), but I'll roll it back.

plugins/user/membyuser
PoroCYon: Fixed.

sumpfralle:
Thanks!

PR #1107, opened 2020-07-31
[plugins/dovecot/dovecot] Make 'connected users' graph DERIVEable

PoroCYon:
This is probably more useful

### DIFF
--- a/plugins/dovecot/dovecot
+++ b/plugins/dovecot/dovecot
@@ -8,6 +8,7 @@
 # Contributions by:
 # - Stephane Enten <tuf@delyth.net>
 # - Steve Schnepp <steve.schnepp@pwkf.org>
+# - pcy <pcy@ulyssis.org> (make 'Connected Users' DERIVE, check existence of logfile in autoconf)
 #
 # Parameters understood:
 #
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for this proposal!

Just for my understanding: the `DERIVE` approach would change the resulting data from *currently connected users* to *change of connected users since last period*, or?

If yes: is this a helpful metric for a dovecot administrator? If you think so, then please rename the setting in order to better reflect the purpose of this configuration.

niclan:
Looks like the plugin reports the cummulative value for each field:

```
$ bash ./dovecot 
login_total.value 2022
connected.value 0
login_tls.value 2022
login_ssl.value 0
login_imap.value 2022
login_pop3.value 0
```
which in my eyes means it really should be DERIVE. I agree with this change.

PoroCYon:
Indeed, this modification turns it into the actual described metrics.

sumpfralle:
> Looks like the plugin reports the cummulative value for each field: [..]

In this case, the `DERIVE` evaluation should be the default (without configurability), or?

PoroCYon:
DERIVE is currently the default, I made it configurable for people who want to keep the old behavior. (Though I can't immediately think of a reason why one would want that. But I can always be wrong.)

sumpfralle:
In this case I am confident, that the previous behaviour was just plain wrong (or turned into being wrong at some time). Thus I would prefer, if you remove the configurability.

In case of unknown corner-cases we will find a way to distinguish between them.

Thank you!

PoroCYon:
Done that.

sumpfralle:
Thanks!

PR #1108, opened 2020-07-31
[plugin/user/cronjobs] Added plugin to monitor cronjobs

PoroCYon:
Graphs the number of cronjobs active at a certain moment, per user

### DIFF
--- a/plugins/user/cronjobs
+++ b/plugins/user/cronjobs
@@ -1,0 +1,165 @@
+#!/usr/bin/env python3
+# -*- python -*-
+"""
+=head1 NAME
+
+Plugin to monitor the number of cronjobs running per user, gathering data from
+syslog.
+
+=head1 INSTALLATION
+
+Usage: Place in /etc/munin/plugins/ (or link it there using ln -s)
+
+=head1 CONFIGURATION
+
+Add this to your /etc/munin/plugin-conf.d/munin-node:
+
+=over 2
+
+      [cronjobs]
+      user root
+      env.syslog_path /var/log/syslog # default value
+      env.cron_ident_regex crond? # for finding cron entries in the syslog,
+                                    case-insensitive (default value)
+
+=back
+
+The plugin needs to run as root in order to read from syslog.
+
+=head1 HISTORY
+
+2019-09-09: v 1.0 pcy <pcy.ulyssis.org>: created
+
+=head1 USAGE
+
+Parameters understood:
+
+  config   (required)
+  autoconf (optional - used by munin-config)
+
+=head1 MAGIC MARKERS
+
+#%# family=auto
+#%# capabilities=autoconf
+"""
+
+
+from datetime import datetime, timedelta, timezone
+import os
+import re
+import shutil
+import socket
+import sys
+import struct
+import time
+
+
+syslog_path = os.getenv('syslog_path', '/var/log/syslog')
+cron_ident_regex = os.getenv('cron_ident_regex', 'crond?')
+
+# expected format:
+# <date> <hostname> CRON[<pid>]: (<user>) CMD ...
+# example:
+# Aug 16 22:00:01 zap CRON[23060]: (root) CMD (/usr/local/bin/dnsconfig -s)
+cron_syslog_regex = re.compile('^.* %s %s\[[0-9]*\]: \((.*?)\)'  # noqa: W605
+                               % (socket.gethostname(), cron_ident_regex),
+                               re.I)
+
+STATEFILE = os.getenv('MUNIN_STATEFILE')
+
+
+def loadstate():
+    if not os.path.isfile(STATEFILE):
+        return None
+
+    with open(STATEFILE, 'rb') as f:
+        tstamp = struct.unpack('d', f.read())[0]
+        return datetime.fromtimestamp(tstamp, tz=timezone.utc)
+
+
+def savestate(state):
+    with open(STATEFILE, 'wb') as f:
+        f.write(struct.pack('d', state.timestamp()))
+
+
+def extrcronuser(line: str):
+    t = cron_syslog_regex.search(line)
+    if t is None:
+        return None
+    gr = t.groups()
+    return gr[0] if len(gr) == 1 else None
+
+
+def getcronlines():
+    with open(syslog_path, 'r') as f:
+        for x in f.readlines():
+            user = extrcronuser(x)
+            if user is not None:
+                yield (x, user)
+
+
+def getcronusers(lines):
+    return set(x[1] for x in lines)
+
+
+def autoconf():
+    if shutil.which('crontab') is None:
+        print("no (need cron installed)")
+    elif not os.access(syslog_path, os.R_OK):
+        print("no (cannot access syslog file '%s')" % syslog_path)
+    else:
+        print("yes")
+
+
+def config():
+    usernames = getcronusers(getcronlines())
+    print("""\
+graph_title Cron jobs per user
+graph_vlabel jobs
+graph_category processes""")
+    for n in usernames:
+        print(n + ".label " + n)
+        print(n + ".info jobs of user " + n)
+
+
+def fetch():
+    STATE = loadstate()
+    # why is there no stdlib function for this?!
+    localtz = timezone(timedelta(seconds=time.localtime().tm_gmtoff))
+    now = datetime.now()
+    now_withtz = datetime.now(tz=localtz)
+    yearsfx = ' ' + str(now.year)
+    pyearsfx = ' ' + str(now.year - 1)
+    cronlines = list(getcronlines())
+    allnames = getcronusers(cronlines)
+    counts = {}
+
+    hostname = socket.gethostname()
+    for ln, name in cronlines:
+        datestr = ln[:ln.index(hostname)].strip()
+        logdate = datetime.strptime(datestr + yearsfx, "%b %d %H:%M:%S %Y")
+        if logdate > now:
+            logdate = datetime.strptime(datestr + pyearsfx, "%b %d %H:%M:%S %Y")
+        # add timezone info (ugly hack), as strptime doesn't want to do this
+        logdate = now_withtz + (logdate - now)
+        if STATE is None or logdate > STATE:
+            counts[name] = (counts[name] + 1) if name in counts else 1
+
+    for n in allnames:
+        if n in counts:
+            print("%s.value %d" % (n, counts[n]))
+        else:
+            print("%s.value 0" % n)
+
+    savestate(now_withtz)
+
+
+if len(sys.argv) >= 2:
+    if sys.argv[1] == 'autoconf':
+        autoconf()
+    elif sys.argv[1] == 'config':
+        config()
+    else:
+        fetch()
+else:
+    fetch()


plugins/user/cronjobs
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
+# -*- python -*-
+# -*- coding: utf-8 -*-
```
sumpfralle, line 3: This is python2-only, I guess.

plugins/user/cronjobs
```diff
@@ -0,0 +1,93 @@
...
+#
+# Plugin to monitor the number of cronjobs running per user
```
sumpfralle, line 5: Please use a perldoc header.

plugins/user/cronjobs
```diff
@@ -0,0 +1,93 @@
...
+#
```
sumpfralle, line 6: Since the plugin uses the content of `/var/log/syslog`, you should mention this in the plugin description. This gives the users of this plugin a better idea, under which circumstances this plugin works for them.

plugins/user/cronjobs
```diff
@@ -0,0 +1,93 @@
...
+
+def isokline(ln: str):
+    if "CRON" not in ln:
+        return None
+    sp = ln.split()
+    return sp if sp[5][0] == '(' and sp[5][-1] == ')' and sp[5] != "(root)" else None
+
+
+def getfilelines():
+    with open('/var/log/syslog', 'r') as f:
```
sumpfralle, line 36: This location should be configurable.

plugins/user/cronjobs
```diff
@@ -0,0 +1,93 @@
...
+from datetime import datetime, timedelta
+import shutil
+import sys
+
+
+def isokline(ln: str):
+    if "CRON" not in ln:
+        return None
+    sp = ln.split()
+    return sp if sp[5][0] == '(' and sp[5][-1] == ')' and sp[5] != "(root)" else None
```
sumpfralle, line 32: I think, this would be more stable and better readable, if you would use a regular expression instead.

In any case: please document the expected line format as an example in a comment. This helps future co-authors of this plugin to understand the parsing.

plugins/user/cronjobs
```diff
@@ -0,0 +1,93 @@
...
+
+
+def getfilelines():
+    with open('/var/log/syslog', 'r') as f:
+        return [x for x in map(isokline, f.readlines()) if x is not None]
+
+
+def autoconf():
+    print("no (need cron installed)" if shutil.which('cron') is None else "yes")
```
sumpfralle, line 41: Is `cron` the only supported implementation here? How about `chrony` or `anacron`? How about parsing the syslog file instead and checking its content?

Additionally you should test, whether you can read the configured log file.

plugins/user/cronjobs
```diff
@@ -0,0 +1,93 @@
...
+
+
+def config():
+    usernames = set(map(lambda e: e[5][1:-1], getfilelines()))
```
sumpfralle, line 45: Maybe move the interpretation of the lines into `getfilelines` (and rename it), in order to create a simpler interface?

plugins/user/cronjobs
```diff
@@ -0,0 +1,93 @@
...
+        if logdate > now:
+            logdate = datetime.strptime(datestr + pyearsfx, "%b %d %H:%M:%S %Y")
+        if now - logdate < cutoff:
+            counts[lnm] = (counts[lnm] + 1) if lnm in counts else 1
+
+    for n in allnames:
+        if n in counts:
+            print(n + ".value " + str(counts[n]))
+        else:
+            print(n + ".value 0")
```
sumpfralle, line 80: Is this really zero? Or do you mean *unknown*? (this would be `U` for munin)

plugins/user/cronjobs
```diff
@@ -0,0 +1,93 @@
...
+
+
+def fetch():
+    cutoff = timedelta(minutes=5)
+    now = datetime.now()
+    yearsfx = ' ' + str(now.year)
+    pyearsfx = ' ' + str(now.year - 1)
+    filelines = getfilelines()
+    allnames = set(map(lambda e: e[5][1:-1], filelines))
+    counts = dict({})
```
sumpfralle, line 65: maybe just `counts = {}`?

plugins/user/cronjobs
```diff
@@ -0,0 +1,93 @@
...
+""")
+    for n in usernames:
+        print(n + ".label " + n)
+        print(n + ".info jobs of user " + n)
+        print(n + ".type GAUGE")
+        print(n + ".draw LINE")
+
+
+def fetch():
+    cutoff = timedelta(minutes=5)
```
sumpfralle, line 59: This approach is a bit tricky, since it relies on the munin-node being requested in a static interval (exactly every 5 minutes). It will cause overlaps, if the timing of the munin master request is not perfectly stable (e.g. random order of requesting nodes  or varying timing of processing the plugins *before* this one).

The perl module `Munin::Plugin` offers an `open_tail` function for this purpose. Sadly there is no similar implementation provided for python plugins.

If you cannot find a way around this potential instability, then I would suggest to document this in the plugins perldoc header clearly.

plugins/user/cronjobs
PoroCYon: It is most likely zero, it just means that there wasn't a cronjob of this particular user running *in the last 5 minutes*, which is very likely, as most cronjobs don't run that often.

Secondly, the user most likely has such cronjobs occurring, otherwise their username wouldn't've been found in the syslog in the first place.

plugins/user/cronjobs
PoroCYon: For `journald`, it'd be possible to use the `--show-cursor`/`--after-cursor=` arguments, but I don't want to rely on `journalctl` being installed.

Another approach would be storing the timestamp of the last run time in a pickle and writing that to the statefile, and comparing that to the current timestamp in the next run. But that feels a bit dirty, so would that be ok?

plugins/user/cronjobs
PoroCYon: Removed.

plugins/user/cronjobs
PoroCYon: Done. (I hope I'm using the correct `=head1` names?)

plugins/user/cronjobs
PoroCYon: Done.

plugins/user/cronjobs
PoroCYon: Done.

plugins/user/cronjobs
PoroCYon: Done and done.

plugins/user/cronjobs
PoroCYon: Also tested it with dcron, only needed to change the syslog tag name, so I made that configurable and now I just check the presence of `crontab` (which every implementation seems to have).

Also checking whether that file is readable now.

plugins/user/cronjobs
PoroCYon: Done (reworked it a little bit). Can't exactly move the username filtering to (what was) `getfilelines`, as I also use it in `config` for checking timestamps as well.

plugins/user/cronjobs
PoroCYon: Done. (Wasn't sure if just `{}` would construct a set or a dict.)

plugins/user/cronjobs
sumpfralle: Yes, they look good to me.

plugins/user/cronjobs
sumpfralle: Yes, sadly the beautiful `set` datatype lacks a shorthand initializer ...

plugins/user/cronjobs
sumpfralle: Yes, using a statefile would be a suitable approach in my opinion. I would suggest to just store the timestamp (without pickling) or using a json format, if you feel like.

This would indeed solve the fragility, that I felt when reading the plugin.

The rest of the plugin looks great after your latest changes. Thanks!

PoroCYon: Added the statefile now. I wanted to use pickle or sth instead because parsing/serializing datetimes to strings has proven to be annoying once again. But it's done this way instead now.

plugins/user/cronjobs
```diff
@@ -0,0 +1,193 @@
...
+                               % (socket.gethostname(), cron_ident_regex),
+                               re.I)
+
+STATEFILE = os.getenv('MUNIN_STATEFILE')
+STATE = None
+
+
+if os.path.isfile(STATEFILE):
+    with open(STATEFILE, 'r') as f:
+        if sys.version_info >= (3, 7):
```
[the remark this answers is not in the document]
PoroCYon: I guess an epoch value might work.

Also, I just thought, the log file position: wouldn't I also have to check and compare the mtime against the last run, because sometimes after a logrotate, the file might be large enough again for the seek position to be ok. (Won't happen very often, but still, not completely impossible.) So I'd have to store a timestamp regardless.

plugins/user/cronjobs
```diff
@@ -0,0 +1,193 @@
...
+        print(n + ".label " + n)
+        print(n + ".info jobs of user " + n)
+        print(n + ".type GAUGE")
+        print(n + ".draw LINE")
+
+    savestate()
+
+
+def fetch():
+    global STATE
```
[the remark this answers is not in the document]
PoroCYon: Done.

plugins/user/cronjobs
```diff
@@ -0,0 +1,193 @@
...
+    if sys.argv[1] == 'autoconf':
+        autoconf()
+    elif sys.argv[1] == 'config':
+        config()
+    else:
+        fetch()
+else:
+    fetch()
+
+# flake8: noqa: E265
```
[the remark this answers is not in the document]
PoroCYon: Done.

plugins/user/cronjobs
```diff
@@ -0,0 +1,193 @@
...
+graph_title Cron jobs per user
+graph_vlabel jobs
+graph_category processes""")
+    for n in usernames:
+        print(n + ".label " + n)
+        print(n + ".info jobs of user " + n)
+        print(n + ".type GAUGE")
+        print(n + ".draw LINE")
+
+    savestate()
```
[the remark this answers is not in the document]
PoroCYon: Done (added an if+parameter)

plugins/user/cronjobs
```diff
@@ -0,0 +1,193 @@
...
+        print("no (cannot access syslog file '%s')" % syslog_path)
+    else:
+        print("yes")
+
+
+def config():
+    global STATE
+
+    localtz = timezone(timedelta(seconds=time.localtime().tm_gmtoff))
+    STATE = datetime.now(tz=localtz)
```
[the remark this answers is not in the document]
PoroCYon: Removed.

plugins/user/cronjobs
```diff
@@ -0,0 +1,193 @@
...
+
+    usernames = getcronusers(getcronlines())
+    print("""\
+graph_title Cron jobs per user
+graph_vlabel jobs
+graph_category processes""")
+    for n in usernames:
+        print(n + ".label " + n)
+        print(n + ".info jobs of user " + n)
+        print(n + ".type GAUGE")
```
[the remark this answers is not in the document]
PoroCYon: Removed.

sumpfralle:
Thanks!

PR #1109, opened 2020-07-31
[plugins/user/cpubyuser] Bugfixes and enhancements

PoroCYon:
- Fix USERS=ALL only enumerating processes with a tty or pty attached
  (and thus not eg. user-owned FastCGI processes, owned by eg. Apache)
- ps(1) is used to make process list parsing easier and more robust
- OTHERS_FIELD now defaults to disabled

### DIFF
--- a/plugins/user/cpubyuser
+++ b/plugins/user/cpubyuser
@@ -2,15 +2,26 @@
 #
 # Plugin to monitor CPU usage, for a selected set of users
 #
-# Usage: Place in /etc/munin/node.d/ (or link it there using ln -s)
+# Usage: Place in /etc/munin/plugins/ (or link it there using ln -s)
 #	Add this to your /etc/munin/plugin-conf.d/munin-node:
 #		[cpubyuser]
+#		user root # required if /proc can't be read from by any user!
 #		env.USERS root yann
+#		env.OTHER_FIELD others # enable 'others'-list, set the label/field name
 #
-#	If env.USERS is set to ALL, count all logged in users.
+#	root and yann being a list of the users to monitor.
 #
-#	 root and yann being a list of the users to monitor.
-#	 You need to also make sure that awk is installed
+#	If env.USERS is set to ALL, count all logged in users, and if set to
+#	ALLPROC, *all* users with a running process will be listed, except for root.
+#
+#	You need to also make sure that awk is installed
+#
+# 2019-08-30 v 1.4 pcy <pcy@ulyssis.org>:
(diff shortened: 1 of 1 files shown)


plugins/user/cpubyuser
```diff
@@ -39,9 +48,12 @@
 
 . "$MUNIN_LIBDIR/plugins/plugin.sh"
 
-OTHER_FIELD="others"
-[ "$USERS" = "ALL" ] && USERS=$(w --no-header | awk '{ print $1 }' | sort | uniq)
-
+#[ -z "$OTHER_FIELD" ] && OTHER_FIELD="others"
```
sumpfralle, line 51: Please remove commented code.

plugins/user/cpubyuser
```diff
@@ -55,50 +67,48 @@ fi
 if [ "$1" = "config" ]; then
 	echo "graph_args --base 1000 -r --lower-limit 0"
 	echo "graph_title CPU usage, by user"
-	echo "graph_category cpu"
+	echo "graph_category system"
```
sumpfralle, line 70: I have the feeling that `cpu` is quite suitable here, since the plugin focuses on the CPU usage.

plugins/user/cpubyuser
```diff
@@ -39,9 +48,12 @@
 
 . "$MUNIN_LIBDIR/plugins/plugin.sh"
 
-OTHER_FIELD="others"
-[ "$USERS" = "ALL" ] && USERS=$(w --no-header | awk '{ print $1 }' | sort | uniq)
-
+#[ -z "$OTHER_FIELD" ] && OTHER_FIELD="others"
+[ -z "$USERS" ] && USERS=ALL
+if [ "$USERS" = "ALL" ]; then
+	LU="$(ps ax --format uname | tail +2 | sort -u | grep -v -e '^root$')"
```
sumpfralle, line 54: The current documentation says, that only logged in users are accounted, if `ALL` is given. Your approach would change this.

Instead I would suggest to add another configuration variable, that allows to pick the suitable approach (and keeps the current behaviour unchanged).

plugins/user/cpubyuser
PoroCYon: Rolled back.

plugins/user/cpubyuser
PoroCYon: Done.

plugins/user/cpubyuser
PoroCYon: Done.

sumpfralle:
Thank you!

PR #1110, opened 2020-08-04
Added env to exclude containers from cpu/memory docker graphs

rwky:
It maybe desirable to not generate stats for ephemeral containers.
This commit adds an env variable EXCLUDE which is a regex, if it matches
then the container is excluded from the cpu and memory graphs.

### DIFF
--- a/plugins/docker/docker_
+++ b/plugins/docker/docker_
@@ -38,11 +38,22 @@ directory /etc/munin/plugin-conf.d/ with the following config (you can also use
 Docker environment variables here as described in
 https://docs.docker.com/compose/reference/envvars/):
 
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thanks for your contribution!

I have three suggestions:
* maybe rename `EXCLUDE` to something more specific - maybe `EXCLUDE_CONTAINER_NAME`?
* you are using `re.match` instead of `re.search` - I could imagine, that this is surprising for users?
* please add a short explanation of this new configuration setting to the documentation header

What do you think?

rwky:
All good points! MR updated.

sumpfralle:

PR #1111, opened 2020-08-08
[dm_cache_] support volumes with multiple dashes

zivillian:
Promox is creating logical volumes with names like `vm--100--disk--0`


sumpfralle:
Your change looks good to me. Thank you!

PR #1113, opened 2020-08-22
plugins/puppet/puppet_runtime: fix formatting

kenyon:
Used [rufo](https://github.com/ruby-formatter/rufo) to correct the formatting (it had a mix of tabs and spaces).

### DIFF
--- a/plugins/puppet/puppet_runtime
+++ b/plugins/puppet/puppet_runtime
@@ -15,9 +15,9 @@
 
 # reports how long the puppet agent took to apply the catalog
 def get_runtime
-  logfile = ENV['puppet_logfile'] || '/var/log/messages'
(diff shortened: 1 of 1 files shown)


sumpfralle:
> Yes, rubocop.

Ah - that looks good (similar to `flake8` and `black` for python).

I just applied the hints from rubocop for the plugins in `contrib`: b0b39b018e200cd0b771675fd0a6489f6ee9f045 and 809639ab65d5460442c3b88beea0dace7665f416. Additionally I enabled such checks for existing and future plugins (758c3ecdd7c6a709af449e2fc78571e28c7fe508).

Sadly at the moment *every* existing ruby-based plugin (all 33 of them) fail for at least one check. Thus these plugins are added to the exemption list for now. But anyway: new plugins will trigger a test failure and thus may help to improve the quality in the long term.

PR #1114, opened 2020-08-23
[plugins/disk/] Add plugins to monitor btrfs pools on a per device base

HaseHarald:
Graph the usage (in bytes and percent) and the device statistics (mainly error counters) of the devices in a btrfs pool.

I've written these plugins to scratch my own itch and have been using them for about a year now. They work great for me, but that doesn't mean they are perfect.

### DIFF
--- a/plugins/disk/btrfs_device_stats
+++ b/plugins/disk/btrfs_device_stats
@@ -11,7 +11,10 @@ btrfs_device_stats - Script to monitor btrfs device statistics
 =head1 CONFIGURATION
 
 Simply create a symlink in your plugins directory like with any other plugin.
-No configuration needed.
+Must be run as root.
+
+[btrfs_device_stats]
+user root
 
 =head2 DEFAULT CONFIGURATION
 
@@ -24,7 +27,7 @@ No configuration needed.
 =head1 MAGIC MARKERS
 
  #%# family=auto
- #%# capabilities=autoconf suggest
+ #%# capabilities=autoconf
 
 =head1 LICENSE
 
--- a/plugins/disk/btrfs_device_stats
+++ b/plugins/disk/btrfs_device_stats
@@ -80,7 +80,6 @@ def munin_values(fs):
     devices = fs.devices()
     
     for this_device in devices:
-        this_dev_info = fs.dev_info(this_device.devid)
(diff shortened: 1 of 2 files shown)


sumpfralle:
Thank you for contributing these plugins!

Your plugins look good to me. The only minor detail is the lack of a perldoc-style header (used by `munin-get` and the plugin gallery for describing the plugin). You may take a look at other plugins providing it, e.g. `zfs/zfs_usage_`. In python you would put it into a multi-line documentation string at the file level.

Regarding `btrfs_device_usage`: is the scope of this plugin similar to the existing `disk/btrfs_subvol_usage`? I am not used to the btrfs terms - thus I could not tell the difference.

Thanks!

HaseHarald:
Hm, ok, forgive me if the following explanation is not entirely correct. But I hope to get the point across without getting in to too much detail.

btrfs is capable of pooling multiple physical disks (or any block device) into one big filesystem and care about redundancy, checksumming and stuff like that. It's often compared to ZFS in this regard, even though this comparison is not really fair.

btrfs_subvol_usage gets statistics about the usage over the entire filesystem and differentiates between actual data, metadata and such. But it always looks on the entire filesystem, regardless on the devices underneath. So it will tell you the entire filesystem is using 10.1 TB and 512 GB of that are metadata of which only 148 GB are really used. (Numbers are just an example.)

btrfs_usage-day

btrfs_device_usage looks at the usage on each device. So it will tell you that sda1 has 2.75 TB of used storage, and sdb1 and sdd1 are using 2.79 TB each and sdc1 only uses 1.84 TB. It will not tell you how much of that is metadata and stuff like that, because I honestly don't know if you can even figure that out (would be a great improvement though). But it's usage on a per device bases instead of the entire filesystem. This gets more interesting when you grow the pool over time or exchange failed disks because then the values for each disk start to really vary.

btrfs_device_usage_byte-day

By the way, btrfs_subvol_usage is a very useful plugin! Love it.

sumpfralle:
Thank you for your finishing touches and for the explanation of the different plugins!

I took the freedom to add your two example graphs to the suitable location (see  bd872f7af1e3076376fcdcac2dec3c63403fb573).

PR #1115, opened 2020-08-30
[plugins/libvirt/kvm_mem] remove empty line from config

zivillian:
fix `[WARNING] 1 lines had errors while 66 lines were correct in data from 'config kvm_mem' on <host>`

### DIFF
--- a/plugins/libvirt/kvm_mem
+++ b/plugins/libvirt/kvm_mem
@@ -25,8 +25,7 @@ def config(vm_names):
 graph_vlabel Bytes
 graph_category virtualization
 graph_info This graph shows the current amount of memory used by virtual machines
-graph_args --base 1024
-    """
+graph_args --base 1024"""
     print base_config
     draw = "AREA"
     for vm in vm_names:
--- a/plugins/libvirt/kvm_mem
+++ b/plugins/libvirt/kvm_mem
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 # -*- coding: utf-8 -*-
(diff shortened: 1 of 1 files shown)


sumpfralle:
Are you sure, that the error was related to the empty line?

In general empty lines should not cause problems.

Please verify, that the error is gone with your change and that it appears again without your change.

I would still be confused, but I would merge it anyway, if it fixes a real issue ...

btw: in case you want to improve the plugin, then I just saw a few possible spots needing help:
* line 38 seems to be mis-indented (I am confused, that python does not complain about it)
* the `AREA`/`STACK` handling around line 31 can be simplified by using `AREASTACK` everywhere (it was introduced quite some time ago)
* switch to python3 (switching `print FOO` to `print(FOO)` should suffice ...)

Your contribution would be very welcome!

zivillian:
I've reverted the fix and the warning reappeared - so I'm sure this fixes the warning. I guess the reason is not the newline, but the additional spaces.

I've also addressed your suggestions.

sumpfralle:
> I've reverted the fix and the warning reappeared - so I'm sure this fixes the warning. I guess the reason is not the newline, but the additional spaces.

Thanks for verifying this - yes, maybe the spaces were the source of the issue.

> I've also addressed your suggestions.

Great!

I you want to do the final step of the switch to `python3`, then please change the shebang and check, whether the plugin still works. That would be one down for the number of python2 plugins :)

zivillian:
I've switched the plugin to python3 and confirmed it's still working.

sumpfralle:
Great - thank you!

PR #1117, opened 2020-09-03
Nextcloud fixups

shtrom:


plugins/nextcloud/nextcloud_
```diff
@@ -47,15 +50,23 @@ GPLv2
...
+API_PATH="${api_path:-/ocs/v2.php/apps/serverinfo/api/v1/info}?format=json"
 DOMAIN="${0##*nextcloud_}"
+TIMEOUT="${timeout:-2}"
+UPDATES_WARNING="${updates_warning:-1}"
 CLEANDOMAIN="$(clean_fieldname "$DOMAIN")"
 USERNAME="${username:-}"
 PASSWORD="${password:-}"
 
 print_json_data() {
+    # shellcheck disable=SC2039
```
sumpfralle, line 69: You may omit this, if you like, since `shellcheck` is called as a CI test with the arguments `--exclude=SC1090,SC2009,SC2126,SC2230 --shell=dash` (see `t/test.t`). Sadly this is not configurable in a separate file, thus there is no independent way of running `shellcheck` for a separate file withe the arguments used by the CI job.

plugins/nextcloud/nextcloud_
```diff
@@ -47,15 +50,23 @@ GPLv2
...
     local FIRST="$1"
     [ -z "$FIRST" ] && exit 0
```
sumpfralle, line 71: This is unrelated to your current suggestions, but maybe you could check, whether `exit 0` is really the proper approach here (instead of `return 0`).

plugins/nextcloud/nextcloud_
```diff
@@ -201,8 +215,10 @@ EOM
 esac
 
 # Get JSON data
-[ "$(test_https)" ] && DOMAIN="https://$DOMAIN" || DOMAIN="http://$DOMAIN"
-JSONSTATS=$(curl -s -f -m 2 -u "$USERNAME:$PASSWORD" "$DOMAIN$API_PATH" | sed 's/\\/\\\\/g' | jq -cr ".ocs.data")
+JSONSTATS=$(
+	fetch_url -u "$USERNAME:$PASSWORD" "https://${DOMAIN}${API_PATH}" | sed 's/\\/\\\\/g' | jq -cr ".ocs.data" 2>&1 \
```
sumpfralle, line 219: How about `{ fetch_url ... || fetch_url ...; } | sed ...` instead?

sumpfralle (commented):
Thanks for your suggestions!

Please take a look at my comments below. But the current state is also fit for merge, I would say ...

plugins/nextcloud/nextcloud_
shtrom: I added a `# shellcheck shell=dash` directive

plugins/nextcloud/nextcloud_
shtrom: The detection is a bit dodgy, and depends on whether we are able to parse JSON out of the received data, so we need to go all the way through `sed` and `jq` before we know if we've received what we want.

plugins/nextcloud/nextcloud_
shtrom: It's a bit barbaric, but it looks like it's protecting against miscall from other places in the code. It kinda makes sense to use `exit` to terminate the script, rather than `return`, which would let it continue with weird values, but maybe it should actually be a non-zero exit status.

What do you think?

sumpfralle: > *barbaric*

:)

How about a message to stderr in order to not die silently?

plugins/nextcloud/nextcloud_
sumpfralle: I think, `curl --fail` (as you used above) should already take care for the proper detection of the API's availability. Or maybe you should add (`--location`?) in order to handle redirects (which could look like success, but have an empty body)?

Anyway: feel free to keep it, if you feel like.

shtrom: Alright, I bit that bullet. I was about to explain why it was done like this, and how it could be done better. Then I figured it would be best to actually do it. (:+1: 

So I added a `scheme` configuration variable (defaults to `https` if not set), and that removes all the uncertainty about what to hit, and the weird duplicates.

The CAVEAT is that it's a breaking change for servers where Nextcloud is only available over HTTP. It would have previously been autodetected, while it will now be required for `scheme http` to be present in the config for the plugin.

sumpfralle:
I appreciate your move to using https by default - it should be reasonable and common nowadays, thus it will probably not hurt.

But there is just one last minor nitpick, that sneaked into your changeset: you enclosed `exit 1` in brackets (`( ... )`). This defeats the purpose of `exit`, since it will only exit the sub-shell spawned by the brackets. Instead you should use curly brackets instead (no sub-shell).

Apart from that, I am happy to merge your improvements.

shtrom:
Ah, you are right. I tested it, and the `exit 1` didn't bubble up, but the script failed later while parsing data. I realised that I misunderstood the check, as it actually verifies that some data was received for the requested fields.

I just reworked that bit, so it just outputs `U` for all the requested keys if data is missing.

sumpfralle:
Perfect - thanks!

Issue #1095, opened 2020-07-07
unifi_api: Error for use of uninitialized values


kboenke:
Running `unifi_api` generally works, but produces some errors during each run:
> me@host:/etc/munin# /usr/sbin/munin-run unifi_api
> Use of uninitialized value in addition (+) at /etc/munin/plugins/unifi_api line 1036.

Values are generated as expected. `errors` are not reported, likely due to this error. I am not familiar with the structure of the JSON being processed, but looking at the code it might be as easy as assuming 0 values if the return-value is empty:
> 'err'   => ($_->{$name . '-rx_errors'} **|| 0**) + ($_->{$name . '-tx_errors'} **|| 0**),

@jtsage can likely assess this easily.

sumpfralle:
Thanks for your proposal.

I assume, this is the proper way of dealing with the response.

@jtsage: feel free to re-open, if you see a better way.

jtsage:
oh, sweet.  @kboenke didn't realize you had already done a pull request, was trying to find time to come visit this.  Works for me, thanks!!

~j

PR #1121, opened 2020-09-10
Fixed data for UAP

kboenke:
Data for UAP's is (now?) encapsulated within `stat->ap`, changed reference to reflect this change.
Also: Completely implemented handling of uninitialized values in case they do not exist (`// 0`) as suggested with #1095.

PS: I don't remember ever writting a single line of Perl-code. So for the sake of all that's holy, please double-check this pull-request! (_That being said, it seems to work on my box._)


sumpfralle:
My perl is also rusty - but I think, your suggestion is fine. Thanks!

PR #1123, opened 2020-09-22
Add plugins for Arris Cable Modem SB6183

utopiabound:
This provides two plugins:
* arris-sb6183 - Power levels, SNR, error counts
* arris-sb6183_uptime - Uptime in days

These are two seperate plugins because it's two different
urls it needs to hit, and this seemed simpler.

arris-sb6183 provides multigraphs for each channel.

Signed-off-by: Nathaniel Clark <Nathaniel.Clark@misrule.us>


plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
+#!/usr/bin/python
```
sumpfralle, line 1: I think, you are using python3 syntax here. Thus please clarify this by using `python3`. Additionally it is recommended to allow local overrides of the specific python interpreter (including a virtualenv) by using `#!/usr/bin/env python3` instead.

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
...
+
+# modem:
```
sumpfralle, line 3: Please move these comments into the perldoc documentation below.

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
...
+ #%# capabilities=autoconf
+
+=cut
+"""
+
+import re
+import os
+import sys
+import requests
+from lxml import html
```
sumpfralle, line 61: I would suggest to use a standard library instead (e.g. https://docs.python.org/3/library/xml.dom.html). Otherwise you force users to install an additional package and you will need to handle the potentially missing library in the code below (by failing with a reasonable error message in case of an `ImportError`).

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
...
+
+
+URL = os.getenv("url", "http://192.168.100.1/RgConnect.asp")
```
sumpfralle, line 64: I would suggest to reduce the configurable part of the URL to the scheme and the host (omitting the path). This will allow you to adjust to a changed URL provided by a later version of a firmware within the plugin code (without the user being forced to change his configuration). For example: `http://192.168.100.1`.

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
...
+Make sure 192.168.100.1 is accessible through your firewall.
+
+To have this register with munin as it's own host set the "env.hostname" in config.
+Also ensure that the hostname set is listed in munin.conf.
+
+[arris*]
+env.hostname modem
+
+=head1 VERSION
+
```
sumpfralle, line 38: Please mention the other configuration settings (`url`, `up`, `down`) here, too.

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
...
+import re
+import os
+import sys
+import requests
+from lxml import html
+
+
+URL = os.getenv("url", "http://192.168.100.1/RgConnect.asp")
+HOSTNAME = os.getenv("hostname", None)
+UPCOUNT = int(os.getenv("up", 4))
```
sumpfralle, line 66: What is the purpose of this `up` variable? Is it related to slots/ports of the physical device? I would suggest to use a self-explaining variable name here (especially for the configuration variable).

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
...
+        print("graph_vlabel Power (dBmV)")
+        print("graph_category network")
+        for i in range(1, DOWNCOUNT + 1):
+            print("down_{0}.label Down Ch {1}".format(i, i))
+            print("down_{0}.type GAUGE".format(i))
+            print("down_{0}.draw LINE1".format(i))
+        for i in range(1, UPCOUNT + 1):
+            print("up_{0}.label Up Ch {1}".format(i, i))
+            print("up_{0}.type GAUGE".format(i))
+            # print("up_{0}.draw LINE1".format(i))
```
sumpfralle, line 86: Please remove commented code.

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
...
+ #%# family=contrib
+ #%# capabilities=autoconf
+
+=cut
+"""
+
+import re
+import os
+import sys
+import requests
```
sumpfralle, line 60: Maybe `urllib` (the standard library) would be sufficient here? Fewer dependencies and less code for handling the potentially missing library.

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
...
+            print("uncr.type DERIVE")
+            print("uncr.min 0")
+            print("uncr.warning 1")
+
+        sys.exit(0)
+
+    if sys.argv[1] == "autoconfig":
+        try:
+            page = requests.get(URL)
+        except:
```
sumpfralle, line 160: Please avoid bare exceptions (see [PEP8](https://www.python.org/dev/peps/pep-0008/)). I guess, `IOError` would be suitable here.

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
...
+            print("corr.min 0")
+            print("uncr.label Uncorrectable")
+            print("uncr.type DERIVE")
+            print("uncr.min 0")
+            print("uncr.warning 1")
+
+        sys.exit(0)
+
+    if sys.argv[1] == "autoconfig":
+        try:
```
sumpfralle, line 158: Here you would also need to verify the availability of the non-core modules (`lxml` and `requests`).

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,267 @@
...
+upstream = arr[2]
+
+trs = downstream.findall("tr")
+# drop title
+trs.pop(0)
+
+headings = ["".join(x.itertext()).strip() for x in trs.pop(0).findall("td")]
+# ['Channel', 'Lock Status', 'Modulation', 'Channel ID', 'Frequency', 'Power', 'SNR', 'Corrected', 'Uncorrectables']
+
+mapper = lambda x, y: (x, y)
```
sumpfralle, line 188: I guess, `zip` would do the same, or? (in a list context)

sumpfralle (commented):
Thank you for contributing these plugins!

Please take a look at my notes below.

Regarding the separate plugins: I think, it would be very beneficial to merge both plugins into one (since you are using `multigraph` already). The separate URLs (or *paths*) should be no issue, since it would be advisable to configure the URL without the path, anyway (see one of my comments below). This would reduce the burden of maintaining two copies of similar code.

Maybe you also want to add [examples graphs](http://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-gallery.html#rules-for-plugin-contributors)? This would be perfect.

Thank you!

plugins/router/arris-sb6183
utopiabound: I would have liked to, but the HTML from the router is poorly structured and xml.etree throws up on parsing it.

utopiabound:
Example Output:
arris_error-day
arris_snr-day
arris_uptime-day

I've updated plug-in to be single file, and added more error handling.

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,366 @@
...
+            "failed to get status page %d: %s" % (resp.status, resp.reason),
+            file=sys.stderr,
+        )
+        return []
+    data = rxscript.sub(
+        "",
+        rxcomment.sub(
+            "",
+            rxblank.sub(
+                " ", "".join(map(lambda x: x.decode("utf-8"), resp.readlines()))
```
sumpfralle, line 241: Just a style recommendation - maybe a generator comprehension would be easier to read:
```python
"".join(line.decode() for line in resp.readlines())
```

Or maybe just use `resp.read().decode()`?

plugins/router/arris-sb6183
```diff
@@ -0,0 +1,366 @@
...
+=cut
+"""
+
+import re
+import os
+import sys
+from urllib import request
+
+HOSTNAME = os.getenv("hostname", None)
+STATUS_URL = "http://192.168.100.1/RgConnect.asp"
```
sumpfralle, line 78: Are you sure, that the IP address of the router is always the same? Or maybe make it configurable?

(I leave this up to you, since you know that device ...)

sumpfralle (commented):
Thank you for merging these plugins into one!

I added two more minor comments below. But the plugin is ready to be merged, if you prefer that.

Regarding the example graphs: please add them to the sub directory `example-graphs` (following the naming convention described in the [munin plugin documentation](http://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-gallery.html#rules-for-plugin-contributors)).

Thanks!

plugins/router/arris-sb6183
utopiabound: The documentation lists this as default and every router I've had from Arris has had this IP.  (There is no way *I* can change the IP, maybe my cable company can)

niclan: The 192.168.100.1 seems to be industry standard.

sumpfralle: A local IP that cannot be changed by the local administrator/user. That sounds horrible :(

Anyway: in this case it is not necessary to make it configurable.

sumpfralle:
Thank you for the new plugin - I just merged it now.

Feel free to add the example graphs in a separate pull request, if you feel like ...

PR #1124, opened 2020-10-28
avoid blocking in ssl_

gnuheidix:
Just in case there is a firewall dropping packages, this plugin
would block until the plugin timeout is reached. This change introduces
the possibility to configure a much lower individual timeout.


sumpfralle:
This looks like a good change to me. I guess, the `coreutils` package is common enough to allow this additional dependency.

PR #1125, opened 2020-10-28
avoid blocking in http_responsecode

gnuheidix:
Just in case there is a firewall dropping packages, this plugin
would block until the plugin timeout is reached. This change introduces
the possibility to configure a much lower individual timeout.


sumpfralle:
This looks like a good change - thank you!

PR #1135, opened 2020-11-26
[plugins/whois_] stop EXTRACT_RE after the first match

shtrom:


sumpfralle:
Thanks!

PR #1136, opened 2020-11-26
[plugins/transmission] Python3 compat

shtrom:

### DIFF
--- a/plugins/network/transmission
+++ b/plugins/network/transmission
@@ -121,7 +121,7 @@ paused.draw STACK
 paused.min 0
 paused.colour COLOUR8
 """)
-    print conf.safe_substitute(plugin_name=plugin_name, host=host, title_host=title_host)
+    print(conf.safe_substitute(plugin_name=plugin_name, host=host, title_host=title_host))
     sys.exit(0)
 
 
@@ -130,7 +130,7 @@ def autoconf():
         import transmissionrpc
         print('yes')
     except ImportError:
-        print 'no python module \'transmissionrpc\' missing'
+        print('no python module \'transmissionrpc\' missing')
 
 
 def fetch():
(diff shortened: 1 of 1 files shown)


sumpfralle:
That looks good!

How about switching the shebang to `python3`? (Python2 is end-of-life anyway)

shtrom:
Yup. I didn't want to be too destructive for the moment, as the changes are backward compatible, and still work with Python2, too. So I decided to remain on the fence (:

sumpfralle:
Following [PEP-394](https://www.python.org/dev/peps/pep-0394/#for-python-script-publishers) I would like to replace the `python` shebang with either `python2` or `python3` (since we do not expect the plugin to be executed in a virtual environment).

Thus if you hesitate to change the shebang, then I would do it afterwards :)

shtrom:
Ah, I wasn't aware of the PEP. I actually did run into this issue on a Debian machine of mine, where `python` disappeared, and I had to install `python-is-python3`, which then highlighted this issue.

The code as it stands is compatible with both Python 2 and 3, so I'm not sure whether it's great for the users to potentially force a move to a new Python while the rest of their system still hasn't been updated. Arguably, this is only for old systems, but this would force a breaking change in the plugin that is not strictly necessary.

Unless we have a strict requirement in munin or munin-contrib to have Python _3_ installed, I think it would be less disruptive to stick to `python` in the shebang for now.

sumpfralle:
> Arguably, this is only for old systems, but this would force a breaking change in the plugin that is not strictly necessary.

The last actively maintained distribution, that *does not* ship python3 is RHEL5. It will be end-of-life next week.

Thus within 72 hours there will be *no* distribution, that does not contain python3.

From my point of view, everything pushes into the direction of `python3`:
* PEP 394 recommends it
* Debian and derivatives need manual interaction for providing `python`
* every distribution released in the last few years has more likely `python3` installed, than `python`

It is about time to move on :)

shtrom:
On Fri 27 Nov 2020 at 04:08:18 -0800, Lars Kruse wrote:
>It is about time to move on :)

Oh, yeah, that's for sure. My concern is more about breaking old systems 
running in the wild that may not actually get upgraded.

But ok, I'll update the shebang.

-- 

sumpfralle:
Thanks!

PR #1137, opened 2020-11-27
[plugins/nextcloud_] Only use one call to JQ to parse and report the …

shtrom:
…info

tl;dr: Use a single call to `jq` to format the whole output, rather than one per data point.

This greatly improves the performance of the plugin. Here's a before/after test (replacing the `curl` call with canned data, to only test the processing part).
* Before: ```sudo -u munin munin-run nextcloud_example.com  1.89s user 0.11s system 98% cpu 2.022 total```
* After: ```sudo -u munin munin-run nextcloud_example.com  0.29s user 0.03s system 93% cpu 0.337 total```

### DIFF
--- a/plugins/nextcloud/nextcloud_
+++ b/plugins/nextcloud/nextcloud_
@@ -79,7 +79,7 @@ case $1 in
 	    if [ ! -x "$(command -v curl)" ]; then
 		    echo "no (curl not found)"
 	    elif [ ! -x "$(command -v jq)" ]; then
-		    echo "no (jq not found)"exit 0
+		    echo "no (jq not found)"
 	    else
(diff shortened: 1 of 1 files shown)


plugins/nextcloud/nextcloud_
```diff
@@ -93,18 +81,18 @@ case $1 in
 	    elif [ ! -x "$(command -v jq)" ]; then
 		    echo "no (jq not found)"exit 0
```
sumpfralle, line 82: This line was already broken before?

sumpfralle (commented):
What a nice approach for using `jq` efficiently!

I noticed one small problem with the previous code. Maybe you would like to fix this now, too?

sumpfralle:
Thank you!

PR #1138, opened 2020-11-27
Add "certbot_certs"

jtsage:
This plugin will semi-auto discover certbot (letsencrypt) certificates and output their expiry in days.

At this time, it pulls any renewal active certificates.  In the future I may add an option to ignore some, or specify only those you are interested in.

### DIFF
--- a/plugins/ssl/certbot_certs
+++ b/plugins/ssl/certbot_certs
@@ -1,0 +1,156 @@
+#!/usr/bin/perl 
+# -*- perl -*-
+
+=encoding utf8
+
+=head1 NAME
+
+certbot_certs - Munin plugin to certbot (letsencrypt) certificate expiry
+
+=head1 APPLICABLE SYSTEMS
+
+Servers running certbot certificates.  This script auto-discovers the current set of certificates
+
+=head1 CONFIGURATION
+
+This script requires root permissions to run, in your munin-node configuration:
+
+  [certbot_certs]
+  user root
+
+=head1 CAPABILITIES
+
+This plugin supports L<DIRTYCONFIG|http://guide.munin-monitoring.org/en/latest/plugin/protocol-dirtyconfig.html>
+
+=head1 LICENSE
+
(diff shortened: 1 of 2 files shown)


plugins/ssl/certbot_certs
```diff
@@ -0,0 +1,156 @@
...
+
+use warnings;
+use strict;
+use Munin::Plugin;
+use Date::Parse;
+use POSIX;
+
+if ( defined($ARGV[0]) && $ARGV[0] eq "autoconf" ) {
+	if ( -e "/usr/bin/openssl" ) {
+		print "no (Root User Required)\n";
```
sumpfralle, line 57: You do not accept *yes* for an answer? :)

sumpfralle (commented):
Thank you for this beautifully crafted plugin!

I am a bit hesitant regarding the plugin filename. I would recommend to change it to `certbot_expiry`. What do you think?

plugins/ssl/certbot_certs
jtsage: Nope, but it's possible I totally misunderstood autoconf - I was under the impression that if it returns yes, it can be added to the munin-node configuration without any user intervention? (In this case, it requires elevated permissions to read the certificates).  If i'm wrong, happy to make that edit.

Re: name change - yes, good call.

sumpfralle: It is possible, that the plugin is pre-configured somehow (e.g. the Debian package for munin contains the necessary plugin configuration for all distributed plugins).

Usually plugins check whether dependencies are missing and if runtime conditions are met. In your case you could check whether the directory `/etc/letsencrypt/renewal` exists and is readable (both should be reported separately, if missing).

jtsage: That makes sense.  I fleshed out that logic a bit, and added a number of other traps to keep the plugin sane.  Thanks for the input.

sumpfralle:
That looks good - thank you!

PR #1139, opened 2020-12-01
amavis: add three more plugins to monitor amavisd mail filter

kimheino:


plugins/amavis/amavischilds
```diff
@@ -0,0 +1,66 @@
...
+GPLv2
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+
+. ${MUNIN_LIBDIR}/plugins/plugin.sh
```
sumpfralle, line 29: Please quote the path.

plugins/amavis/amavischilds
```diff
@@ -0,0 +1,66 @@
...
+
+OLD=(0 0)
+[ -f ${STATE} ] && OLD=($(cat ${STATE}))
+
+IFS=""
+/usr/bin/amavisd-nanny -c 1 2>&1 | grep '^PID [0-9]*: .*:' | sed 's/^PID [0-9]*//' | (
+    total=0
+    busy=0
+    while read line; do
+	total=$[ ${total} + 1 ]
```
sumpfralle, line 61: The bash manual says:

> The old format $[expression] is deprecated and will be removed in upcoming versions of bash.

Thus maybe switch to `$(( ... )))` instead?

plugins/amavis/amavischilds
```diff
@@ -0,0 +1,66 @@
...
+fi
+
+OLD=(0 0)
+[ -f ${STATE} ] && OLD=($(cat ${STATE}))
+
+IFS=""
+/usr/bin/amavisd-nanny -c 1 2>&1 | grep '^PID [0-9]*: .*:' | sed 's/^PID [0-9]*//' | (
+    total=0
+    busy=0
+    while read line; do
```
sumpfralle, line 60: If I remember correctly, `shellcheck` complains about a missing `-r` for `read` (*raw* input). I guess, you should add it here.

plugins/amavis/amavischilds
```diff
@@ -0,0 +1,66 @@
...
+	total=$[ ${total} + 1 ]
+	[ "${line:0:4}" != ":   " ] && busy=$[ ${busy} + 1 ]
```
sumpfralle, line 62: Just for style nitpicking: if the last statement of a loop (or conditional expression) is a `&&` list operator, then a potential non-zero exitcode leaks to the outside (becoming the result of the loop). This is problematic, if `set -e` is in effect (which should be the default for shell scripts, in my humble opinion).
Personally I would add a `true` statement below or reverse the logic. But this is really just a style opinion ...

plugins/amavis/amavischilds
```diff
@@ -0,0 +1,66 @@
...
+    done
+    echo "busy.value $[ 100 * ( ${OLD[0]} + ${OLD[1]} + ${busy} ) / ( 3 * ${total} ) ]"
```
sumpfralle, line 64: `${OLD[0]}` looks like a bash array access to me? Maybe change the shell to `bash` or replace it?

And please replace `$[ ... ]` with `$(( ... ))`, too.

plugins/amavis/amavistimes
```diff
@@ -0,0 +1,76 @@
...
+GPLv2
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+
+. $MUNIN_LIBDIR/plugins/plugin.sh
```
sumpfralle, line 29: Please quote the path.

sumpfralle (commented):
Thank you for these contributions!

Regarding the name of the plugins: maybe you want to place an underscore after `amavis`?

Do these plugins require separate tools (`amavisd-nanny`, `amavisd-agent`) or are they all shipped in a common amavisd package? In the latter case it would be nice, if you could assemble these plugins into a single multigraph plugin.

Thanks!

kimheino:
> Do these plugins require separate tools (`amavisd-nanny`, `amavisd-agent`) or are they all shipped in a common amavisd package? In the latter case it would be nice, if you could assemble these plugins into a single multigraph plugin.

They are shipped in common amavisd package.

I'll close this PR and open a new one for Grand Unified Amavis Plugin, written in python.

PR #1140, opened 2020-12-01
ssl: add new plugin to monitor certificate expiry on web servers

kimheino:
This is similar to ssl-certificate-expiry, but creates single graph.

### DIFF
--- a/plugins/ssl/certificate
+++ b/plugins/ssl/certificate
@@ -1,0 +1,126 @@
+#!/usr/bin/python3 -tt
+# -*- coding: utf-8 -*-
+# pylint: disable=invalid-name
+# pylint: enable=invalid-name
+
+"""Munin plugin to monitor certificate expiry.
+Copyright 2015, Kim B. Heino, b@bbbs.net, Foobar Oy
+License GPLv2+
+
+Requires: pyOpenSSL
+
+
+Config file:
+
+[certificate]
+env.hosts domain.invalid some.domain:888
+"""
+
+import datetime
+import os
+import socket
+import sys
+import time
+from OpenSSL import SSL
+
+
+RESULT = 0
+
+
+def cert_check(dummy_connection, x509, dummy_errnum, dummy_errdepth, dummy_ok):
+    """Callback function to get certificate expire time."""
+    after = datetime.datetime.strptime(x509.get_notAfter().decode('utf-8'),
+                                       '%Y%m%d%H%M%SZ')
+    global RESULT  # pylint: disable=global-statement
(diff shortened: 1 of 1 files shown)


plugins/ssl/certificate
```diff
@@ -0,0 +1,126 @@
...
+            print('{}.critical 1:'.format(key))
+        if show_values:
+            value = check_site(host, port)
+            if value is not None:
+                print('{}.value {}'.format(key, value))
+
+
+if __name__ == '__main__':
+    if len(sys.argv) > 1 and sys.argv[1] == 'autoconf':
+        print('no')
```
sumpfralle, line 122: Please omit the autoconf feature, if it is bound to fail.

plugins/ssl/certificate
```diff
@@ -0,0 +1,126 @@
...
+
+import datetime
+import os
+import socket
+import sys
+import time
+from OpenSSL import SSL
+
+
+RESULT = 0
```
sumpfralle, line 27: The usage of `RESULT` is a bit hard to comprehend. Maybe you can simplify it?

sumpfralle (commented):
Thank yo four this contribution!

Since there are already some SSL-related plugins, it would be good, if you could find a less generic name for this one. How about `ssl_certificate_expiry_py`?

plugins/ssl/certificate
kimheino: This plugin does not have autoconf magic marker, so munin should not run autoconf. Do you still want to remove above two lines? I rather keep them and improve "no" text. If I remove them, then plugin would print out values instead.

sumpfralle: I think, the handling of `autoconf` in the code is a source of confusion, since the plugin did not announce this ability.

Sadly munin's written [autoconf specification](http://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-concise.html#autoconf) is not very precise, thus I cannot really object your approach.
You decide ...

shtrom:
> This is similar to ssl-certificate-expiry, but creates single graph.

Doesn't `ssl-certificate-expiry` already create a single graph?

I get this:
Screen Shot 2020-12-16 at 12 03 03

kimheino:
You're correct, ssl-certificate-expiry can create single graph. I somehow missed that and only saw "ssl-certificate-expiry_www.example.net" etc lines in documentation. It can also do STARTTLS, mine doesn't.

My plugin was written in 2015 when ssl-certificate-expiry didn't even exist yet (ssl_ did).

I'll close this PR, ssl-certificate-expiry is better.

PR #1142, opened 2020-12-01
hue: plugin to monitor Philips Hue light status and temperature sensors

kimheino:


plugins/hue/hue
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/python3 -tt
```
sumpfralle, line 1: Please use `/usr/bin/env python3` without arguments, if possible.

plugins/hue/hue
```diff
@@ -0,0 +1,139 @@
...
+# -*- coding: utf-8 -*-
```
sumpfralle, line 2: python2-only, I guess.

plugins/hue/hue
```diff
@@ -0,0 +1,139 @@
...
+
+"""Munin plugin to read various data from Philips Hue.
```
sumpfralle, line 4: Please apply perlpod formatting.

plugins/hue/hue
```diff
@@ -0,0 +1,139 @@
...
+    if config:
+        print('lights.label Number of lights on')
+    if not config or both:
+        print('lights.value {}'.format(count))
+
+
+def print_data(config):
+    """Print config or values."""
+    # Get values
+    data = requests.get('http://{}/api/{}/'.format(
```
sumpfralle, line 123: How about using `json.loads(urllib.request.urlopen(URL).read())` instead? This would remove the dependency on `requests`.

plugins/hue/hue
```diff
@@ -0,0 +1,139 @@
...
+
+    # Print config/values
+    print_temperatures(data, config, both)
+    print_light_levels(data, config, both)
+    print_lights(data, config, both)
+
+
+if __name__ == '__main__':
+    if len(sys.argv) > 1 and sys.argv[1] == 'autoconf':
+        print('no')
```
sumpfralle, line 135: If there is only `no`, then you should better omit the `autoconf` handling.

plugins/hue/hue
```diff
@@ -0,0 +1,139 @@
...
+    uniqueid = sensor['uniqueid'][:27]
+    name = sensor['name']
+    for item in data['sensors'].values():
+        if item.get('uniqueid', '').startswith(uniqueid):
+            name = item['name']
+            break
+
+    # Convert ä to a and return it as utf8 string, py2/3 compatible way
+    name = unicodedata.normalize('NFKD', name)
+    return name.encode('ASCII', 'ignore').decode('utf-8')
```
sumpfralle, line 46: Is it necessary to reformat the name here? Munin should handle special characters well, I think.

sumpfralle (commented):
Thank you for this contribution!

Do you think it is safe to assume, that your plugin encompasses all the features of these light bulbs? Otherwise you may want to rename the plugin (being less generic) in order to allow later plugins to add more features. (e.g. statistics about usage periods or energy consumption ...)

plugins/hue/hue
kimheino: Munin 2.0.6something on RHEL7 does *not* support UTF-8. It might be because of old rrdtool instead of Munin, but the result looks bad. So I insist doing ä->a conversion here.

sumpfralle: That sounds reasonable.
But please add a comment referring to RHEL7 (improves understanding and allows retiring of this workaround, when RHEL7 goes end-of-life).
The issue should be fixed since Munin 2.0.11 (bb9466db1813ca9bc50e74e67ecbf49bcd011095).

kimheino:
> Do you think it is safe to assume, that your plugin encompasses all the features of these light bulbs?

My plugin includes all **useful** values from supported Hue devices. It doesn't include "is firmware up to date?" etc values. I'm willing to add more graphs if someone invents something useful.

> (e.g. statistics about usage periods or energy consumption ...)

Usage: This plugin includes count of lights currently on.

Energy: To calculate energy consumption you must manually configure each light bulbs average energy consumption, and then calculate an estimate based on that. I wonder if anyone really wants to do that?

Do you have any better name than "hue"? Maybe later energy-plugin could be called "hue_energy"?

sumpfralle:
> Do you have any better name than "hue"? Maybe later energy-plugin could be called "hue_energy"?

Thank you for explaining your reasoning. You convinced me, that your choice feels reasonable.

Please take a look at my other minor comments above. Then I will happily merge it.

kimheino:
Perlpod etc fixed.

sumpfralle:
Thank you!

PR #1145, opened 2020-12-01
nsd: new plugin to monitor nsd DNS server

kimheino:


plugins/nsd/nsd
```diff
@@ -0,0 +1,127 @@
...
+    echo 'snxd.type DERIVE'
+    echo 'snxd.min 0'
+    echo 'rq.label Total Successful'
+    echo 'rq.type DERIVE'
+    echo 'rq.min 0'
+    exit 0
+fi
+
+/usr/sbin/nsd-control stats_noreset | sed 's/=/ /; s/\.//g' | (
+    numtypeA=0
```
sumpfralle, line 86: Maybe `U` (failed to retrieve value) would be more suitable here?
Or should the missing key being interpreted as *no requests of this type*?

sumpfralle (commented):
This looks good!

Since the plugin seems to care only for requests, you may want to rename it to `nsd_requests` or similar?

PR #1146, opened 2020-12-01
nut: new plugin to monitor Network UPS Tools (NUT)

kimheino:


plugins/nut/nut
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/python3 -tt
```
sumpfralle, line 1: Please use `/usr/bin/env python3` without arguments.

plugins/nut/nut
```diff
@@ -0,0 +1,98 @@
...
+# -*- coding: utf-8 -*-
+
+"""Munin plugin to monitor Network UPS Tools (NUT)
```
sumpfralle, line 4: Please apply `perlpod`.

sumpfralle (commented):
It looks good!

Do you think, that this could supersede the already existing plugin `power/nut`?
In this case, you could remove the other one.
If not, then it would be nice, if you could rename the plugin. For now `munin-get` and the plugin gallery do not really support plugins with the same name properly ...

kimheino:
I somehow missed already existing plugins in "power" subdir and therefore wrote my own. I'll close this pull request as existing nutups2_ has almost same features.

See pull request #1199 instead to improve nutups2_.

PR #1149, opened 2020-12-01
systemd_status: yet another plugin to monitor systemd status

kimheino:
Similar to systemd_units, but more states and ignores scopes and user
services. They must be allowed to fail.

### DIFF
--- a/plugins/systemd/systemd_status
+++ b/plugins/systemd/systemd_status
@@ -1,0 +1,84 @@
+#!/usr/bin/python3 -tt
+# -*- coding: utf-8 -*-
+# pylint: disable=invalid-name
+# pylint: enable=invalid-name
+
+"""Munin plugin to monitor systemd service status.
+Copyright 2018, Kim B. Heino, b@bbbs.net, Foobar Oy
+License GPLv2+
+
+#%# capabilities=autoconf
+#%# family=auto
+"""
+
+import os
+import re
+import subprocess
+import sys
+
+
+STATES = (
+    'failed',
+    'dead',
+    'running',
+    'exited',
+    'active',
+    'listening',
+    'waiting',
+    'plugged',
+    'mounted',
+)
+
+
+def config():
+    """Autoconfig values."""
+    print('graph_title systemd services')
+    print('graph_vlabel Services')
+    print('graph_category processes')
+    print('graph_args --base 1000 --lower-limit 0')
+    print('graph_scale no')
+    print('graph_info Number of services in given activation state.')
+    for state in STATES:
+        print('{state}.label Services in {state} state'.format(state=state))
+    print('failed.warning 0:0')
+    if os.environ.get('MUNIN_CAP_DIRTYCONFIG') == '1':
+        fetch()
+
+
+def fetch():
+    """Print runtime values."""
+    # Get data
+    try:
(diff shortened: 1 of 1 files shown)


plugins/systemd/systemd_status
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/python3 -tt
```
sumpfralle, line 1: Why `-tt`?

Also please use `/usr/bin/env python` in order to allow virtualenv usage.

plugins/systemd/systemd_status
```diff
@@ -0,0 +1,84 @@
...
+# -*- coding: utf-8 -*-
```
sumpfralle, line 2: I think, this is a `python2` marker?

plugins/systemd/systemd_status
```diff
@@ -0,0 +1,84 @@
...
+# pylint: disable=invalid-name
+# pylint: enable=invalid-name
+
+"""Munin plugin to monitor systemd service status.
```
sumpfralle, line 6: Please reformat for `perlpod`.

plugins/systemd/systemd_status
```diff
@@ -0,0 +1,84 @@
...
+        output = subprocess.check_output([
+            '/bin/systemctl',
+            'list-units'])  # deb9/py3.5 doesn't have encoding parameter
+    except (OSError, subprocess.CalledProcessError):
+        return
+
+    # Parse data
+    states = {state: 0 for state in STATES}
+    for line in output.splitlines():
+        line = line.decode('utf-8').split()
```
sumpfralle, line 61: Just for my taste: I would avoid overriding an existing variable `line` with a different type of content (*line* changes to *list of line tokens*). Maybe use `tokens` instead?
Of course, this is just a matter of taste ...

sumpfralle (commented):
This looks good to me!

Could you please clarify the specific features of this plugin in the perlpod header? Users should be guided to the right plugin for their use-case ...

plugins/systemd/systemd_status
kimheino: python3's default is utf-8, so above "coding" marker just sets it again to utf-8 = does nothing. So it could be omitted. It (and -tt) is mostly leftover from quick py2->py3 conversion. 

Some older editors still require it as they default to py2 and it's charset. I can take it away as my $EDITOR doesn't need it.

niclan: The -*- tag is for the benefit of Emacs I believe.

plugins/systemd/systemd_status
```diff
@@ -0,0 +1,110 @@
...
+            states[token[3]] = states[token[3]] + 1
+
+    # Output
+    for state in STATES:
+        print('{}.value {}'.format(state, states[state]))
+
+
+if __name__ == '__main__':
+    if len(sys.argv) > 1 and sys.argv[1] == 'autoconf':
+        print('yes')
```
sumpfralle, line 106: One last thing (I did not notice it before): here you should check for `systemd` being the currently active init implementation (e.g. Debian checks for the existence of the directory `/run/systemd/system`).

sumpfralle:
Great - thank you!

PR #1151, opened 2020-12-01
varnish4_multigraph: multigraph version of varnish4_, rewritten in python

kimheino:


plugins/varnish/varnish4_multigraph
```diff
@@ -0,0 +1,453 @@
+#!/usr/bin/python3 -tt
```
sumpfralle, line 1: What is the effect of `-tt`? I could not find it in the man page (but `python3` accepts it - thus it seems to work).

btw.: I remember vaguely, that shebang parameters are not evaluated on every platform. Thus I would prefer to remove it. But maybe I am mistaken ...

plugins/varnish/varnish4_multigraph
```diff
@@ -0,0 +1,453 @@
...
+# -*- coding: utf-8 -*-
+
+""" Munin plugin to monitor Varnish 4 status.
```
sumpfralle, line 4: Please convert this to the `perlpod` format. Users of the plugin gallery and `munin-get` will appreciate this :)

plugins/varnish/varnish4_multigraph
```diff
@@ -0,0 +1,453 @@
...
+    output = run_binary(['/usr/bin/varnishstat', '-j'])
+    try:
+        return json.loads(output)
+    except (TypeError, ValueError):
+        return {}
+
+
+def autoconf():
+    """Print "yes" or "no"."""
+    status = 'yes' if get_values() else 'no'
```
sumpfralle, line 46: It would be great, if you could distinguish between *varnishstat is missing* and *execution of varnishstat failed*.

plugins/varnish/varnish4_multigraph
```diff
@@ -0,0 +1,453 @@
...
+    print('s_resp_bodybytes.min 0')
+    print('s_resp_bodybytes.draw AREA')
+    print('s_resp_bodybytes.cdef s_resp_bodybytes,8,*')
+    print('s_resp_bodybytes.type DERIVE')
+    print('s_resp_hdrbytes.label Header traffic')
+    print('s_resp_hdrbytes.min 0')
+    print('s_resp_hdrbytes.draw STACK')
+    print('s_resp_hdrbytes.cdef s_resp_hdrbytes,8,*')
+    print(
+        's_resp_hdrbytes.info HTTP Header traffic. TCP/IP overhead is not '
```
sumpfralle, line 292: Feel free to use a line limit of 99 characters. But shorter lines are of course also allowed :)

sumpfralle (commented):
Thank you for this well-written plugin!

Besides my comments on the code, I have two questions:
* A) Is this plugin specific to Varnish v4? Or would it run with v5 or v6, too? In this case, I would appreciate a different name.
* B) Somehow I would prefer a different name (more specific) without the `_multigraph` suffix, but I cannot really think of a better one. Maybe you have an idea? Otherwise just leave it, as it is ...

Thank you!

kimheino:
I don't know if this works on varnish 5/6 or not. It might, as 6's output is json. Maybe some new/missing fields, but it should be easy to fix.

This plugin was written in 2016 for some of my customers. They don't use varnish anymore, so unfortunately I don't have any real interest to test this with 6. And nobody should be using 4 anymore, actually.

What should we do? I could fix perlpod and shebang, but is there really any use for this? Maybe only as a template for varnish6-plugin? 

I'm fine if you just close and reject this PR. Varnish 4 is dead.

sumpfralle:
> Maybe only as a template for varnish6-plugin?

I am not a varnish user, thus I cannot really tell, whether your plugin is more suitable, than the ones already present. Your `json`-based filtering looks like a good start (I think, the others use xml or something else).

In case you think, that this plugin could be a good start for a varnish plugin superseding the others in functionality or robustness, then I would appreciate it being part of the repository. In this case, its name should not refer to v4.

kimheino:
What should it be called? varnish_json? varnish_py?

sumpfralle:
> What should it be called? varnish_json? varnish_py?

In case you consider it to be a more-or-less general plugin for varnish, then simply `varnish` could be a proper name. Or you may add any kind of suitable suffix, if you feel like.

kimheino:
It should be more-or-less general plugin, so renamed to varnish + perlpod etc fixed.

sumpfralle:
Thank you!

PR #1152, opened 2020-12-04
Plugin timesync_status: add support for 'fetch' command

droscy:
My installation of munin server (version 2.0.49) executes plugins with "fetch" command instead of an empty command line.

### DIFF
--- a/plugins/systemd/timesync_status
+++ b/plugins/systemd/timesync_status
@@ -132,9 +132,9 @@ def config():
 
 
 if __name__ == '__main__':
-    if len(sys.argv) == 1 or sys.argv[1] == 'fetch':
-        retrieve()
-    elif sys.argv[1] == 'config':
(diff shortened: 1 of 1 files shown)


sumpfralle:
I am not aware of Munin (specifically *munin-node*) emitting the argument *fetch* when calling plugins. But maybe I overlooked something ...

But alternative *munin-node* implementations may do this (e.g. [muninlite](https://github.com/munin-monitoring/muninlite/)). Thus it is good, if plugins handle this properly.

Thus I appreciate, that you want to fix this.
But I would suggest to use the pattern, that is applied by most other plugins:
```python
    if len(sys.argv) > 1 and sys.argv[1] == "config":
        config()
    else:
        retrieve()
```

Would you adjust your proposal along these lines, please?

droscy:
Sure, I'll use that pattern, even for "autoconf" argument.

sumpfralle:
Thank you!

PR #1155, opened 2020-12-12
fix the bird plugin to work with perl 5.28

gfa:
this is required to make the bird plugin to work with perl 5.28


sumpfralle:
Thank you!

PR #1157, opened 2020-12-15
[plugin/nextcould] Fix incorrect data output

shtrom:
I stupidly forgot to output the `.value` for the data fields in my previous refactor, which completely broke the plugin. Sorry about that...


sumpfralle:
hehe - things happen - thank you!

PR #1158, opened 2020-12-22
certificate_file_expiry: add pattern support for CERT env variable

ap-wtioit:
It became tedious to maintain our env.CERT values so i added support for patterns to the plugin. Now it's easier to support a bunch of certs that are saved in a certain pattern (e.g. letsencrypt certs or crls for multiple vpns)

Info @wt-io-it

### DIFF
--- a/plugins/ssl/certificate_file_expiry
+++ b/plugins/ssl/certificate_file_expiry
@@ -18,7 +18,7 @@ For letsencrypt certificates
 
  [certificate_file_expiry]
  user root
- env.CERTS x509:/etc/letsencrypt/live/domain1.example.com/cert.pem x509:/etc/letsencrypt/live/domain2.example.com/cert.pem
+ env.CERTS x509:/etc/letsencrypt/live/*/cert.pem
 
 Warning and Critical levels can also be configured with env variables like this:
 
@@ -28,6 +28,14 @@ Warning and Critical levels can also be configured with env variables like this:
  env.warning 5:
  # critical when certificate will be invalid within 1 day
  env.critical 1:
+
+env.CERTS should be a space separated list of patterns prefixed by the type of certificate to check and a colon. All types of
+certificates that openssl supports as standard commands and have a validity output are supported (e.g. x509, crl).
+File patterns can be a single file (e.g. /etc/openvpn/easy-rsa/keys/crl.pem) or a pattern that matches multiple files
+(e.g. /etc/letsencrypt/live/*/cert.pem).
+
+env.warning and env.critical are configurable values for the warning and critical levels according to
+http://munin-monitoring.org/wiki/fieldname.warning and http://munin-monitoring.org/wiki/fieldname.critical
 
 =head1 Dependencies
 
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for this suggestion!

Since the plugin is available for two years, I would like to keep it backward-compatible (supporting multiple items in `CERT`).
I think, this is rather easy, by adding a loop around the code block (the wildcards will not be substituted at that point, since the `x509:` prefix prevents the paths from expanding).
Additionally it would be great, if you could add a documentation note regarding the support for shell-based wildcards in the `CERT` fields.

What do you think?

ap-wtioit:
@sumpfralle i tested the backwards compatibility (locally and then on our servers, we have on server with a pattern and one without a pattern)

the second loop makes only one iteration if the pattern is not a pattern but a file name:

```
for home in /home/$USER; do echo $home; done
```
should give you your users home dir

```
for home in /home/*; do echo $home; done
```
should give you all home dirs on the system (you are allowed to list)

sumpfralle:
My question was probably a bit vague, sorry :)

I meant *multiple space-separated tokens*, e.g.:
```
CERTS=x509:/etc/letsencrypt/live/*/cert.pem x509:/var/lib/dehydrated/certs/*/cert.pem
```

I think, your proposal will not work with this (previously valid) input anymore. But maybe I just overlooked something?

ap-wtioit:
The input `CERTS=x509:/etc/letsencrypt/live/*/cert.pem x509:/var/lib/dehydrated/certs/*/cert.pem` meaning a directories named `*`?
At least for letsencrypt with certbot this would not be a valid directory.

If meaning pattern syntax so all files /etc/letsencrypt/live/*/cert.pem and /var/lib/dehydrated/certs/*/cert.pem this was not working before and is only working with this patch.

And this one is still working on our servers
```
env.CERTS x509:/etc/letsencrypt/live/domain1.tld/cert.pem x509:/etc/letsencrypt/live/domain2.tld/cert.pem x509:/etc/letsencrypt/live/domain3.tld/cert.pem x509:/etc/letsencrypt/live/domain4.tld/cert.pem
```

ap-wtioit:
The outer for loop (https://github.com/munin-monitoring/contrib/pull/1158/files#diff-a922d0a34322b760b165648a182fa41b6964e249f8b6f9fc9ee7e38cabde64caR58) splits the definitions:
* One iteration for x509:/etc/letsencrypt/live/*/cert.pem and one for x509:/var/lib/dehydrated/certs/*/cert.pem
* Then for each definition the type and pattern is determined
* The inner loop (https://github.com/munin-monitoring/contrib/pull/1158/files#diff-a922d0a34322b760b165648a182fa41b6964e249f8b6f9fc9ee7e38cabde64caR61) iterates over each patterns matching files:
** if the pattern is just a file (either because it was not defined as a pattern or because the pattern only matches one file) then the inner for loop just has one iteration
** and checks the validity according to the expected certificate type

Also if you replace `env.CERTS x509:/etc/letsencrypt/live/domain1.tld/cert.pem x509:/etc/letsencrypt/live/domain2.tld/cert.pem x509:/etc/letsencrypt/live/domain3.tld/cert.pem x509:/etc/letsencrypt/live/domain4.tld/cert.pem` by `env.CERTS x509:/etc/letsencrypt/live/*/cert.pem` if the files remain the same the old values in the rrd files are still valid because `cert_file` holds the evaluated pattern.

So besides one literally naming directories with a `*` in it and only wanting a single directory i cannot see what's not backwards compatible.

ap-wtioit:
The previous code did not split patterns:
* in https://github.com/munin-monitoring/contrib/blob/09b88141b34940d100df10b95b60206979e9df17/plugins/ssl/certificate_file_expiry#L58 there was still the `x509:` prefix before the pattern preventing the linux pattern matching for files and directories
* in https://github.com/munin-monitoring/contrib/blob/09b88141b34940d100df10b95b60206979e9df17/plugins/ssl/certificate_file_expiry#L59 and https://github.com/munin-monitoring/contrib/blob/09b88141b34940d100df10b95b60206979e9df17/plugins/ssl/certificate_file_expiry#L60 those string operations also do not evaluate patterns
* `clean_filename` also should not evaluate patterns in here https://github.com/munin-monitoring/contrib/blob/09b88141b34940d100df10b95b60206979e9df17/plugins/ssl/certificate_file_expiry#L61
* and also every other occurence of `cert_file` is put in quotes to prevent expansion there.

sumpfralle:
Sorry, for causing the confusion:
* I overlooked the outer loop (which you kept), thus multiple space-separated tokens are still working (my assumption was wrong)
* My example was meant to refer only to *space-separated tokens* (not *pattern evaluation*). Of course, your proposal introduces this new feature.

Thank you for correcting my wrong assumption.

I am happy to merge your proposal, after you add a sentence to the documentation header, which explains your shiny new feature.
Thanks!

sumpfralle:

PR #1159, opened 2021-01-02
openstack_swift_stats_: Add openstack_swift_stats_ plugin

cortices:
This plugin written in Bash script provides two graphs of the number of objects and size in bytes of an OpenStack Swift object store container. It uses the python-swiftclient command-line-tool (invoked as `swift`) which is provided by openstack.org and available on many distro package repositories.


cortices:
I'm not quite sure how to implement shellcheck's suggestion here.

plugins/openstack_swift/openstack_swift_stats_
```diff
@@ -0,0 +1,67 @@
...
+objects.draw AREA
+objects.min 0
+objgrowth.label Growth/\${graph_period}
+objgrowth.type DERIVE
+objgrowth.min 0
+EOM
+        exit 0;;
+esac
+
+IFS=" "
```
sumpfralle, line 41: Is the `IFS` assignment really necessary here? It could be confusing for future readers, since it is effective until the end of the script.
(does it provide any required effect right now?)

plugins/openstack_swift/openstack_swift_stats_
```diff
@@ -0,0 +1,67 @@
...
+STAT_OUTPUT=( $(swift stat "$CONTAINER" | awk 'NR>=3&&NR<=4 {print $2}' FS=": " ORS=" ") )
```
sumpfralle, line 42: I think, you can omit `ORS=" "` here, since the default `IFS` would also treat line breaks as a separator.

plugins/openstack_swift/openstack_swift_stats_
```diff
@@ -0,0 +1,67 @@
...
+printf "multigraph openstack_swift_stats_size\n"
+
+# Check value is an integer.
+if [ -z "${STAT_OUTPUT[1]}" ] || [ "${STAT_OUTPUT[1]}" != "${STAT_OUTPUT[1]}" ] 2>/dev/null; then
+    printf "size.value 0\n"
+    printf "Error: no usable size value from swift command.\n" 1>&2
+    exit 1
+fi
+
+printf "size.value %s\ngrowth.value %s\n" "${STAT_OUTPUT[1]}" "${STAT_OUTPUT[1]}"
```
sumpfralle, line 54: Is the `size` and the `growth` field always the same?

plugins/openstack_swift/openstack_swift_stats_.pod
```diff
@@ -0,0 +1,77 @@
+=cut
```
sumpfralle, line 1: Please move this nicely-formatted documentation into the header of the plugin. Otherwise it will not be available for `munin-get` or the [plugin gallery](https://gallery.munin-monitoring.org/).

plugins/openstack_swift/openstack_swift_stats_
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+ME="${0##*/}"
+CONTAINER=$(printf "%s" "$ME" | cut -d'_' -f4 -)
+# printf "$CONTAINER\n"
```
sumpfralle, line 5: Please removing debugging statements.

sumpfralle (commented):

Regarding your proposed directory (`openstack_swift`): I think, `openstack` would be more appropriate, as *openstack* is the environment and *swift* is just a tool for querying/managing it. Or what do you think?

btw: did you think about creating a multigraph plugin instead of a wildcard plugin? It could emit data for *all* containers, instead of being required to configure each single container manually. Both is fine, of course ...

sumpfralle (commented):
The suggestion from shellcheck points at [SC2207](https://github.com/koalaman/shellcheck/wiki/SC2207). I am rarely using `bash` (instead of `sh`), thus I cannot really recommend the proper approach.

In an `sh` world, I would use something like the following:
```sh
data=$(swift | sed -n '3,4s/^.*: //p'
```

Accessing a field would be something like this: `printf '%s' "$data" | sed -n 1p`

But this is just another taste :)

Feel free to override the shellcheck test by putting the line `# shellcheck disable=SC2207` just before the line containing the array assignment. Or maybe the [documentation of SC2207](https://github.com/koalaman/shellcheck/wiki/SC2207) refers to useful alternatives.

cortices:
I will look at making recommended improvements soon, for now just addressing your initial remarks:

> Regarding your proposed directory (`openstack_swift`): I think, `openstack` would be more appropriate, as _openstack_ is the environment and _swift_ is just a tool for querying/managing it. Or what do you think?

OpenStack is a broader framework supporting a range of services including object storage, block storage, compute units etc. I used the suffix as Swift refers specifically to the object storage management protocol, leaving room for e.g. `openstack_compute_` plugins.
I think you are right that the directory should be `openstack` though.

> btw: did you think about creating a multigraph plugin instead of a wildcard plugin? It could emit data for _all_ containers, instead of being required to configure each single container manually. Both is fine, of course ...

This was also an intentional choice as user permissions with openstack are not necessarily very granular. An account may have access to more containers that aren’t relevant to that particular munin node. For example, dev & prod containers should only display on the relevant node. I will look into implementing logic to allow `_*` to display all containers in a future version.

sumpfralle:
> This was also an intentional choice as user permissions with openstack are not necessarily very granular. An account may have access to more containers that aren’t relevant to that particular munin node. For example, dev & prod containers should only display on the relevant node.

Thank you for the explanation. Maybe an exclude list (or a pattern) would would be useful in this case.

> I will look into implementing logic to allow _* to display all containers in a future version.

Specifically the asterisk is probably not a good choice, since this special character is not allowed in plugin names. Maybe defining a magic name like *all* would be a good alternative.
On the other hand, mixing wildcard and multigraph logic in a single plugin may be hard to read (and to explain to users). Anyway: the decision is up to you.

Thank you!

sumpfralle:
@cortices: what do you think about your proposal? Do you want to work on it or should I merge it as it is right now? I would not like to see this useful contribution rot away ...

cortices:
Thanks for checking in. I think it is definitely worthwhile for me to tidy up and make the script more robust first. I’ve just had a lot to work on. Give me another month and I hope to have it refactored nicely.

sumpfralle:
> Give me another month and I hope to have it refactored nicely.

That sounds great!

sumpfralle:
Ping?

cortices:
I'm still alive but haven't yet gotten around to the refactoring.

cortices:
Thanks for the reminder earlier. Finally had just the right moment to sit down and properly fix this up. Hopefully this addresses all your concerns. It runs in posix sh and passes shellcheck with zero warnings now :)

sumpfralle:
Yeah - thanks!

Just one more thing, in case you are still in the mood: please replace the many single-line comments (`#`) in the header with one multi-line comment:
```shell
: <<=cut
...
=cut
```

This will create a beautiful perlpod comment header. The plugin's future page on https://gallery.munin-monitoring.org would appreciate this :)

(if you are not in the mood anymore, then I will merge your plugin as-is)

cortices:
No problem. I just didn’t know how to do that in a shell script.

sumpfralle:
Great - thank you!

PR #1160, opened 2021-01-07
Update for extra fields in newer kernels

smith153:
The fields have changed from 11 or 15 on recent kernels but does not effect what the plugin gathers. 
See 5.4 https://www.kernel.org/doc/html/v5.4/block/stat.html versus 4.16 https://mjmwired.net/kernel/Documentation/block/stat.txt


sumpfralle:
Thanks!

PR #1161, opened 2021-01-12
added bigbluebutton plugin

zettrik:
This is a new plugin to monitor the usage (user, meeting rooms ...) of a bigbluebutton server.

### DIFF
--- a/plugins/bigbluebutton/bbb_usage
+++ b/plugins/bigbluebutton/bbb_usage
@@ -1,0 +1,168 @@
+#!/usr/bin/env python3
+"""
+
+=head1 NAME
+
+bbb - monitor Bigbluebutton server (users, rooms, videostreams ...)
+
+=head1 APPLICABLE SYSTEMS
+
+Bigbluebutton server
+
+=head1 CONFIGURATION
+
+No configuration is needed. This script will fetch the secret from the local
+bbb instance
+
+https://docs.bigbluebutton.org/dev/api.html
+
+=head1 AUTHOR
+
+Copyright 2021 Henning Rieger <age@systemausfall.org>
+
+With many thanks to the original Author Bernd Wurst (bwurst.org)
+
+=head1 LICENSE
+GNU General Public License v3.0 or later
+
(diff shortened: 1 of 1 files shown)


plugins/bigbluebutton/bbb_usage
```diff
@@ -0,0 +1,174 @@
...
+
+    requesturl = APIURL + apimethod + '?' + querystring + 'checksum=' + checksum
+
+    # make api request
+    response = urllib.request.urlopen(requesturl)
+    responsedata = response.read()
+    try:
+        tree = ElementTree.fromstring(responsedata)
+    except HTTPError as e:
+        print('There was an error with the recieved data from bbb api.')
```
sumpfralle, line 115: Here you could add `file=sys.stderr`. Otherwise munin would try to parse the mess.

plugins/bigbluebutton/bbb_usage
```diff
@@ -0,0 +1,174 @@
...
+import sys
+import subprocess
+import re
+import hashlib
+import urllib.request
+from urllib.error import URLError, HTTPError
+import os
+from xml.etree import ElementTree
+
+bbbconf_path = "/usr/bin/bbb-conf"
```
sumpfralle, line 53: If you want to provide a fancy interface, you could take this path from the environment (e.g. `bbbconf_path = os.getenv("BBBCONF_BIN", "/usr/bin/bbb-conf")`) and describe this in the *Configuration* section.

sumpfralle (commented):
This looks good to me, thanks!

Please take a look at my two nitpicks below. But you may feel free to ignore them - the plugin is fine, as it is now ...

sumpfralle:
What do you think?

Neustradamus:
@zettrik: Can you look @sumpfralle comments?

What do you think?

plugins/bigbluebutton/bbb_usage
zettrik: This looks also much better. But as I don't have the BBB environment anymore I can't commit a tested example.

PR #1162, opened 2021-01-16
mpdstats_ improvements

letompouce:
* Introduces a new environment variable to pass arguments to the `nc(1)` binary
* Uses `stats` instead of `version` in the `autoconf` call

cc @sumpfralle - thanks for #759 !

### DIFF
--- a/plugins/mpd/mpdstats_
+++ b/plugins/mpd/mpdstats_
@@ -89,9 +89,10 @@ ACTION="$(basename "$0" | sed 's/^.*_//')"
 #
 
 do_autoconf () {
+    # shellcheck disable=SC2086
     if [ -z "$NCBIN" ] ; then
         echo "no (missing netcat program ('nc'))"
-    elif ! echo stats | "$NCBIN" "$NCARGS" "$MPDHOST" "$MPDPORT" >/dev/null 2>&1; then
+    elif ! echo stats | "$NCBIN" $NCARGS "$MPDHOST" "$MPDPORT" >/dev/null 2>&1; then
         echo "no (connection failed)"
     else
         echo "yes"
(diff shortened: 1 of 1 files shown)


plugins/mpd/mpdstats_
```diff
@@ -83,7 +91,7 @@ ACTION="$(basename "$0" | sed 's/^.*_//')"
 do_autoconf () {
     if [ -z "$NCBIN" ] ; then
         echo "no (missing netcat program ('nc'))"
-    elif ! echo version | "$NCBIN" "$MPDHOST" "$MPDPORT" >/dev/null 2>&1; then
+    elif ! echo stats | "$NCBIN" "$NCARGS" "$MPDHOST" "$MPDPORT" >/dev/null 2>&1; then
```
sumpfralle, line 94: I guess, you want to access *multiple* arguments for netcat here (or even none at all), thus you may not quote these in the call (they would be a single token and thus rejected by netcat).
In order to work around the resulting shellcheck complain, an override (something like `# shellcheck disable=SC....`) in the line above (or at the beginning of the `if` conditional) would be necessary.

sumpfralle (commented):

I added a comment below regarding one implementation detail. The rest looks fine to me.

plugins/mpd/mpdstats_
letompouce: Well, actually it works like this with `env.ncargs -w 1` but I unquoted those anyway!

sumpfralle: Indeed - `nc` seems to handle parameter pairs (here `-w 1`) somehow special. With `-4 -w 1` it breaks (as expected).
Thanks for fixing it!

PR #1164, opened 2021-01-19
Docker improvements

shtrom:
* Work around missing `total_rss` in some case (maybe cgroupv2?)
* Show and warn on unhealthy containers separately to those running
* Add a multigraph mode, to reap all the goodness of the wrapper in reducing API calls

### DIFF
--- a/plugins/docker/docker_
+++ b/plugins/docker/docker_
@@ -118,6 +118,10 @@ class ClientWrapper:
         if exclude_re:
             self.exclude = re.compile(exclude_re)
 
+    @property
+    def api(self):
+        return self.client.api
+
     @cached_property
     @sorted_by_creation_date
     def containers(self):
--- a/plugins/docker/docker_
+++ b/plugins/docker/docker_
@@ -169,18 +169,18 @@ class ClientWrapper:
         return self.client.volumes.list()
 
 
-def container_summary(container):
+def container_summary(container, *args):
     summary = container.name
-    attributes = container_attributes(container)
+    attributes = container_attributes(container, *args)
     if attributes:
         summary += f' ({attributes})'
     return summary
 
 
-def container_attributes(container):
+def container_attributes(container, *args):
     attributes = container.image.tags
     attributes.append(container.attrs['Created'])
-    return ', '.join(attributes)
+    return ', '.join(attributes + list(args))
 
 
 def print_containers_status(client):
--- a/plugins/docker/docker_
+++ b/plugins/docker/docker_
@@ -4,8 +4,9 @@
 
 docker_ - Docker wildcard-plugin to monitor a L<Docker|https://www.docker.com> host.
 
-This wildcard plugin provides at the moment only the suffixes C<containers>, C<images>, C<status>,
-C<volumes>, C<cpu>, C<memory> and C<network>.
+This wildcard plugin provides series C<containers>, C<images>, C<status>,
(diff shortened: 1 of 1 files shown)


plugins/docker/docker_
```diff
@@ -193,7 +205,11 @@ def print_containers_status(client):
     dead = []
     for container in client.all_containers:
         if container.status == 'running':
-            running.append(container)
+            state = client.api.inspect_container(container.name)['State']
+            if 'Health' in state and state['Health']['Status'] == 'unhealthy':
```
sumpfralle, line 209: You could simplify this (and add a bit of robustness) with `if state.get('Health', {}).get('Status') == 'unhealthy':` - but this is just a matter of taste ...

plugins/docker/docker_
```diff
@@ -319,155 +509,22 @@ def main():
...
-            print_containers_network(client)
+    if wildcard in series:
+        # dereference the function name by looking in the globals()
+        globals()[wildcard](client, mode)
+    elif wildcard == 'multi':
+        for s in series:
+            print(f'multigraph docker_{s}')
+            globals()[s](client, mode)
+    else:
+        print(f'unknown series ({wildcard})')
```
sumpfralle, line 526: You should redirect error messages via `file=sys.stderr`.

plugins/docker/docker_
```diff
@@ -319,155 +509,22 @@ def main():
...
-                print("{}_up.info {}".format(container.name, container_attributes(container)))
-        else:
-            print_containers_network(client)
+    if wildcard in series:
+        # dereference the function name by looking in the globals()
+        globals()[wildcard](client, mode)
+    elif wildcard == 'multi':
+        for s in series:
+            print(f'multigraph docker_{s}')
+            globals()[s](client, mode)
```
sumpfralle, line 524: I think, specifying a dict (mapping the requested features to their corresponding functions) would be a bit less hacky :)

plugins/docker/docker_
```diff
@@ -319,155 +509,22 @@ def main():
         sys.exit(1)
 
     if mode == "suggest":
-        print("cpu")
-        print("images")
-        print("memory")
-        print("network")
-        print("status")
-        print("volumes")
+        print("multi")
```
sumpfralle, line 512: I like your approach to change the default to *multi* (while keeping the single graphs working). But please add a comment here, why only one of the available wildcard features is announced.
Your future readers (including yourself) will surely appreciate this :)

sumpfralle (commented):
This looks like a good amount of fine changes, thanks!

plugins/docker/docker_
shtrom: Heh, I was hoping to avoid this, the dict is completely redundant code.

sumpfralle:
Thanks!

The hacky usage of `globals()` is really not my taste, but of course, it works :)

From my point of view, these nice improvements are ready to be merged. Or what do you think?
Please rebase your fixup commits - then I will merge them. Thanks!

shtrom:
Heh, fair enough. I added a long comment around the `globals()` use (but kept it q:)

I just rebased and squashed, so it should be good to merge.

sumpfralle:
Yes, that looks like it can be understood by future generations :)
Thank you!

PS: personally I prefer to use the obvious code pattern (a dict) instead of writing a longer comment for a hacky workaround - but this is obviously a matter of taste :)

PR #1165, opened 2021-01-19
initial commit of systemd_mem (v1.0)

paulalexandrow:
The idea is, that systemctl calculates memory consumption quite differently than existing plugins, which usually query /proc/... Hence I wrote this plugin, to monitor what memory a service uses according to systemctl.


plugins/systemd/systemd_mem
```diff
@@ -0,0 +1,121 @@
...
+}
+
+output_usage() {
+    printf >&2 "%s - munin plugin to graph memory usage for systemd services\n" "${0##*/}"
+    printf >&2 "Usage: %s [config]\n" "${0##*/}"
+}
+
+output_autoconf() {
+    missing=0
+    command -v systemctl >/dev/null 2>&1 || missing=1
```
sumpfralle, line 91: You could simplify this, if you like:
```shell
if command -v systemctl >/dev/null 2>&1; then
```

plugins/systemd/systemd_mem
```diff
@@ -0,0 +1,121 @@
...
+        printf "%s.info memory usage\n" "$clean_name"
+    done
+}
+
+output_values() {
+    for service in $(get_services); do
+        clean_name="$(clean_fieldname "$service")"
+        usage=$(systemctl show "$service" --all | grep ^MemoryCurrent= | cut -d '=' -f 2)
+        if [ "$usage" = "[not set]" ]; then
+            usage=0
```
sumpfralle, line 78: Maybe you could use `U` here. Munin would interpret this as *unknown* (e.g. due to an unexpected condition).

plugins/systemd/systemd_mem
```diff
@@ -0,0 +1,121 @@
...
+#%# family=auto
+#%# capabilities=autoconf
+
+=cut
+
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
+
+get_services() {
+    services=${services:-"munin-node"}
+    echo "$services" | awk 'BEGIN{RS=" "}{$1=$1}1'
```
sumpfralle, line 56: What is the purpose of the `awk` filtering here?
Maybe `service=${services:-munin-node}` would be sufficient?

This function feels a bit weird, since it changes its input (as a side-effect due to the non-local variable `services`). Thus multiple invocations of the function *could* (in theory) change the result.

plugins/systemd/systemd_mem
```diff
@@ -0,0 +1,121 @@
...
+        echo "no (systemctl not found)";
+    fi
+}
+
+case $# in
+    0)
+        output_values
+        ;;
+    1)
+        case $1 in
```
sumpfralle, line 104: You should also accept `fetch` as a single argument (for `output_values`).

sumpfralle (commented):

The plugin looks fine to me. Please take a look at my comments below.

plugins/systemd/systemd_mem
paulalexandrow: Afaik systemctl returns `[not set]` only in case a service is just not running. So from a logical point of view, it is not necessarily unexpected and also it _is_ known how much memory is used - zero.

And from a practical point of view, I use a lower alerting limit of 1 on some services to get notified when they stopped running. I know that there are better solutions for that, but it's still a convenient side effect.

plugins/systemd/systemd_mem
paulalexandrow: I am in no way a bourne-pro, so any help is appreciated. What I am trying to achieve is, to get the space separated plugin setting `apache php munin ...` to be newline separated in order to use it later in one of the for-loops. Because there are two of them, I wanted to put the space-to-newline conversion in a function. Could you kindly propose what would be a better way to achieve that?

sumpfralle: You can loop any separator in a `for` loop (see `IFS`: *<space>, <tab> and <newline>*). Thus the conversion is probably not necessary at all?

plugins/systemd/systemd_mem
sumpfralle: Yes, I agree - zero seems to be a reasonable value in this case.

plugins/systemd/systemd_mem
paulalexandrow: Thanks for the hint - i was not aware of `IFS` (and the fact, that `space` is already a default). You were right that the whole part in question could be removed.

sumpfralle:
Thank you!

PR #1171, opened 2021-02-01
[plugins/whois] Add support for multi-domain check

ruliane:
This PR adds support for checking several domains in one plugin, through the DOMAIN environment variable. Plugin is renamed from whois_ to whois, to reflect this change.
It also changes the shellcheck when importing plugin.sh (formerly SC1090, now SC1091).


plugins/network/whois
```diff
@@ -6,35 +6,37 @@
 
 =head1 NAME
 
-whois_ - Plugin to monitor expiry dates of domain in the WHOIS database
+whois - Plugin to monitor expiry dates of domain in the WHOIS database
```
sumpfralle, line 9: Maybe *of domains* (instead of *of domain*)?

plugins/network/whois
```diff
@@ -78,43 +80,48 @@ EOF
...
 	local FIELDNAME
 	local CACHEFILE
 
-	FIELDNAME="$(clean_fieldname "${NAME}")"
+	for NAME in $DOMAINS
+	do
+		FIELDNAME="$(clean_fieldname "${NAME}")"
 
-	CACHEFILE="${MUNIN_PLUGSTATE}/$(basename "${0}").${FIELDNAME}.cache"
+		CACHEFILE="${MUNIN_PLUGSTATE}/$(basename "${0}").${FIELDNAME}.cache"
```
sumpfralle, line 108: I did not look at the new file, but the github visualization indicates, that you are mixing tabs and spaces here (and in some more lines below)?

sumpfralle (commented):
That looks like a good improvement of the plugin!

Please take a look at my minor comments below. Thanks!

plugins/network/whois
ruliane: Right, now fixed.

plugins/network/whois
ruliane: Nope, only tabs are used. I included existing code in a `for` loop and GitHub highlights tabs added for indentation.

sumpfralle:
Thank you!

PR #1172, opened 2021-02-05
Add multigraph plugin for http response codes and times

klaus-tux:
In contrast to using http_responsecode and http_loadtime with the same
configuration, this plugin performs only one request per site and munin
run to gather its statistics.


gnuheidix (approved):
It works great and looks well crafted! :-) I already got rid of the other http_* plugins in my setups.

sumpfralle:
Indeed, the plugin looks fine - thank you for your contribution!

klaus-tux:
> Indeed, the plugin looks fine - thank you for your contribution!

Thank you for merging. I'm pleased that you like it :-)

sumpfralle:
In fact it is one of the very few *new* plugins, being just fine as-is-it-is without the need for (even nitpicking-level) improvements.
Congratulations! :)

klaus-tux:
Thank you :-) I fully understand what you mean. I see it as a result of "standing on the shoulders of giants" in combination with applied software craftsmanship.

PR #1173, opened 2021-02-05
cpu and memory plugins for OSX

sdarwin:
Hi,

Out of the box, the munin macports package for mac osx doesn't include graphs for CPU or MEMORY, which are possibly the most useful plugins. It would be great to provide something right away after a new installation.

You may have a better idea about file organization. An idea would be to include these in the contrib repository with the file names "cpu_osx" and "memory_osx", and in the macports distribution with the plain file names "cpu" and "memory".

cpu
memory

### DIFF
--- a/plugins/cpu/cpu_osx
+++ b/plugins/cpu/cpu_osx
@@ -12,7 +12,7 @@ cpu - Plugin to measure cpu on osx.
 =head1 NOTES
 
 This plugin runs the top command once per interval, to discover the current value of CPU usage on OSX.
-The result is scaled to # of CPU's, so a 4 core machine will reach 400% utilization (unless $scaleto100 is set to "yes").
+The result is scaled to # of CPU's, so a 4 core machine will reach 400% utilization (unless $scaleto100 is set to "yes", in which case the maximum will be 100%).
 
 Contributions are welcome to convert the plugin to use a long running counter such as /proc/stat in Linux.
 
@@ -32,7 +32,7 @@ if [ "$1" = "autoconf" ]; then
         echo yes
         exit 0
     else
-        echo no
+        echo "no (uname does not report 'Darwin')"
         exit 0
     fi
 fi
--- a/plugins/cpu/cpu_osx
+++ b/plugins/cpu/cpu_osx
@@ -1,0 +1,75 @@
+#!/bin/sh
+# -*- sh -*-
+
+# shellcheck disable=SC2046
+
+: << =cut
+
+=head1 NAME
+
+cpu - Plugin to measure cpu on osx.
+
+=head1 NOTES
+
+This plugin runs the top command once per interval, to discover the current value of CPU usage on OSX.
+The result is scaled to # of CPU's, so a 4 core machine will reach 400% utilization.
+Contributions are welcome to convert the plugin to use a long running counter such as /proc/stat in Linux.
+
+=head1 LICENSE
+
+GPLv2
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+
+NCPU=$(sysctl hw.ncpu | cut -d" " -f2)
+
+if [ "$1" = "autoconf" ]; then
+        echo yes
+        exit 0
+fi
+
+scaleto100=""
+
+if [ "$1" = "config" ]; then
+        # Linux had NCPU=$(egrep '^cpu[0-9]+ ' /proc/stat | wc -l)
+        if [ "$scaleto100" = "yes" ]; then
+                graphlimit=100
+        else
+                graphlimit=$(echo "$NCPU"*100 | bc)
+        fi
+        echo 'graph_title CPU usage'
+        echo "graph_order system user idle"
+        echo "graph_args --base 1000 -r --lower-limit 0 --upper-limit $graphlimit"
+        echo 'graph_scale no'
+        echo 'graph_vlabel %'
+        echo 'graph_category system'
+        echo 'system.label system'
+        echo 'system.draw AREA'
+        echo 'system.min 0'
+        echo "system.info CPU time spent by the kernel in system activities"
+        echo 'user.label user'
+        echo 'user.draw STACK'
+        echo 'user.min 0'
+        echo 'user.info CPU time spent by normal programs and daemons'
+        echo 'idle.label idle'
+        echo 'idle.draw STACK'
+        echo 'idle.min 0'
+        echo 'idle.info Idle CPU time'
+
+        exit 0
+fi
+
+# The second cpu reading is more accurate than the first, so "-l 2":
+TOPINFO=$(top -l 2 | grep "CPU usage: " | tail -n 1)
+
+CPU_USER=$(echo "$TOPINFO" | awk '/CPU usage: / { print substr($3, 1, length($3)-1) }')
+CPU_SYS=$(echo "$TOPINFO" | awk '/CPU usage: / { print substr($5, 1, length($5)-1) }') 
+CPU_IDLE=$(echo "$TOPINFO" | awk '/CPU usage: / { print substr($7, 1, length($7)-1) }') 
+echo "system.value" $( echo "$NCPU"*"$CPU_SYS"/1 | bc)
+echo "user.value" $( echo "$NCPU"*"$CPU_USER"/1 | bc)
+echo "idle.value" $( echo "$NCPU"*"$CPU_IDLE"/1 | bc)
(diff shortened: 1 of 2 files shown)


plugins/cpu/cpu_osx
```diff
@@ -0,0 +1,75 @@
...
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+
+NCPU=$(sysctl hw.ncpu | cut -d" " -f2)
+
+if [ "$1" = "autoconf" ]; then
```
sumpfralle, line 31: The `autoconf` capability is optional. Since it does nothing here, you should remove it.

plugins/cpu/cpu_osx
```diff
@@ -0,0 +1,75 @@
...
+        echo yes
+        exit 0
+fi
+
+scaleto100=""
```
sumpfralle, line 36: If you think, this setting may be useful, then you should make it configurable (e.g. `scaleto100=${scaleto100:-}`) and mention it in the documentation header.

plugins/cpu/cpu_osx
```diff
@@ -0,0 +1,75 @@
...
+
+if [ "$1" = "config" ]; then
+        # Linux had NCPU=$(egrep '^cpu[0-9]+ ' /proc/stat | wc -l)
+        if [ "$scaleto100" = "yes" ]; then
+                graphlimit=100
+        else
+                graphlimit=$(echo "$NCPU"*100 | bc)
```
sumpfralle, line 43: Please avoid using `bc`, since this would be an external dependencies (which you would need to check via `autoconf`). Instead you could just use `$((NCPU * 100))` (for integers) or `$(echo "$NCPU" | awk '{print($1 * 100)}')` for floating point numbers.

plugins/cpu/cpu_osx
```diff
@@ -0,0 +1,75 @@
...
+        exit 0
+fi
+
+# The second cpu reading is more accurate than the first, so "-l 2":
+TOPINFO=$(top -l 2 | grep "CPU usage: " | tail -n 1)
+
+CPU_USER=$(echo "$TOPINFO" | awk '/CPU usage: / { print substr($3, 1, length($3)-1) }')
+CPU_SYS=$(echo "$TOPINFO" | awk '/CPU usage: / { print substr($5, 1, length($5)-1) }') 
+CPU_IDLE=$(echo "$TOPINFO" | awk '/CPU usage: / { print substr($7, 1, length($7)-1) }') 
+echo "system.value" $( echo "$NCPU"*"$CPU_SYS"/1 | bc)
```
sumpfralle, line 73: Please avoid `bc`.

plugins/memory/memory_osx
```diff
@@ -0,0 +1,74 @@
...
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+
+TOTALMEM=$(sysctl hw.memsize | cut -d" " -f2)
+graphlimit=$(echo "$TOTALMEM"/1048576 | bc)
```
sumpfralle, line 29: Without `bc`, if possible.

plugins/memory/memory_osx
```diff
@@ -0,0 +1,74 @@
...
+
+dehumanise() {
+    echo "$1" | sed -e "s/K/*1024/g;s/M/*1024*1024/;s/G/*1024*1024*1024/;s/T/*1024*1024*1024*1024/" | bc
+}
+dehumanise_to_M() {
+    echo "$1" | sed -e "s/K/\/1024/g;s/M//;s/G/*1024/;s/T/*1024*1024/" | bc
+}
+
+if [ "$1" = "autoconf" ]; then
```
sumpfralle, line 38: You can omit `autoconf` here.

plugins/memory/memory_osx
```diff
@@ -0,0 +1,74 @@
...
+        echo 'graph_category system'
+        echo 'used.label used (not incl. wired)'
+        echo 'used.draw AREA'
+        echo 'used.min 0'
+        echo "used.info Used memory, not including wired, in MB"
+        echo 'wired.label wired'
+        echo 'wired.draw STACK'
+        echo 'wired.min 0'
+        echo 'wired.info Wired memory, in MB'
+        # echo 'unused.label unused'
```
sumpfralle, line 58: Please remove comments.

plugins/memory/memory_osx
```diff
@@ -0,0 +1,74 @@
...
+        exit 0
+fi
+
+# Note: while it's possible to show all three values "used", "wired", "unused", the sum total often differs by around 1% from an exact 100% value.
+# This makes the top of the chart uneven, and can be solved by omitting "unused" which will be implied.
+TOPINFO=$(top -l 1 | grep "PhysMem: ")
+MEM_USED=$(echo "$TOPINFO" | awk '/PhysMem: / { print substr($2, 1, length($2)) }')
+MEM_WIRED=$(echo "$TOPINFO" | awk '/PhysMem: / { print substr($4, 2, length($4)-1) }') 
+# MEM_UNUSED=$(echo "$TOPINFO" | awk '/PhysMem: / { print substr($6, 1, length($6)) }') 
+echo "used.value" $(echo $(dehumanise_to_M "$MEM_USED") - $(dehumanise_to_M "$MEM_WIRED") | bc)
```
sumpfralle, line 72: No comments and no `bc`, if possible.

sumpfralle (commented):

I think, these plugins could go into the main munin repository (see https://github.com/munin-monitoring/munin/tree/stable-2.0/plugins/node.d.darwin). Thus maybe you want to create a new pull request against the `stable-2.0` branch of that repository?

Please also take a look at my comments below. Thank you!

steveschnepp:
@sumpfralle, let's have a working plugin here, in `contrib.git` first. And the folks from macports can pick it if they want. I'd life to avoid adding things in `munin.git` for now, if we don't have a core maintainer for it.

sumpfralle:
> I'd like to avoid adding things in munin.git for now, if we don't have a core maintainer for it.

Yes, the lack of a developer for that platform is a pity.

So let us keep it here in *contrib*.

sdarwin:
@sumpfralle @steveschnepp , thanks for the feedback.  I will submit modifications soon.

plugins/cpu/cpu_osx
```diff
@@ -0,0 +1,80 @@
...
+    if [ "$(uname)" = "Darwin" ]; then
+        echo yes
+        exit 0
+    else
+        echo no
+        exit 0
+    fi
+fi
+
+scaleto100=${scaleto100:-}
```
sumpfralle, line 40: Please mention this configuration in the perldoc header.

plugins/cpu/cpu_osx
```diff
@@ -0,0 +1,80 @@
...
+ #%# capabilities=autoconf
+
+=cut
+
+if [ "$1" = "autoconf" ]; then
+    if [ "$(uname)" = "Darwin" ]; then
+        echo yes
+        exit 0
+    else
+        echo no
```
sumpfralle, line 35: You can help users by adding a reason is brackets (e.g. `no (uname does not report 'Darwin')`).

plugins/memory/memory_osx
```diff
@@ -0,0 +1,68 @@
...
+ #%# capabilities=autoconf
+
+=cut
+
+if [ "$1" = "autoconf" ]; then
+    if [ "$(uname)" = "Darwin" ]; then
+        echo yes
+        exit 0
+    else
+        echo no
```
sumpfralle, line 33: Same here: please add a reason.

plugins/memory/memory_osx
```diff
@@ -0,0 +1,68 @@
...
+        exit 0
+    fi
+fi
+
+TOTALMEM=$(sysctl hw.memsize | cut -d" " -f2)
+graphlimit=$(( TOTALMEM/1048576 ))
+
+dehumanise_to_M() {
```
sumpfralle, line 41: Please use base SI units (here: bytes) and enable `graph_scale`.
Users will appreciate this, as soon as 256 TB RAM start to be a commodity :)

sumpfralle (commented):
Thank you for your improvements. I added two more nitpicking comments.
Feel free to squash your commits, if you like.

sdarwin:
@sumpfralle those are good suggestions. I might not be able to get to it this week. hopefully soon.

sdarwin:
> scaleto100=${scaleto100:-}

The variable is mentioned in the notes.  I think it's sufficient, if you believe more details need to be added, maybe post in the comments what the whole notes section would look like.

> squash commits

you can squash when merging right?   just keeping it separate at this moment so the new commit is visible.

sumpfralle:
It is fine, as it is right now.

PR #1175, opened 2021-02-08
Use exact values in zfs/zpool_capacity#allocated

TimWolla:
As per the zpool(8) man page:

> zpool list [-HgLpPv] [-o property[,property]...] [-T u|d] [pool]... [interval [count]]
>         Lists the given pools along with a health status and space usage.  If no pools are specified, all pools in the system are listed.  When given an interval, the information is printed every
>         interval seconds until ^C is pressed.  If count is specified, the command exits after count reports are printed.
> [...]
>         -p      Display numbers in parsable (exact) values.


sumpfralle:
That looks better now - thanks!

Issue #1180, opened 2021-02-21
docker_ plugins say 'containers' serie is not supported but it seems it is


rsanzante:
docker_ plugin comments say:

```
This wildcard plugin provides series C<containers>, C<images>, C<status>,
C<volumes>, C<cpu>, C<memory> and C<network> as separate graphs. It also
supports a C<multi> suffix that provides all of those as a multigraph.
```

However, if you create a plugin symlink to the docker_ file called docker_containers when munin tries to execute it the response is:

```
munin-run docker_containers
# Unknown service 'docker_containers'
```

Checking the code it seems the 'containers' series is missing when checking for valid series:
```
    series = [
        'cpu',
        'images',
        'memory',
        'network',
        'status',
        'volumes',
    ]
```

Adding 'containers' to that variable result in a valid plugin execution:
```
munin-run docker_containers
containers_quantity.value 1
```

rsanzante:
PR added: https://github.com/munin-monitoring/contrib/pull/1181

sumpfralle:
Thanks for providing the fix for this!

PR #1181, opened 2021-02-21
Add 'containers' to the supported series

rsanzante:
Apparently, the code support that serie, as the docs say.

This addresses https://github.com/munin-monitoring/contrib/issues/1180


sumpfralle:
Thanks for fixing this issue!

PR #1183, opened 2021-03-04
Update munin-libvirtpy

AloisKlingler:
if config has sys.exit(1) the plugin config does not work.


sumpfralle:
Thank you for fixing this problem!

Since you are already looking at the code: maybe you would like to port the plugin to python3 and maybe redirect the `print` just before `sys.exit(1)` to `stderr`?
The fewer python2 plugins, the merrier :)

AloisKlingler:
> Thank you for fixing this problem!
> 
> Since you are already looking at the code: maybe you would like to port the plugin to python3 and maybe redirect the `print` just before `sys.exit(1)` to `stderr`?
> The fewer python2 plugins, the merrier :)

sorry, I am more a "look what is not working and try to fix it" person. :-)

PR #1184, opened 2021-03-04
Update kvm_cpu

AloisKlingler:
in python3 filter returns an object, not an list. I think a list is required here? at least the plugin works as soon as list(...) is added.
need to change from pid to pid.decode, as pid is binary and not a string.

### DIFF
--- a/plugins/libvirt/kvm_cpu
+++ b/plugins/libvirt/kvm_cpu
@@ -47,7 +47,7 @@ def config(vm_names):
     @param vm_names : a list of "cleaned" vms' name
     '''
     percent = 100 * len(
-        filter(lambda x: x[0:3] == 'cpu' and x[3] != ' ', open('/proc/stat', 'r').readlines()))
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for taking a look at that plugin!
(sorry: I am just reading your comments and contributions in the wrong order, thus my comments may sound a bit weird and out-of-time)

Could you please take a look at the following details:
* A) line 50 is too long (`flake8` complains about it in https://travis-ci.org/github/munin-monitoring/contrib/builds/761420969)
* B) maybe move the `decode` to `list_pids`

(A) is necessary for the style tests. (B) is just a bonus, if you feel like ...

Thank you!

sumpfralle:
Thank you!

> the line length change doesn't make the world better [..]

... but flake8 is a great tool for spotting unused imports, typos in variable names and a lot more :)
btw: in case you are using python frequently, then I would suggest [black](https://github.com/psf/black) - it lets you forget about code style.
Anyway: thanks!

PR #1186, opened 2021-03-09
beanstalkd python3

lifeofguenter:
minimal effort to get the beanstalkd plugin work with python3 + a more modern/maintained client library.

will require `pip3 install pystalk` to work


sumpfralle:
What a nice improvement - thanks!

PR #1188, opened 2021-03-10
nvme: add graph for spare capacity.  add support for limits

kjetilho:
The default limits are sort of randomly chosen.  My nvme drives report
a warning level at 10% spare capacity, so I kept that here.


sumpfralle:
Thank you!

PR #1190, opened 2021-03-16
Proposed update: all mongodb scripts load MONGO_DB_URI env setting.

jrmsgit:
Hi:

I'm just proposing this patch so we can have "one config line to set them all". If all scripts load the same env var, I'm using MONGO_DB_URI as there were some scripts that were already using that var name (ie: mongodb_docs).

Let me know please if you have any question.

Thanks all for your work on this project.

Cheers,

Jeremías


sumpfralle:
That looks like a good improvement - the users will surely appreciate it.

I am happy, that you provided backwards-compatibility for the previous type of settings.

Thanks!

PR #1191, opened 2021-03-16
Proposed update: mongodb scripts configurable graph category.

jrmsgit:
Hi:

I'm proposing this patch so we can configure the graphs category for mongodb scripts, as some of them were setting it as 'db', but others were using 'mongodb'.

Making it configurable I think it's the best backward-compatibility way to set them under the same category (or not) if needed.

Let me know please if you have any question.

Thanks all for your work on this project.

Cheers,

Jeremías

### DIFF
--- a/plugins/mongodb/mongo_collection_
+++ b/plugins/mongodb/mongo_collection_
@@ -35,6 +35,7 @@
 
 import pymongo
 from operator import itemgetter
+from os import getenv
 
 settings_host = '127.0.0.1'
 settings_port = 27017
@@ -44,6 +45,7 @@ settings_db = 'admin'
 settings_user = ''
 settings_password = ''
 settings_ignoredb = {}
+settings_graph_category = getenv('graph_category', 'db')
 
 typeIndex = {}
 typeIndex['collcount'] = {}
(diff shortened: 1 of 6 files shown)


sumpfralle:
That looks like a reasonable improvement to me. Thank you!

How about unifying the default category for `mongo_mem` to `db`, too?
(otherwise I would just merge your changeset as it is)

jrmsgit:
Hi:

> That looks like a reasonable improvement to me. Thank you!

Thanks for looking into it.

> How about unifying the default category for `mongo_mem` to `db`, too?

I didn't want to do that as I think it could break current setups. I mean, if we change the default it will affect all the current setups.

Personally I use mongodb as category for all of them so the unification would work for me, but that's just myself.

> (otherwise I would just merge your changeset as it is)

Thank you!!

Cheers,


Jeremías

sumpfralle:
> I didn't want to do that as I think it could break current setups.

Personally I consider the plugins in the `contrib` repository as more fluid (less strict), than the ones in the `munin` repository itself.
Thus I would appreciate this unified default (`db`) for all mongo-based plugins (or someone else will do it later somewhen).

Anyway: do you want to do it now or should I just merge the current state?

jrmsgit:
Hi:

> > I didn't want to do that as I think it could break current setups.
> 
> Personally I consider the plugins in the `contrib` repository as more fluid (less strict), than the ones in the `munin` repository itself.
> Thus I would appreciate this unified default (`db`) for all mongo-based plugins (or someone else will do it later somewhen).
> 
> Anyway: do you want to do it now or should I just merge the current state?

Great then! I just pushed the patch to make db the default category for mongo_mem.

Thanks!!

Jeremías

sumpfralle:
Thank you for these improvements!

jrmsgit:
Thanks for your time on this!

PR #1196, opened 2021-03-23
http_responsecode: Some improvements

brknkfr:
- differ between warning and critical
- get only head with curl (--head)
- follow redirects (--location)
- U when not reaching the server

### DIFF
--- a/plugins/http/http_responsecode
+++ b/plugins/http/http_responsecode
@@ -57,7 +57,8 @@ if [ "$1" = "config" ]; then
 	siteid=`echo $site | sed 's/[^a-z]*//g'`
 	echo "$siteid.label $site"
 	echo "$siteid.info HTTP response code statistics for $site."
-	echo "$siteid.critical 99:399";
+	echo "$siteid.warning 99:299";
+	echo "$siteid.critical 399";
     done
     exit 0
 fi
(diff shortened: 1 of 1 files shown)


sumpfralle:

> differ between warning and critical

I think, this change is not effective at the moment, since you enabled "follow redirects"? (I think, there cannot be response codes between 299 and 399 anymore)

> get only head with curl (--head)

This change looks good to me.

> follow redirects (--location)

I am just guessing here, but I think, some (many?) users of this plugin will use very specific URLs to be tracked. Some may be interested in the specific state of redirect responses.

I would suggest to keep the previous state or maybe allow to enable `--location` via an environment variable?

> U when not reaching the server

This is a nice change.
Maybe you could make the regular expression a bit more precise (e.g. `^000$`?).

brknkfr:
> Thank you for your contribution!
> 
> > differ between warning and critical
> 
> I think, this change is not effective at the moment, since you enabled "follow redirects"? (I think, there cannot be response codes between 299 and 399 anymore)
> 
Seems reasonable.
Why I did this change? Sometimes, when a page takes too long to connect, there will be 301 "error" codes, but normally that's just temporary. In my munin config, I set the warning level to notify my only once, while ciritical level always notifies.

> > follow redirects (--location)
> 
> I am just guessing here, but I think, some (many?) users of this plugin will use very specific URLs to be tracked. Some may be interested in the specific state of redirect responses.
> 
> I would suggest to keep the previous state or maybe allow to enable `--location` via an environment variable?
> 
That's a good idea. I'm gonna check to implement that on the `http_response` plugin.

> > U when not reaching the server
> 
> This is a nice change.
> Maybe you could make the regular expression a bit more precise (e.g. `^000$`?).
>
Mhm.

PR #1197, opened 2021-03-23
http_responsetime plugin with curl based on http_responsecode

brknkfr:


brknkfr:
There is already a http_responsetime plugin in the directory. That's the reason I appended a "-2" and that seems to be the reason for failing the travis build.

sumpfralle:

> There is already a http_responsetime plugin in the directory.

Indeed, the following plugins seem to track the response time:
* `http_responsetime`
* `http_response`
* `multi_http_responsetime`
* `http_request_time`

Could you please take a look at these plugins and check, whether one of these already covers the features, that you are looking for?
Maintaining multiple plugins with the same feature is a burden, which I would prefer to avoid.

> That's the reason I appended a "-2" and that seems to be the reason for failing the travis build.

In fact `shellcheck` had some hints regarding your code :)

https://travis-ci.org/github/munin-monitoring/contrib/builds/764194779#L1574

PR #1198, opened 2021-03-26
jitsi-videobridge: videostreams aren't anymore

brknkfr:
Remove videostreams graph (colibri stats doesn't give informations about videostreams).
Specify curl timeout.


sumpfralle:
Thank you!

PR #1201, opened 2021-03-31
http_response: Add follow_redirect and undefined

brknkfr:
- Add "--location" to curl with env.follow_redirect true
- When connections fails, set U (undefined)

### DIFF
--- a/plugins/http/http_response
+++ b/plugins/http/http_response
@@ -11,14 +11,16 @@ http_response - Monitor HTTP response statistics
 
 The following environment variables are used
 
- sites       - Sites to check
-                 - separated by spaces
-                 - can contain basic auth credentials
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for these improvements!

btw: instead of closing your other pull request, you could have also just force-pushed your new commit into it. This approach allows to track the discussion in one place. But this may be a matter of taste. Thanks, anyway!

brknkfr:
> btw: instead of closing your other pull request, you could have also just force-pushed your new commit into it. This approach allows to track the discussion in one place. But this may be a matter of taste. Thanks, anyway!
>
Yes, I know. I was a little bit too fast and yes, it's a matter of taste. Thanks for the feedback.

PR #1202, opened 2021-04-01
nutups2: fix for status line without value

kimheino:
"upsc name" can output line "input.transfer.reason: " (key without value).
Allow it.


sumpfralle:
Thanks!

PR #1203, opened 2021-04-01
knot: cache results, needed for first run after server reboot

kimheino:
Sometimes after reboot munin-async + munin-node runs before knot is
ready. This will result missing knot stats, as there is no static
config in plugin. Cache results and use them instead if knot's output
is empty.

### DIFF
--- a/plugins/knot/knot
+++ b/plugins/knot/knot
@@ -122,7 +122,7 @@ def get_stats():
 
     # After server reboot output can be almost empty. Use cached results
     # instead, needed for plugin config when using munin-async.
-    cachename = os.getenv('MUNIN_PLUGSTATE') + '/knot.state'
+    cachename = os.path.join(os.getenv('MUNIN_PLUGSTATE'), 'knot.state')
     if len(output) > 2048:
         with open(cachename, 'wt') as cache:
             cache.write(output)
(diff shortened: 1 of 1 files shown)


plugins/knot/knot
```diff
@@ -112,14 +113,26 @@ def get_stats():
...
+        output = subprocess.run(['knotc', '--force', 'stats'],
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE, check=False,
+                                encoding='utf-8', errors='ignore').stdout
+    except FileNotFoundError:
         return {}
 
+    # After server reboot output can be almost empty. Use cached results
+    # instead, needed for plugin config when using munin-async.
+    cachename = os.getenv('MUNIN_PLUGSTATE') + '/knot.state'
```
sumpfralle, line 125: `os.path.join(...)` would be better for non-posix platforms, I guess.

sumpfralle (commented):
Maybe take a look at the path composition (commented below). Probably it serves no practical relevance, but is reasonable easy to use here, anyway.

plugins/knot/knot
kimheino: I wonder if it supports non-posix: "Knot DNS itself is written in a portable way and can be compiled and run on most UNIX-like systems, such as Linux, *BSD, and macOS."

But changed anyway....

sumpfralle:
Thanks!

PR #1205, opened 2021-04-04
[plugins/disk/] Fix warning on btrfs_device_stats.

HaseHarald:
It's always those typos!
I don't know why I didn't notice this in all those years. I had "warming" instead of "warning" all over the code. Stupid mistake!


sumpfralle:
hehe - things happen - thanks!

PR #1206, opened 2021-04-06
amavis_multi: multigraph plugin to monitor amavis mail filter

kimheino:
Here's unified, updated and rewritten plugin of my PR #1139.

Two comments:

1) This plugin includes all **useful** data from amavis, and it's easy to add more if there's some ideas. Therefore I'm happy to rename this plugin as "amavis", dropping "_multi", if that's ok?

2) This plugin includes all the info that other plugins (amavis_, amavis_awk, amavis-debian) provide, plus more, all in single multigraph plugin. Other access amavis's db directly (bad, can change) or parse it from log (bad, lot of cpu time). Amavisd-agent was created to provide all that info.

### DIFF
--- a/plugins/amavis/amavis_multi
+++ b/plugins/amavis/amavis_multi
@@ -1,0 +1,189 @@
+#!/usr/bin/env python3
+
+"""Munin plugin to monitor amavisd mail filter status.
+
+=head1 NAME
+
+amavis_multi - monitor amavisd mail filter status
+
+=head1 CONFIGURATION
+
+Following config is needed:
+
+    [amavis_multi]
+    user amavis
+
+=head1 AUTHOR
+
+Kim Heino <b@bbbs.net>
+
+=head1 LICENSE
+
+GPLv2
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+"""
+
+import os
+import subprocess
+import sys
+
+
+def run_binary(arg):
+    """Run binary and return output."""
+    try:
+        return subprocess.run(arg, stdout=subprocess.PIPE,
+                              stderr=subprocess.STDOUT, check=False,
+                              encoding='utf-8', errors='ignore').stdout
+    except FileNotFoundError:
+        return ''
+
+
+def get_values():
+    """Get status from nanny and agent. They are part of amavisd."""
+    ret = {
+        'nanny': {
+            'total': 0,
+            'busy': 0,
+        },
+        'agent': {},
+    }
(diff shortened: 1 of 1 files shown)


yunal:
Awsome. And I've noticed you've been very active here, thanks for that!

Not sure if this is meaningful since it's multigraph and may be totaly different from the old one. But if you named the multigraphs and data series strategically (and also keep the data types) people would be able to keep their history because munin would continue to use the same RRD files for the things with the same names and of course start new RRD files for all the new things the plugin report for.

Regards - an old project maintainer.

kimheino:
History for amavis_content (from amavis_) could have be kept. I didn't do that as I used field names from amavisd-agent for smaller code. I can change that if you like to - I don't think if that matters much.

One more thing: Old plugins have category "antivirus", I use "spamfilter". It's listed in http://guide.munin-monitoring.org/en/latest/reference/graph-category.html and is better one.

sumpfralle:
Thanks for this well-written plugin!

Regarding the plugin name: since there is already an `amavis` plugin distributed directly within munin, I would prefer to avoid a name clash. Maybe `amavisd` could be appropriate, since you are using `amavisd-agent` and `amavisd-nanny`? (or maybe this is stupid - I do not know amavis's naming conventions)

Regarding the history: I think it is acceptable to use new names, since the plugin is not supposed to be a drop-in replacement of `amavis`.

Thank you!

kimheino:
Amavis naming is a bit of a mess. It has been called amavisd, amavis-ng and now it's amavis. Binary names are still with "d", so "amavisd-nanny" etc.

So I really want to call this plugin as amavisSOMETHING, not amavisdSOMETHING. Any ideas?

sumpfralle:
> So I really want to call this plugin as amavisSOMETHING, not amavisdSOMETHING. Any ideas?

That is quite a challenge :)

Indeed I can only think of boring suffixes:
* `_multi`: your choice up to now
* `_py`: may look weird due to the underscore instead of a dot

Sorry, that is probably not too inspiring :(

Feel free to pick your favorite ...

btw: in your script you use the term `childs`. Should this be `children` or is it an amavis-specific term?

kimheino:
> * `_multi`: your choice up to now
> * `_py`: may look weird due to the underscore instead of a dot

`_py` is just an implementation detail and doesn't give any extra information to end user. I still prefer `_multi` so I'll keep that.

> btw: in your script you use the term `childs`. Should this be `children` or is it an amavis-specific term?

You're correct, I once again misspelled it. It actually should be "processes", so fixed.

sumpfralle:
yeah - thanks!

PR #1208, opened 2021-04-14
signaling: Plugin for spreed standalone signaling server

brknkfr:
Plugin for spreed standalone signaling server used for Nextcloud Talk (https://github.com/strukturag/nextcloud-spreed-signaling)


sumpfralle:
Thank you for this well-written plugin!

PR #1210, opened 2021-04-18
[plugins/disk/] Allow for configuration of warning levels on btrfs_device_* plugins

HaseHarald:
This allows for an individual configuration of warning (and critical) levels of the btrfs_device_stats and btrfs_device_usage plugins. The changes are designed to be backwards compatible. So you don't have to adapt your configuration if you don't want to. But at least you can now.


sumpfralle:
Thank you for your good enhancements of these plugins!

Just minor comments for the python style:
* it is quite common to use `None` instead of `False` for undefined value
* string formatting is usually easier to read than string concatenation (but the rest of the plugin already used concatenation before, anyway ...)

I merged your commits (excluding the merge commits) manually. Thank you!

PR #1211, opened 2021-04-27
redis: fix "autoconf" to return "no" if no instances are configured

kimheino:
Previously it returned "yes" and empty graph with all labels having
value "0". Also remove bogus "multigraph redis" line.

TODO: die() in get_conn() makes this plugin to fail if any Redis
instance can't be connected.


sumpfralle:
Thank you for cleaning the plugin up a bit!

Issue #1212, opened 2021-05-01
docker_ plugin: disparity of sanitized container names results in missing data


dneuge:
`docker_` plugin sanitizes container names through `clean_fieldname` in configuration but not when dumping actual values; for example dots (`.`) are being replaced by underscores (`_`). This leads to a disparity which affects `cpu`, `memory` and `network` graphs. Graphs will only show `-nan` instead of a value for affected containers.

### Example
For a container name `my.container` configuration will announce keys containing `my_container` although values are still printed as `my.container`.

sumpfralle:
fixed in #1213, thanks!

PR #1213, opened 2021-05-01
docker_: sanitize value container names

dneuge:
docker_ plugin sanitized container names when printing config but
not values which led to a disparity resulting in no data on graphs
for all containers whose names are subject to sanitization. (#1212)


dneuge:
Unless anything else comes up during review, issue #1212 can be closed after pulling these changes.

sumpfralle:
Thank you for fixing this!

PR #1214, opened 2021-05-04
Added check to ensure network stats for container exist before reading them

zsoi:
In some cases, the docker plugin will not report a "networks" statistics dictionary when no networks are attached to a container (e.g. when attaching to host network).

This pull request adds a check to ensure the network stats exist before reading from them. Without this check, the `docker_network` plugin can exit with an error under these circumstances and won't produce any valid data.


sumpfralle:

I just merged your commit manually, since other changes were just merged into that plugin ...

PR #1215, opened 2021-05-05
LXC v3.1 compatibility

ls-initiatives:
The cgroup data was (yet again) moved to a new location in recent LXC
releases, starting with v3.1.0 from what I could gather.

https://github.com/lxc/lxc/issues/2782 states:
> [The cgroup lxc.payload] exists to adhere to cgroup2 delegation requirements.


sumpfralle:
Thanks for following up with this quickly moving target!

PR #1216, opened 2021-05-14
nutups2_: "@" is invalid character in plugin name so use "." instead

kimheino:
This is needed when you monitor local and remote UPSes from single
host. You must be able to specify "ups@host" in plugin name so do
it as "ups.host".


sumpfralle:
Thank you!

PR #1219, opened 2021-05-27
nextcloud_groupfolders_: Monitor sizes of groupfolders

brknkfr:
Small plugin based on nextcloud_ to monitor in nextcloud specified groupfolders


sumpfralle:
Thank you for this well-looking plugin!

Just a minor comment regarding the style: you could replace `jq -r .foo | sed 's/ null$/ U/'` with `jq -r ".foo // \"U\""` (in order to avoid ambiguities of the raw sed-based replacement).

PR #1220, opened 2021-06-07
Fix some AUTHORS so they all show properly in the plugin gallery

shtrom:


sumpfralle:
Thank you!

PR #1221, opened 2021-06-07
Snmp  if combined stacked

shtrom:
This PR fixes a few bugs in the `snmp__if_combined` plugin, then adds a `stacked` option to stack each port's traffic into a summary graph for the root.


shtrom:
Please, do not merge this PR just yet. I think there is a bug in the `Total traffic` logic, and I want to add example graphs.

The rest of the code is good for review, though.

plugins/snmp/snmp__if_combined
```diff
@@ -631,7 +679,22 @@ sub do_config_if {
 
     my $warn = undef;
     my $speed = 0;
-    my $extrainfo = "";
+
+    my $extrainfo="";
+
+    if (defined ($snmpinfoX->{$if}->{ifHCInOctets})) {
+	# If we get an answer at the 64 bit OID then this switch
```
sumpfralle, line 686: I think, these lines deserve a bit of indentation?

plugins/snmp/snmp__if_combined
```diff
@@ -631,7 +679,22 @@ sub do_config_if {
...
+	# supports the extended MIB
+
+	$extrainfo .= " This switch supports 64 bit byte counters and these are used by this plugin.";
+    } else {
+	# If not we only have a 32 bit counter and are lost.
```
sumpfralle, line 691: Indentation?

sumpfralle (commented):
Thank you for this contribution!

I just noticed minor issues with indentation (which may be caused by inconsistent style in the file).
Besides this, I did not see any problems (but I hardly know the plugin).

Just tell me, when you think, it is ready ...

shtrom:
Ok, @sumpfralle I think it's good to go.

Seeing as the bug is not new, and dates back to `snmp__if_multi`, I simply documented it, and left some notes about how to fix the data when it happens.

sumpfralle:
Great!

Your comments regarding jumping sums due to wrapping sounds reasonable.
*Maybe* this could be solved by using a cdef (summing up the fields after they are wrap-cleaned)?

Anyway: I would merge it as it is. Just confirm ...

shtrom:
Cdef was my first approach but
1. It doesn't create loanable series
2. It doesn't play well with negatives (same as `graph_total`
-- 
Olivier Mehani ***@***.***>
Sent from my mobile, please excuse my brevity.

sumpfralle:
Thanks for clarifying this. I merged your improvements now - thanks again!

PR #1222, opened 2021-06-12
bugfix cpubyuser plugin

mkadlof:
Prior plugin version returns wrong results. I'm not sure why.
This commit fix the bug however it reduce functionality since
there is no way now to include/exclude certain users.


sumpfralle:

I have the feeling, that the exclusion of certain users is a relevant part of this plugin.
Thus I would prefer to keep this feature, if possible.

Maybe you could describe in detail, what kind of problems you noticed?

PR #1225, opened 2021-07-12
ssl-certificate-expiry: fix handling domains starting with digits

### LINKED ISSUE #1224: ssl-certificate-expiry loses results of domains starting with digits
I have domains `0.example.com`, `1.example.com` and so on.

`ssl-certificate-expiry` uses `clean_fieldname` function on them, which converts them all into `__example_com`.

My solution proposal: `fieldname=$(clean_fieldname "_${service}")`. Prepending `_` allows `clean_fieldname` function to retain the digit.

etam:
resolves #1224

### DIFF
--- a/plugins/ssl/ssl-certificate-expiry
+++ b/plugins/ssl/ssl-certificate-expiry
@@ -165,6 +165,11 @@ print_expire_days() {
     fi
 }
 
+my_clean_fieldname() {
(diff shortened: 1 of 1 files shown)


niclan (changes requested):
This will loose history on everyones existing graphs to fix yours.  IMHO you should make it conditional on the name starting with a digit.

sumpfralle:
This looks good to me - thank you!

PR #1227, opened 2021-07-19
Multigraph root loans

shtrom:
Loan data from sub-graphs rather than storing it in duplicated series in the root graph of some multigrapdh plugins.


sumpfralle:
That looks good to me, thank you!

btw: you used two hyphens for the range of years in copyright statements (e.g. `2016--2021`). Was this intentionally?

shtrom:
On Mon 19 Jul 2021 at 04:27:28 -0700, Lars Kruse wrote:
>btw: you used two hyphens for the range of years in copyright statements (e.g. `2016--2021`). Was this intentionally?

Ah, old LaTeX habits die hard. Those get typeset as ndashes.

-- 
Olivier Mehani ***@***.***>

Issue #1228, opened 2021-07-21
docker docker_containers throws an error


kaihen:
The current version throws following error:
`Traceback (most recent call last):
  File "/etc/munin/plugins/docker_containers", line 552, in <module>
    main()
  File "/etc/munin/plugins/docker_containers", line 540, in main
    globals()[wildcard](client, mode)
  File "/etc/munin/plugins/docker_containers", line 391, in containers
    print('containers_quantity.value', len(client.containers))
AttributeError: 'ClientWrapper' object has no attribute 'containers'`

The error is that the method `def containers(self)` (line 143-146) got deleted in one of the last commits (https://github.com/munin-monitoring/contrib/commit/a3ae1af155e2fa07b3c8b97fd7e538935b059705) and not all depency got removed... In line 391 need be `print('containers_quantity.value', len(client.all_containers))` or the method `def containers(self)` need to be added again.

Issue #1178, opened 2021-02-19
docker cpu is not using env.EXCLUDE_CONTAINER_NAME


ogmueller:
The docker plugin should use `env.EXCLUDE_CONTAINER_NAME` as stated in the documentation of the plugin, but this is not used when e.g. `munin-run docker_cpu` is used.

The issue is in function `containers(self)`, which is not filtering the cointainer list. It should do so using:
```perl
 124         return [c for c in self.client.containers.list()
 125                 if not self.exclude
 126                 or not self.exclude.search(c.name)]
```
instead of
```perl
 124         return self.client.containers.list()
```

sumpfralle:
Your recommendation for `def containers(self)` looks similar to the current code of `def all_containers(self)`.
Maybe we should replace all(?) references to `.containers` with `.all_containers`?
(I cannot judge myself, since I do not use the plugin)

ogmueller:
As far as I can see, the `def containers` is only called by `munin-run docker_containers`, which returns the total numbers of running/restarting containers like `docker ps`.

In contrast to that, all_containers will also retrieve other containers (e.g. paused, stopped, dead, exited, ...) like `docker ps -a`.

So both function might be neccessary, but they probably should filter the results depending on the exclusion list in `env.EXCLUDE_CONTAINER_NAME`.

sumpfralle:
I *think*, this is fixed now (a3ae1af155e2fa07b3c8b97fd7e538935b059705), since `all_containers` is used everywhere in the plugin (and thus the filtering applies).

Please re-open, if you think, that something is missing.

PR #1229, opened 2021-07-22
Fix for docker_containers

kaihen:
This PR just fixes the issue (https://github.com/munin-monitoring/contrib/issues/1228).

The commit (https://github.com/munin-monitoring/contrib/commit/a3ae1af155e2fa07b3c8b97fd7e538935b059705) removed the method `client.containers` and didn't changed/removed all dependencies of the script.

### DIFF
--- a/plugins/docker/docker_
+++ b/plugins/docker/docker_
@@ -388,7 +388,7 @@ def containers(client, mode):
         print("graph_category virtualization")
         print("containers_quantity.label Containers")
     else:
-        print('containers_quantity.value', len(client.containers))
+        print('containers_quantity.value', len(client.all_containers))
 
 
 def images(client, mode):


kaihen:
Just watched the other issues and read the discussion of the issue (https://github.com/munin-monitoring/contrib/issues/1178). 

The pull request just fixes that there is no more trace output. The output of `docker_containers` is nearly identically with the output of `docker_status` (=> RUNNING) because the `client.all_containers` filters on RUNNING (`if (c.status == 'running') and (not self.exclude or not self.exclude.search(c.name))`).

The graph `docker_containers` is basically redundant because the information of the graph (value/output) is contained in other graphs (like `docker_status` => All containers).

sumpfralle:
Thank you for fixing this issue!

Regarding your other comments: personally I am not a user of the plugin, thus I cannot really tell.
Any improvements, that you could suggest, would be very welcome!

PR #1230, opened 2021-07-23
bind9_*_stats: fix using env variable stat_file

ap-wtioit:
previously only stat_file would only be used for data cleanup not for
fetching the data. also fixed some shellcheck warnings.

This fixes the following problem:
If you configured `stat_file` to be `/var/run/named/named.stats` like this:
```
[bind9_*]
   env.stat_file /var/run/named/named.stats
```
and enabled writing statistics with `statistics-file "/var/run/named/named.stats";` in bind9
you got the error `cat: /var/cache/bind/named.stats: No such file or directory` when running the plugin.

With this fix stat_file is also considered for `cat` not only for `rm`.

I fixed some shellcheck stuff in the affected files as well, but didn't want to change too much for a simple bugfix (e.g. there is still unecessary cat uses, chaining of seds and some other stuff in there).

Info @wt-io-it

### DIFF
--- a/plugins/bind/bind9_resolver_stats
+++ b/plugins/bind/bind9_resolver_stats
@@ -40,19 +40,19 @@ section="Resolver Statistics"
 # - use sed to get lines after (and excluding) the section start tag
 # - use sed to get lines up to (and excluding) the next section start tag
 # - use grep to remove any lines starting with a [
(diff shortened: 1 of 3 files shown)


sumpfralle:
Great - thanks!

PR #1226, opened 2021-07-14
nvidia_gpu_: add Rx/Tx throughput outputs

byteneumann:

PR #1231, opened 2021-07-26
nvidia_gpu: fix units of Rx/Tx throughput (multiple gpus)

byteneumann:
Today, I noticed that my recent PR for Rx/Tx throughput outputs #1226 does not give correct readings on systems with multiple gpus.

The problem is that the multiplication `$((valueGpus*1024))` does not work for unseparated input values (as returned by the echo-grep-cut combo). The result is unscaled data being processed by munin - therefore showing kilobits per second as bits per second.

Here is a fix for that! The code replicates the loop already found under `mem)` (lines 250-259).

Sorry for having overlooked this issue.


sumpfralle:
Thank you for fixing the issue!

PR #1232, opened 2021-07-26
Update the Deluge plugin to handle Deluge 2.x

kmkr:
Version 1.0.0 of this plugin requires version 1 of Deluge. Deluge 2 was released in June 2019.

This commits bumps the plugin to version 2.0.0.

This plugin has three modes. The modes "bandwidth" and "states" are working equally to the previous version.  The third mode, "connections", is changed.

The connections mode previously printed one value: the total number of connections.  That RPC call is removed in Deluge 2, and I'm instead using `get_session_status`, which delegates the call to libtorrent's [session statistics function](https://libtorrent.org/manual-ref.html#session-statistics). libtorrent is better documented, so it's easier to work with compared to Deluge.

"Number of peers" replaces the previous "number of connections", and I've added half open peers, and peers interested in download and upload slots to the same graph.

## Screenshot

Here's how the plugin is rendered.

filetransfer (Munin)

### DIFF
--- a/plugins/torrent/deluge_
+++ b/plugins/torrent/deluge_
@@ -1,14 +1,14 @@
-#! /usr/bin/env python2
+#! /usr/bin/env python3
 
 """=cut
 =head1 NAME
 
-deluge_ - Munin wildcard plugin to monitor Deluge torrent client
+deluge_ - Munin wildcard plugin to monitor the Deluge torrent client
 
 =head1 REQUIREMENTS
 
- - Python2.5+ (Deluge itself won't work with python3)
- - Deluge
+ - Python3+
+ - Deluge2+
 
 This plugin also uses
  - deluge.ui.client
@@ -19,13 +19,13 @@ These modules are required by Deluge itself.
 =head1 INSTALLATION
 
 This plugin has 3 modes :
- - connections : monitors the number of connections
+ - peers : monitors the number of peers
  - bandwidth : monitors the bandwidth (up, up overhead, down, down overhead)
  - states : monitors the torrents' states
 
 To use one of these modes, link the this plugin as 'deluge_<mode>'
 For example :
-ln -s /path/to/deluge_ /etc/munin/plugins/deluge_connections
+ln -s /path/to/deluge_ /etc/munin/plugins/deluge_peers
 
 =head1 CONFIGURATION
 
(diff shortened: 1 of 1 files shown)


plugins/torrent/deluge_
```diff
@@ -207,10 +222,13 @@ class StatClient:
                 self.end_session,
                 errbackArgs=("get_session_state failed"))
```
sumpfralle, line 223: Please add an `else` branch here (and around line 378) and emit an error message in case of an unknown plugin view.
This is specifically relevant, since the update removes a previously working view ("connections"). Users of the previous plugin should get a fair chance to notice the change.

plugins/torrent/deluge_
```diff
@@ -396,4 +413,4 @@ if len(sys.argv) > 1:
     else:
         fetch_info(get_mode())
 else:
-    fetch_info(get_mode())
+    fetch_info(get_mode())
```
sumpfralle, line 416: Please do not remove the final newline (by convention a text file ends with the *end of line* marker).

sumpfralle (commented):
Thank you for improving the plugin and for porting it to python3!

plugins/torrent/deluge_
kmkr: I did that now. 

It's not really required with the current code flow since the plugin confirms the plugin view via the `get_mode` function. I still added the `else` clause to solve your comment. Let me know if you think it's overkill given this new information and I can remove it again.

```
def get_mode():
    script_name = os.path.basename(sys.argv[0])
    mode = script_name[script_name.rindex('_') + 1:]

    log.debug("Mode : %s", mode)

    if mode not in modes:
        log.error("Unknown mode '%s'", mode)
        log.info("Available modes are : %s", modes)
        sys.exit(1)

    return mode
```

plugins/torrent/deluge_
kmkr: Yes that was my editor that wasn't configured correctly. Sorry, and fixed!

plugins/torrent/deluge_
sumpfralle: Thank you for committing to my style obsessions :)

Personally I always feel the urge to verify such checks, whenever I stumble upon a non-trivial `if/elif` flow. Adding the `else` clause takes this burden away from me (and maybe others). This somehow follows the thought of "code is read much more often than it is written" ...

sumpfralle:
Thank you!

PR #1234, opened 2021-08-06
rtom_vol: remove unused config parameters

tofurky:
Since there are no associated values for 'hashing', 'seeding', and
'active', a warning is seen in munin-update.log every time this plugin
is fetched.


sumpfralle:
The same issue is also visible in `rtom_allsessions_vol`. I fixed it there, now.

Thank you!

PR #1235, opened 2021-08-11
Improve code readability and consistency

kmkr:
This PR:

- Changes to Python 3 template string rather than the formatted string syntax.
- Calls the Deluge API once per plugin run for the "states" mode, rather than once per torrent. Lines of code are reduced. The readability is improved too, since the API call is similar to how the "peers" mode works.
- Updates the label for the "state" entry "paused" to match it's name in the libtorrent API: "stopped". "Queued" is split into "Queued  seeding" and "Queued downloads" to match the libtorrent API too.

## Screenshot

sc


sumpfralle:
Thank you for these welcome improvements of the plugin!

While you are at it, I would be happy if you could take a look at the following additionals details, if you feel like:
* verify that the style is accepted by flake8 (I think, the whitespace in your tuples would be rejected)
* maybe [add your example graphs](http://guide.munin-monitoring.org/en/latest/develop/plugins/plugin-gallery.html#rules-for-plugin-contributors) to the repository (the gallery will appreciate this)
* I would suggest to lean towards double quotes (currently there is a mix) or simply use [black](https://github.com/psf/black/) for a full reformatting

What do you think?
(I would merge your current improvements, if you are not in the mood for further beatification ...)

kmkr:
Thank you for your kind feedback. I agree; all your proposals make sense. I'll be away for vacation for a few days so I'll work on your suggestions next week :)

kmkr:
I've uploaded example graphs, and reformatted with black and flake8 :heavy_check_mark:

sumpfralle:
Yeah - thanks a lot!

PR #1236, opened 2021-08-17
Add swichbotmeter plugin

shakemid:
Hi,
I would like to add switchbotmeter plugin. 
SwitchBot Meter is IoT temperature/humidity meter and has capability to get data with REST API. 
https://www.switch-bot.com/products/switchbot-meter
This plugin aims to get temperature/humidity from SwitchBot Meter by using REST API.
Please review the pull request.

Thank you,
K.Cima


plugins/sensors/switchbotmeter
```diff
@@ -0,0 +1,128 @@
...
+}
+
+# Main
+case ${1:-} in
+autoconf)
+    autoconf
+    ;;
+config)
+    config
+    [ "${MUNIN_CAP_DIRTYCONFIG:-}" = "1" ] && fetch
```
sumpfralle, line 121: Just for the style: this compound statement can be tricky, if `errexit` is enabled, since it could turn the exitcode of the `case` command into non-zero (and cause an early exit with failure).
Thus it is advisable to avoid the above compound statement *as the last operation* anywhere (or simply: always use `if` instead).

Example:
```shell
$ (set -e; if true; then true && false; fi; echo "FINISH"); echo $?
1
````

Just a hint - feel free to ignore this.

File: plugins/sensors/switchbotmeter [hunk 2]
```diff
@@ -0,0 +1,128 @@
...
+deviceid=${deviceid:?}
+
+# Interval for API access (second)
+interval=${interval:-0}
+
+# Temperature unit
+tempunit=${tempunit:-C}
+
+autoconf() {
+    echo 'no'
```
sumpfralle, line 62: Please do not pretend to handle `autoconf`, if you do not advertise this feature.

plugins/sensors/switchbotmeter
```diff
@@ -0,0 +1,128 @@
...
+  env.interval 900, the API response will be cached to local file for 900 seconds(15 minutes).
+
+=head1 AUTHOR
+
+  K.Cima https://github.com/shakemid
+
+=head1 LICENSE
+
+  GPLv2
+
```
sumpfralle, line 36: Maybe you want to add an SPDX identifier:
```
SPDX-License-Identifier: GPL-2.0-only
```

plugins/sensors/switchbotmeter
```diff
@@ -0,0 +1,128 @@
...
+    echo temperature.value "${temperature}"
+    echo humidity.value "${humidity}"
+}
+
+fetch_api() {
+    response=$( curl -s -H "Authorization:${token}" "https://api.switch-bot.com/v1.0/devices/${deviceid}/status" )
+
+    statusCode=$( echo "${response}" | jq '.statusCode' )
+    if [ "${statusCode}" -ne 100 ]; then
+        echo Error with statusCode = "${statusCode}"
```
sumpfralle, line 106: Please send this message to `stderr`.

sumpfralle (commented):

It looks good to me.
Please take a look at my minor nitpicks. Most of them are just cosmetic - feel free to ignore these.

shakemid:
Thank you for your review! I will fix it.

plugins/sensors/switchbotmeter
shakemid: I will fix it.

plugins/sensors/switchbotmeter
shakemid: All right. I will add it.

plugins/sensors/switchbotmeter
shakemid: I agree. I will change it to `if`

plugins/sensors/switchbotmeter
shakemid: Does 'advertise' mean `capabilities` statement in magic markers? I would like to remove autoconf function and case selection.  Is it ok?

shakemid:
@sumpfralle 
I have tried to fix it. Could you check it?
Thanks!

sumpfralle:
Thank you for contributing this new plugin and for going through all my nitpicking remarks!

shakemid:
Your detailed reviews are totally acceptable. Thank you, too!

PR #1237, opened 2021-08-31
Add switchbotmeterbt plugin

shakemid:
Hi,
I would like to add switchbotmeterbt plugin.
This plugin aims to get temperature/humidity from SwitchBot Meter by using BLE(Bluetooth Low Energy).
Please review the pull request.

Thank you,
K.Cima


plugins/sensors/switchbotmeterbt
```diff
@@ -0,0 +1,202 @@
...
+  env.scantimeout : Timeout for BLE scan. (Optional, default is 5.0 seconds)
+
+  Example:
+    [switchbotmeterbt]
+      user  root
+      env.macaddr  aa:aa:aa:aa:aa:aa bb:bb:bb:bb:bb:bb cc:cc:cc:cc:cc:cc
+
+=head1 NOTES
+
+  For more details about SwitchBot Mater, see https://www.switch-bot.com/products/switchbot-meter
```
sumpfralle, line 30:
```suggestion
  For more details about SwitchBot Meter, see https://www.switch-bot.com/products/switchbot-meter
```

plugins/sensors/switchbotmeterbt
```diff
@@ -0,0 +1,202 @@
...
+"""
+
+import sys
+import os
+from bluepy.btle import Scanner, DefaultDelegate
+
+
+class SwitchbotScanDelegate(DefaultDelegate):
+    def __init__(self, macaddrs):
+        DefaultDelegate.__init__(self)
```
sumpfralle, line 56: I think, `super().__init__()` is the modern approach for calling parent methods.

sumpfralle (commented):

I noticed only two minor details, which could justify a change. The rest looks fine to me.

plugins/sensors/switchbotmeterbt
shakemid: Thanks! I am not sure which is better, but in the sample code of bluepy, `DefaultDelegate.__init__(self)` is used. What do you think?
https://ianharvey.github.io/bluepy-doc/scanner.html

plugins/sensors/switchbotmeterbt
shakemid: Thanks! I will fix it.

plugins/sensors/switchbotmeterbt
sumpfralle: The `super` call is preferable.
A longer explanation can be found [here](https://stackoverflow.com/a/33469090).
Feel free to suggest this change to the maintainer of the above example code :)

shakemid: Thank you! I would like to fix it.

shakemid:
@sumpfralle 
I have fixed the plugin. Could you review it again?

sumpfralle:
Thank you for reminding me. The new plugin looks fine to me, thanks!

PR #1238, opened 2021-09-17
Add plugin tcp_reachability

klaus-tux:
This pull request adds a plugin to monitor the reachability of (remote) TCP ports. An alarm is triggered if a configured target is not reachable. Additionally the connection time statistics are monitored.


sumpfralle:
Thank you for preparing this proposal!

Could you please check, whether one of the existing tcp-related plugins already solve the same problem?
* [tcp](https://github.com/munin-monitoring/munin/blob/stable-2.0/plugins/node.d.linux/tcp.in) (in core)
* [multi_tcp_ping](https://github.com/munin-monitoring/contrib/blob/master/plugins/ping/multi_tcp_ping)

If you think, that another plugin is worthwhile, then I would be happy, if you could describe its specific features compared to the other ones in the perlpod header.

And please remove the merge commit from history.

Thank you!

PR #1241, opened 2021-09-28
nvme: enable specifying warnings and criticals for serial numbers

ap-wtioit:
On one of our servers equiped with an ASRock Rack X470D4U Mainboard running Ubuntu 18.04 LTS when we reboot the server we may end up with a new order of /dev/nvme* devices. So in the past we would need to reconfigure the warnings per device name at every reboot.

With this patch it is now possible to specify warnings and criticals with the NVME device serial number (label of the graph) as well.
The previously existing configuration options still take precedence over the new one.


sumpfralle:
This looks fine to me - thank you for this improvement!

PR #1242, opened 2021-10-09
[wunderground_] Allow to disable alerts with `env.alerts no`

shtrom:


sumpfralle:
Thank you for your proposal!

What do you think about not providing warning/critical thresholds by default and letting users specify these on their own?
The other plugins are doing this via `print_warning` or similar.
I think, this approach (warnings disabled by default) is suitable for plugins, if there is no universal understanding of a common "warning" or "critical" threshold. And temperatures and the like sound like a good example for a metric, which has no universal "optimum" :)

What do you think?

*If* you would like to prefer to stick to the approach you chose in this PR, then I would suggest to rephrase the following:

> env.alerts yes			# optional, allows to disable all alerts, this is the default

Here I misunderstood `this is the default`, when going through the changeset. Maybe you can clarify its wording.

Thank you!

PR #1243, opened 2021-10-14
fronius: support jq builds without math library

kimheino:
"round" is part of jq's math support. Some distros (rhel8) don't enable it.
Fortunately round can be emulated with "floor" which is always present in
modern jq.


sumpfralle:
Great - thanks!

PR #1244, opened 2021-10-15
systemd_mem: improvements

ap-wtioit:
The following changes are made with this:
* the plugin now reports unknown values as `U` instead of `0`
* when configuring the new option `env.all_services true` it automatically monitors all running services
* on some of our systems unknown / not set values where reported as `18446744073709551615` (UINT64_MAX). this is now handled as unknown / unset by the plugin as well

Info @wt-io-it

### DIFF
--- a/plugins/systemd/systemd_mem
+++ b/plugins/systemd/systemd_mem
@@ -58,12 +58,19 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SO
 
 . "$MUNIN_LIBDIR/plugins/plugin.sh"
 
+# at least on debian stretch services report this as unlimited / unknown value (2^64)
+# see https://lists.freedesktop.org/archives/systemd-devel/2018-July/041092.html
+SYSTEMD_UNLIMITED="18446744073709551615"
+
+all_services=${all_services:-}
+services=${services:-"munin-node"}
+
 for_services() {
     if [ "$all_services" != "" ]; then
         systemctl --type=service --state=running --no-pager --no-legend --output=short-precise \
           | awk '{sub(".service$", "", $1); print $1}'
     else
-        echo ${services:-"munin-node"}
+        echo "$services"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


plugins/systemd/systemd_mem
```diff
@@ -51,28 +58,51 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SO
 
 . "$MUNIN_LIBDIR/plugins/plugin.sh"
 
+# at least on debian stretch services report this as unlimited / unknown value
+SYSTEMD_UNLIMITED="18446744073709551615"
```
sumpfralle, line 62: This value is `2 ^ 64`, thus it may be suitable only for 64 bit machines?
Anyway: maybe there is no alternative for using this hard-coded number. Thus a reference to the 64 bit issue in the comment should be sufficient.

plugins/systemd/systemd_mem
```diff
@@ -51,28 +58,51 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SO
...
+
+for_services() {
+    if [ "$all_services" != "" ]; then
```
sumpfralle, line 65: As a matter of *declaring the interface*, I would suggest to add the following line somewhere at the beginning of the script:
```shell
all_services=${all_services:-}
```

This allows you to use `set -u` and it "declares" prominently, that the plugin is using this environment variable (interpreting it as an empty string, if unset).

The same would probably be good for `services`, too.

plugins/systemd/systemd_mem
```diff
@@ -51,28 +58,51 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SO
...
     echo "graph_title Systemd Service Memory Usage"
     echo "graph_info Memory usage for selected services as reported by systemctl"
     echo "graph_vlabel bytes"
     echo "graph_args --base 1024 --lower-limit 0"
     echo "graph_category memory"
-    for service in ${services:-"munin-node"}; do
+    for service in $(for_services); do
         clean_name="$(clean_fieldname "$service")"
-        description=$(systemctl show "$service" --all | grep ^Description= | cut -d '=' -f 2)
+        description=$(systemctl show "$service" --property=Description | cut -d '=' -f 2)
```
sumpfralle, line 81: Maybe use `-f 2-` (instead of `-f 2`) - otherwise content including an equal character is truncated.

sumpfralle (commented):
Thank you for your improvements for this plugin!

Please take a look at my comments.

plugins/systemd/systemd_mem
ap-wtioit: thanks for the hint. did that for `all_services` and `services`

plugins/systemd/systemd_mem
ap-wtioit: this came from the original plugin. nonetheless `-f 2-` seems to be a better fit for description. for the memory numbers: as we do not expect an `=` to make an appearance in there i kept the `-f 2` for them.

PR #1245, opened 2021-10-15
letsencrypt_weekly: display by registered domain

ap-wtioit:
This enables letsencrypt_weekly to display CSRs per registered domain instead of per server. So for servers with multiple registered domains this now also displays seperate numbers per registered domain.
The plugin tries to detect if a CSR is for a new domain or a renewal (by checking the archive).

In turn by reporting the values per registered domain this enables a more meaningful aggregation if multiple node/server handle certificates for the same domain.


plugins/ssl/letsencrypt_weekly
```diff
@@ -49,13 +107,42 @@ elif [ "$1" = "config" ] ; then
...
+		file=${file_domain% *}
+		domain=${file_domain#* }
+		registered_domain_key=$(echo "$domain" | sed 's/.*\.\([a-z0-9-]\+\.[a-z]\+\)/\1/;s/[-.]/_/g')
+		previous_certs=$(find "/etc/letsencrypt/archive/$domain" -name 'cert*.pem' -not -cnewer "$file" | wc -l)	
+		if [ "$previous_certs" -gt 0 ] ; then
+			value_key="${registered_domain_key}_renewal_weekly.value "
+		else
+			value_key="${registered_domain_key}_weekly.value "
+		fi
+		old_value=$(echo -e $"$values" | grep "$value_key" | sed 's/.* //g')
```
sumpfralle, line 143: Is the `gettext` prefix (`$"...`) useful here? (and a few lines below)

plugins/ssl/letsencrypt_weekly
```diff
@@ -49,13 +107,42 @@ elif [ "$1" = "config" ] ; then
...
+		value=$((old_value + 1))
+		values=${values//$value_key$old_value/$value_key$value}
+	done < <(get_files_and_domains)
```
sumpfralle, line 146: Is the first `<` relevant?
(I do not use bash too often, thus I may be wrong)

plugins/ssl/letsencrypt_weekly
```diff
@@ -1,20 +1,69 @@
...
 : << =cut
 =head1 NAME
 
 letsencrypt_weekly - monitor the number of CSRs by week for /etc/letsencrypt/csr/
 
+=head1 DESCRIPTION
+
+This plugin monitors the number of certificate signing requests (CSRs) done with letsencrypt's certbot.
+It tries to determine the registered domain and reports the number also per registered domain.
+For People running multiple servers this enables aggregation of those numbers across multiple nodes in the munin.conf
```
sumpfralle, line 11: *People* should be lower case?

plugins/ssl/letsencrypt_weekly
```diff
@@ -41,6 +90,15 @@ warning=${warning:-:40}
 critical=${critical:-:50} #letsencrypt doesn't allow more than 50 certificates per week
 # see https://letsencrypt.org/docs/rate-limits/
 
+
+get_files_and_domains() {
+	find /etc/letsencrypt/csr/ -mtime -7 -type f -print0 2>/dev/null | xargs -0 -I pem bash -c 'echo -n "pem "; openssl req -in pem -text -noout | grep DNS: | sed "s/.*DNS://g"'
+}
+
+get_registered_domains() {
+	find /etc/letsencrypt/archive/ -mindepth 1 -maxdepth 1 | sed 's,.*/,,;s/.*\.\([a-z0-9-]\+\.[a-z]\+\)/\1/;s/.* //' | sort | uniq
```
sumpfralle, line 99: Maybe adding spaces between the `sed` commands could increase readability :)

plugins/ssl/letsencrypt_weekly
```diff
@@ -41,6 +90,15 @@ warning=${warning:-:40}
 critical=${critical:-:50} #letsencrypt doesn't allow more than 50 certificates per week
 # see https://letsencrypt.org/docs/rate-limits/
 
+
+get_files_and_domains() {
+	find /etc/letsencrypt/csr/ -mtime -7 -type f -print0 2>/dev/null | xargs -0 -I pem bash -c 'echo -n "pem "; openssl req -in pem -text -noout | grep DNS: | sed "s/.*DNS://g"'
```
sumpfralle, line 95: `--no-run-if-empty` would be suitable here (for `xargs`), or?

plugins/ssl/letsencrypt_weekly
ap-wtioit: yes this is necessary, (in my words maybe not the best explanation):
* the first `<` prepares the while loops stdin to be provided by the given file(descriptor)
* the second `<(...)` is process substitution which prepares a filedescriptor that can only be read by your process and is fed by the command given (`echo <(echo "ok")` displays `/dev/fd/63`, `cat <(echo "ok")` gives you `ok`)

This allows the main process to hold and change the variables while the following would not allow changing variables in the while loop (because the while loop would run in a subprocess):
```
get_files_and_domains | while read -r line; do
    # cannot change parent variables here (running in subprocess)
    echo "$line"
done

while read -r line; do
    # can change variables here (running in main process)
    echo "$line"
done < <(get_files_and_domains)
```

plugins/ssl/letsencrypt_weekly
ap-wtioit: when using `-I ...` you do not need `--no-run-if-empty` because `-I` can only convert the variable to a value if a value is found, if no value is found it does not invoke the command with empty argument:
```bash
set -x
find /tmp -maxdepth 1 -name propably_not_existing -print0 | xargs -0 -I element echo element
```
shows no invocation of echo:
```
+ xargs -0 -I element echo element
+ find /tmp -maxdepth 1 -name propably_not_existing -print0
```

PR #1246, opened 2021-10-26
unit: add new plugin to monitor nginx unit app server

kimheino:
This multigraph plugin does basic monitoring of unit's applications.
See: https://unit.nginx.org/


sumpfralle:
Great - thank you!

Issue #1247, opened 2021-11-04
ubiquiti - unifi_api spamming errors


dnnspaul:
![xjp2pvnlyibtwvff56xatkpy4_ssh](https://user-images.githubusercontent.com/4589085/140297021-a1e92fcd-b7af-4c80-bfd1-7f27326df198.png)

PR #1248, opened 2021-11-04
fix(issue #1247): ubiquiti - unifi_api spamming errors

dnnspaul:
Putting a fallback value so no error is spammed (Issue #1247)


sumpfralle:

What do you think: is zero (your current suggestion) or `U` (munin's magic key for "unknown" - to be emitted in case of retrieval problems) the most suitable fallback?

dnnspaul:
Hi Sumpfralle! Didn't know about the magic key. I'm sure the magic key will be the better fallback. Will change the PR within a minute...

sumpfralle:
Thank you!
I squashed and merged your commits now.

btw: *your pull requests* are your private area - thus you may feel free to squash/fixup/rebase the commits in your branch on your own

PR #1249, opened 2021-11-04
zfs list -t filesystem,volume to suppress snapshots

spacelama:
zfs list might list thousands of transient snapshots if `zpool
listsnapshots=on` and something like zfs-auto-snapshot is used.  It's
also not useful to graph them since they're unchanging in time, and
the parent volume is graphed anyway.  Limit zfs list to
"filesystem,volume" instead of "all" including snapshots


sumpfralle:
That sounds reasonable - thank you!

PR #1251, opened 2021-11-07
added self written proxmox_count plugin

quotengrote:
Hi,

i have writte an small plugin for proxmox.

It displays the count of VM and LXC, each running, stopped and total.

History: https://git.mgrote.net/mg/munin-plugins


plugins/proxmox/proxmox_count
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+#%# family=auto
+#%# capabilities=autoconf
+
+# Michael Grote
```
sumpfralle, line 5: Please use the *perlpod* documentation header style.

plugins/proxmox/proxmox_count
```diff
@@ -0,0 +1,63 @@
...
+# needs to be run as user: root AND group: root
+# [proxmox_count]
+# user root
+# group root
+
+# Pfade zu binaries
+wqm=$(command -v qm)
+wpct=$(command -v pct)
+# Berechne VMs
+total_vm=$($wqm list | sed 1d | wc -l)
```
sumpfralle, line 17: In order to avoid inconsistencies (e.g. a machine was started while the plugin is running), you should retrieve the `$wqm list` just once. Afterwards you can process this output multiple times.

plugins/proxmox/proxmox_count
```diff
@@ -0,0 +1,63 @@
...
+stopped_vm=$($wqm list | sed 1d | grep -c stopped )
+# Berechne LXCs
+total_lxc=$($wpct list | sed 1d | wc -l)
+running_lxc=$($wpct list | sed 1d | grep -c running )
+stopped_lxc=$($wpct list | sed 1d | grep -c stopped )
+# Berechne Gesamtwert
+total=$((total_vm + total_lxc))
+
+# wenn parameter = ...
+if [ "$1" = "autoconf" ]; then
```
sumpfralle, line 28: Please remove `autoconf`, if it is not used.

In this specific case I would recommend to verify the existence of `qm` and `pct`.

sumpfralle (commented):

Regarding the plugin name: how about `plugins/proxmox/proxmox_vm_count`?

plugins/proxmox/proxmox_count
quotengrote: good idea, changed it

plugins/proxmox/proxmox_count
quotengrote: changed it

plugins/proxmox/proxmox_count
quotengrote: do you have an example? i dont knwo what i need from: https://perldoc.perl.org/perlpod

sumpfralle: The [lvm_ plugin](https://github.com/munin-monitoring/contrib/blob/master/plugins/disk/lvm_) is quite a good (and minimal) example.
Or you could visit the [Plugin Gallery](https://gallery.munin-monitoring.org/), pick a random plugin and click at `Source Code` (at the bottom of the page).
Additionally the [plugin contribution hints](https://github.com/munin-monitoring/contrib/tree/master/plugins/#submit-a-new-plugin) could be a useful read, if you are in the mood.
Thanks!

plugins/proxmox/proxmox_count
```diff
@@ -0,0 +1,67 @@
...
+# Michael Grote
+# michael.grote ät posteo.de
+# Outputs the running, stopped and total counts of ProxMox VMs and LXCs.
+# needs to be run as user: root AND group: root
+# [proxmox_count]
+# user root
+# group root
+
+if ! command -v qm &> /dev/null
+then
```
sumpfralle, line 14: Please emit an error message and exit with a non-zero exit code, if you encounter problematic circumstances.

(`autoconf` checks would be the proper way to do it, but you may feel free to keep it as simple as it is now)

sumpfralle (commented):
Thanks for your improvements!

What do you think about the plugin filename? I would prefer something more specific (instead of the generic `_count` suffix).

Feel free to squash/rebase your changes into a single commit.

quotengrote:
Hi,

what do you think of the current name?

plugins/proxmox/proxmox_count
quotengrote: changed

plugins/proxmox/proxmox_count
quotengrote: changed it

quotengrote:
I have added the check within the autoconf check.

sumpfralle:
Thank you for bearing with all of my nitpicks.
I would be happy to merge it.

Will you squash the commits or should I do it?
Thanks!

quotengrote:
You can squash it.

No Problem, these were good suggestions.

sumpfralle:
Thanks!

PR #1254, opened 2021-11-14
kvm_io now in python3

AloisKlingler:
works on debian bullseye with this modification.
removed all codestyle issues, checked with "pycodestyle" on debian bullseye, besides two "line too long"

### DIFF
--- a/plugins/libvirt/kvm_io
+++ b/plugins/libvirt/kvm_io
@@ -50,8 +50,7 @@ def config(vm_names):
 graph_vlabel Bytes read(-)/written(+) per second
 graph_category virtualization
 graph_info This graph shows the block device I/O used of virtual machines
-graph_args --base 1024
-    """
+graph_args --base 1024"""
     print(base_config)
 
     for vm in vm_names:
--- a/plugins/libvirt/kvm_io
+++ b/plugins/libvirt/kvm_io
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 # vim: set fileencoding=utf-8
 #
@@ -11,11 +11,14 @@
 # parsed environment variables:
 # vmsuffix: part of vm name to be removed
 #
-#%# capabilities=autoconf
-#%# family=contrib
+# %# capabilities=autoconf
+# %# family=contrib
 
-import re, os, sys
+import re
+import os
+import sys
 from subprocess import Popen, PIPE
+
 
 def config(vm_names):
     ''' Print the plugin's config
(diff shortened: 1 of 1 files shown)


plugins/libvirt/kvm_io
```diff
@@ -11,12 +11,15 @@
 # parsed environment variables:
 # vmsuffix: part of vm name to be removed
 #
-#%# capabilities=autoconf
-#%# family=contrib
+# %# capabilities=autoconf
```
sumpfralle, line 14: The missing space here is crucial for matching these magic tokens :)

Please use the opportunity for turning the documentation into the perlpod format (see other plugins for examples).

plugins/libvirt/kvm_io
```diff
@@ -59,54 +63,59 @@ def clean_vm_name(vm_name):
...
 
+
 def fetch(vms):
     ''' Fetch values for a list of pids
     @param dictionary {kvm_pid: cleaned vm name}
     '''
     res = {}
     for pid in vms:
-        f = open("/proc/%s/io" % pid, "r")
+        f = open("/proc/%s/io" % pid.decode("utf-8"), "r")
```
sumpfralle, line 73: You may shorten this to `.decode()` - but this is just a matter of style.
In addition you may like to switch to `with open(...) as f:`?

(feel free to ignore these suggestions)

plugins/libvirt/kvm_io
```diff
@@ -59,54 +63,59 @@ def clean_vm_name(vm_name):
...
+
 if __name__ == "__main__":
     if len(sys.argv) > 1:
         if sys.argv[1] in ['autoconf', 'detect']:
             if detect_kvm():
-                print "yes"
+                print("yes")
             else:
-                print "no"
+                print("no")
```
sumpfralle, line 118: Please add a reason for the failure in brackets (e.g. `no (kvm executable missing)`).

sumpfralle (commented):
Thank you for taking the time to update this plugin!

Did you check style compliance with `flake8`?

plugins/libvirt/kvm_io
```diff
@@ -1,24 +1,49 @@
...
+
+import re
+import os
+import sys
 from subprocess import Popen, PIPE
 
+
 def config(vm_names):
-    ''' Print the plugin's config
+    ''' print(the plugin's config
```
sumpfralle, line 46: this one here seems to be a comment - not a command :)

plugins/libvirt/kvm_io
```diff
@@ -59,22 +85,22 @@ def clean_vm_name(vm_name):
...
-        f.close()
+        with open("/proc/%s/io" % pid.decode(), "r") as f:
+            for line in f.readlines():
+                if "read_bytes" in line:
+                    read = line.split()[1]
+                    print("%s_read.value %s" % (vms[pid], read))
+                if "write_bytes" in line:
+                    write = line.split()[1]
+                    print("%s_write.value %s" % (vms[pid], write))
+                    break
```
sumpfralle, line 102: Just for my curiosity: why should the processing stop after `write_bytes`?
Does this imply some specific sorting of the output? Maybe the original code included this erroneously?

sumpfralle (commented):
Thank you for your update!

Now I noticed two more nitpicks, which I overlooked before - sorry!

Afterwards it should be ready to be merged. Thanks!

sumpfralle:
btw: feel free to squash and force-push your commits. Otherwise I will do it, before merging ...

plugins/libvirt/kvm_io
AloisKlingler: oh ... %s did it, not me :D

plugins/libvirt/kvm_io
AloisKlingler: it looks like it is needed.
the output in my case without the break would be not valid:
./kvm_io
KVM_NGFW_AK_read.value 7456280576
KVM_NGFW_AK_write.value 2008285184
KVM_NGFW_AK_write.value 0

whereas with the break it looks as wanted:
./kvm_io
KVM_NGFW_AK_read.value 7456280576
KVM_NGFW_AK_write.value 2008289280

sumpfralle: OK.
In order to avoid the same question to be researched again in the future, it would be perfect to add a comment, explaining this (based on the nature of the input data stream).
But anyway: since you planned to just do the python3 migration, it is fine to leave this old stain here ...

sumpfralle:
>> squash and force-push your commits. Otherwise I will do it, before merging ...
> sorry, I do not know what you mean with this. so please go along :-)

*squash* refers to the act of combining multiple commits into one. This can be nice for history beatification.
(other contributors are not interested in the *private* back-and-forth of a discussion/proposal)

In your case you would run `git rebase -i HEAD~~` and then select the second commit to be "squashed" into the first one.
If you never used `git rebase -i` before: go for it now - it is a great magic tool for beautification.

Anyway: I will do this for you now. But you have your private branch at for trying it on your own :)

sumpfralle:

PR #1255, opened 2021-11-14
chrony_status: add a serverstats graph

kenyon:
This graph shows packet rates for the chronyd server.


sumpfralle:
Thanks for this improvement!

PR #1256, opened 2021-11-15
chrony_status: fix fetch output for serverstats graph

kenyon:
Was not expecting #1255 to be merged so quickly 🤪. Now the new serverstats graph has been fully tested and actually works 😉.

This change also doesn't needlessly convert all values to floating point numbers, instead provides the values to munin in the same format as they are originally output by chrony.

Also this adds some server stats found in chrony 4.0 (but still works with 3.4 and probably earlier).

### DIFF
--- a/plugins/chrony/chrony_status
+++ b/plugins/chrony/chrony_status
@@ -46,6 +46,24 @@ import os
 import subprocess
 import sys
 
+GRAPHS = [
+    'stratum',
+    'systime',
+    'delay',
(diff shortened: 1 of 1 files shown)


sumpfralle:
> Was not expecting #1255 to be merged so quickly [..]

You shall not question the well-meaning review! :)

The changesets look fine and simple (more specification, less code) - I guess, the plugin is quite well structured.

Thanks for your improvements!

PR #1257, opened 2021-11-15
add adguardhome plugin

brknkfr:
Display blocking stats, dns queries, client protocols and average
processing time of AdGuardHome

### DIFF
--- a/plugins/adguardhome/adguardhome_dns
+++ b/plugins/adguardhome/adguardhome_dns
@@ -1,0 +1,185 @@
+#!/bin/sh
+
+set -e
+
+: << =cut
+
+=head1 NAME
+
+adguardhome - Monitor dns queries
+
+=head1 APPLICABLE SYSTEMS
+
+AdGuardHome instances
+
+=head1 CONFIGURATION
+
+On a low traffic AdGuardHome instance you may want to lower 
+querylog_size_memory parameter in AdGuardHome configuration, otherwise 
+there won't be enough data for munin graphs.
+
+ log      - Querylog file of AdGuardHome
+ interval - Interval of munin cronjob in minutes
+ binary   - Path to AdGuardHome binary
+
+ [adguardhome_dns]
+ env.log /opt/AdGuardHome/data/querylog.json
+ env.interval 5
+ env.binary /opt/AdGuardHome/AdGuardHome
+
+=head1 AUTHOR
+
+Copyright (C) 2020 Sebastian L. (https://momou.ch)
+
+=head1 LICENSE
+
+GPLv2
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
+
+RECORDS="A AAAA ANY CNAME MX NS NXDOMAIN PTR SIG SOA SRV TXT"
+CLIENTS="dns doh dot"
+
+case $1 in
+
+    autoconf)
+        if [ -x "${binary:-/opt/AdGuardHome/AdGuardHome}" ]; then
+            if [ -z "$(command -v AdGuardHome)" ]; then
+                echo "no (AdGuardHome not found)"
+                exit 0
+            else
+                echo "yes"
+                exit 0
+            fi
+        else
+            echo "no (/opt/AdGuardHome/AdGuardHome not found)"
+            exit 0
+        fi
+        ;;
+
+   config)
+        echo "multigraph adguardhome_queries"
+        echo "graph_title AdGuardHome - DNS Queries"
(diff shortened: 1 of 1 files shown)


plugins/adguardhome/adguardhome
```diff
@@ -0,0 +1,180 @@
...
+        echo "safebrowsing.info Filtered for safe browsing"
+        echo "safebrowsing.min 0"
+        echo "multigraph adguardhome_avgproc"
+        echo "multigraph adguardhome_query_types"
+        echo "graph_title AdGuardHome - Query types"
+        echo "graph_info Query types made by AdGuardHome"
+        echo "graph_vlabel query types"
+        echo "graph_args -Y -A -l 0"
+        echo "graph_category network"
+        for KEY in A AAAA ANY CNAME MX NS NXDOMAIN PTR SIG SOA SRV TXT; do
```
sumpfralle, line 91: Since you are repeating this list of keys a few times, I would suggest to assign this list to a variable.
(same for the shorter list a few lines below)

sumpfralle (commented):
Thank you for proposing this new plugin!

In general I have two more questions:
1. Are you sure, that you want to squat the complete `adguardhome` namespace? Or maybe `adguardhome_dns` would be more suitable for this specific plugin?
2. You are using regular expressions for parsing the JSON data. How about simply using `jq`? This should ease the maintenance of the code.

brknkfr:
> Please take a look at my comments below.
> 
Thank you for the comments.

> In general I have two more questions:
> 
>     1. Are you sure, that you want to squat the complete `adguardhome` namespace? Or maybe `adguardhome_dns` would be more suitable for this specific plugin?
> 
I agree. It's probably better to call it adguardhome_dns.

>     2. You are using regular expressions for parsing the JSON data. How about simply using `jq`? This should ease the maintenance of the code.
>
The first version of the plugin was made with jq but it was slower than grepping for the values. That's why I don't use jq here.

plugins/adguardhome/adguardhome
```diff
@@ -0,0 +1,180 @@
...
+ #%# capabilities=autoconf
+
+=cut
+
+. "$MUNIN_LIBDIR/plugins/plugin.sh"
+
+case $1 in
+
+    autoconf)
+        if [ -x /opt/AdGuardHome/AdGuardHome ]; then
```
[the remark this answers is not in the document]
brknkfr: Good idea. The path is somewhat predictable for the moment as the installer script of adguardhome installs the binary to this path.

sumpfralle:
> The first version of the plugin was made with jq but it was slower than grepping for the values. That's why I don't use jq here.

Personally I would prefer maintainability over performance - but that's your decision ...

I am looking forward to your update!

brknkfr:
I left it to sed and grep which are at least three times faster than getting the values from the log with jq.

plugins/adguardhome/adguardhome_dns
```diff
@@ -0,0 +1,185 @@
...
+	# Reason 6 = FilteredInvalid
+	# Reason 7 = FilteredSafeSearch
+	# Reason 8 = FilteredBlockedService
+	queries=$(echo "$LOGDATA" | grep -o \"QT\": | wc -l)
+	blocked=$(echo "$LOGDATA" | grep -o '\"Reason\":[3,5,8]' | wc -l)
+	safesearch=$(echo "$LOGDATA" | grep -o \"Reason\":7 | wc -l)
+	safebrowsing=$(echo "$LOGDATA" | grep -o \"Reason\":4 | wc -l)
+	parental=$(echo "$LOGDATA" | grep -o \"Reason\":5 | wc -l)
+
+	for KEY in A AAAA ANY CNAME MX NS NXDOMAIN PTR SIG SOA SRV TXT; do
```
sumpfralle, line 154: Here you could use `$RECORDS`?

plugins/adguardhome/adguardhome_dns
```diff
@@ -0,0 +1,185 @@
...
+		export $KEY="$(echo "$LOGDATA" | grep -o \"QT\":\"$KEY\" | wc -l)"
+	done
+
+	for KEY in dns doh dot; do
```
sumpfralle, line 158: And here you could use `$CLIENTS`?

sumpfralle (commented):
Just two more nitpicks :)

plugins/adguardhome/adguardhome_dns
brknkfr: Ehm. I pushed too quicky, that's what I wanted.

plugins/adguardhome/adguardhome_dns
brknkfr: Ehm. I pushed too quicky, that's what I wanted.

sumpfralle:

PR #1258, opened 2021-11-16
munin-libvirtpy now in python3

AloisKlingler:
works on debian bullseye.
flake8 compatiblity ckecked.
I am not sure with the modified header, just copy&pasted - please feel free to adapt it to be compatible and as wanted - thanks. :-)

### DIFF
--- a/plugins/libvirt/munin-libvirtpy
+++ b/plugins/libvirt/munin-libvirtpy
@@ -1,57 +1,75 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
-# Revision 1.0  2008/05/16 - Steven Wagner
-# First functional release. Works for me.
-#
-# Revision 0.5  2008/05/01  - Julien Rottenberg
-# initial display of variables from libvirt
+"""
+=encoding utf8
 
-#python-libvirt is required
+=head1 NAME
+
+munin-libvirtpy - KVM Domain CPU Utilization
+
+
+=head1 CONFIGURATION
+
+Parsed environment variables:
+python-libvirt is required
+
+=head1 LICENSE
+
+GPLv3
+
+SPDX-License-Identifier: GPL-3.0-only
+
+
+=head1 AUTHORS
+
+Julien Rottenberg
+
+Steven Wagner
+
+
+=head1 MAGIC MARKERS
+
+ #%# capabilities=autoconf
+ #%# family=contrib
+
+=cut
+"""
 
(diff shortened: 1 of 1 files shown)


plugins/libvirt/munin-libvirtpy
```diff
@@ -1,57 +1,75 @@
...
+if conn is None:
+    print('Failed to open connection to the hypervisor')
     sys.exit(1)
 
 try:
     (model, memory, cpus, mhz, nodes, socket, cores, threads) = conn.getInfo()
-except:
-    print 'getInfo failed'
+except BaseException as error:
+    print('getInfo failed: {}'.format(error))
```
sumpfralle, line 49: Please write error messages to `stderr` (`file=sys.stderr`).

plugins/libvirt/munin-libvirtpy
```diff
@@ -1,57 +1,75 @@
...
 
-#print
-#print "KVM running on %d %s %d mhz CPUs w/ %d MB RAM." % (cpus, model, mhz, memory)
-#print
-
 ids = conn.listDomainsID()
-if ids == None or len(ids) == 0:
-    print 'No running domains found.'
+if ids is None or len(ids) == 0:
+    print('No running domains found.')
```
sumpfralle, line 54: `stderr`, please.

sumpfralle (commented):
Thank you for tackling the python3 migration for this plugin!

Below I added two minor comments for desirable changes (unrelated to the python3 migration).
I would be happy, if you could fix these too, while you are  involved with the plugin.
But I would also just merge your changes, as they are now (since they are perfectly fine for a python3 migration).
Just tell me, which path you choose ...

Thanks!

plugins/libvirt/munin-libvirtpy
AloisKlingler: done.
as I do all the changes in web, without having git installed, I cannot squash them as you suggested (or at least I cannot find a way in web to do it). please can you do again this? Thanks!

plugins/libvirt/munin-libvirtpy
AloisKlingler: done.
as I do all the changes in web, without having git installed, I cannot squash them as you suggested (or at least I cannot find a way in web to do it). please can you do again this? Thanks!

sumpfralle:
Thank you for taking care for this plugin!

PR #1259, opened 2021-11-19
munin_events: fix plugin having too much output

ap-wtioit:
in 56cd2c926ca99e396a0b6c2ebee76394bb4de195 `&> /dev/null` was converted
to posix in an invalid way, should have been `>/dev/null 2>&1`

this prevents the plugin dumping the logs to stderr at the end after printing the values

Info @wt-io-it


sumpfralle:
Thank you!

PR #1261, opened 2021-11-26
[sshd_log] Allow for configuration of warning and critical levels

HaseHarald:
Allows for configuring either general or specific warning or critical
levels, or any combination of that.

This is backwards compatible, since it doesn't change a thing to
the output, unless specific variables are specified.


sumpfralle:
Thank you for extending the plugin!

PR #1263, opened 2021-11-30
bind95_: fix value keys starting with numbers

ap-wtioit:
md5_hex yields a value starting with a number in 72% (10/16) of all
cases. for those keys to be usable with munin we need to prefix them
so they do not start with a number.

2d92ac162405d70f12fa875e2a0c4f9a.label becomes
_2d92ac162405d70f12fa875e2a0c4f9a.label

dbb781de320d50f9fabcd67f93026419.label stays as it is

Info: @wt-io-it

### DIFF
--- a/plugins/bind/bind95_
+++ b/plugins/bind/bind95_
@@ -2,27 +2,58 @@
 #
 # Copyright Jean-Samuel Reynaud <js.reynaud@free.fr>
 # Licensed under GPL v2
-#
-# We use this script to produce graph with munin for dns requests
-# This script must have his name start with bind95_
-# bind95_ : Global bind statistic
-# bind95_test.com : Bind statistic for test.com zone (no view)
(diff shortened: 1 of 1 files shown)


ap-wtioit:
Sorry for misreporting, this was not the problem with the plugin reporting lots of `-nan`

sumpfralle:
Thank you for taking a closer look at the plugin.

Your (retracted) proposal also includes a reformatting of the plugin's documentation header.
Would you mind to remove the key-related changes and just submit the improved documentation?
Otherwise I would do this (based on your commit).

Thanks!

ap-wtioit:
@sumpfralle i'm testing a new really fixed version (reporting 0 for all declared values not yet hit in the dns) on our servers for now. this includes the formatting. if it works on our servers i will open a new pull request.

PR #1265, opened 2021-12-01
bind95_: report 0 values for not found lines

ap-wtioit:
This prevents `bind95_` from displaying `-nan` for counters not yet present in stats file (not requests yet) and instead displays them as 0. It makes sure that there is a value for all counters defined in `bind95_ config`.

Info @wt-io-it


ap-wtioit:
@sumpfralle this is the promised pull request from #1263

sumpfralle:
Thank you!

PR #1266, opened 2021-12-02
certificate_file_expiry: label "50" instead of "5e+01"

kimheino:
Scale values were "4e+01, 5e+01, ...". Using linear scale and "48, 50, 52, ..." is much more readable.

Also use shorter vlabel, old one is too long.

### DIFF
--- a/plugins/ssl/certificate_file_expiry
+++ b/plugins/ssl/certificate_file_expiry
@@ -55,8 +55,8 @@ GPLv2
 
 if [ "$1" = "config" ] ; then
   echo "graph_title Certificate validity"
-  echo "graph_args --logarithmic --base 1000"
-  echo "graph_vlabel certificate validity in days"
+  echo "graph_args --base 1000"
+  echo "graph_vlabel days"
   echo "graph_category security"
 fi
 


sumpfralle:
Thank you for cleaning this up!

ap-wtioit:
@kimheino the logarithmic scale was intended to display values of 10years as well as values of a few days within the same graph.

kimheino:
I see... I'm using this plugin for real (not self-signed) certificates which usually have less that two years, in my case 90 days. I assume this is the most common case?

In your case warning level is now close to zero axis but that's why you have warning emails enabled, don't you? You should tune warning limit if you need more time to replace certificates.

ap-wtioit:
we use it for VPN certs (self-signed, CA has 10years lifetime, hosts have 1year, crl has a few months, still real certs) as well as HTTPS certs (90days) we have mails enabled, but still i would like to see what the difference is, and logarithmic worked great for us. i wrote this plugin to monitor certs that are not available on SSL ports (the crl.pem and CA cert of openvpn): https://github.com/munin-monitoring/contrib/pull/962#issuecomment-513661564

i will test this and report my results and get back, currently i would prefer a config option that brings back logarithmic for us.

PR #1267, opened 2021-12-03
certificate_file_expiry: optional logarithmic view for long lived certs

ap-wtioit:
logarithmic was removed in d885a314ade7c1287237d61f4d2b1f24db509533
but still has it's benefits for openvpn CA and CRL certs. when enabling
logarithmic we keep the SI units now instead of rrdtools default
scientific notation.

This enables using logarithmic graph as an option again that was removed in #1266.

Info @wt-io-it

### DIFF
--- a/plugins/ssl/certificate_file_expiry
+++ b/plugins/ssl/certificate_file_expiry
@@ -13,6 +13,7 @@ For openvpn ca.crt and crl.pem
  [certificate_file_expiry]
  user root
  env.CERTS crl:/etc/openvpn/easy-rsa/keys/crl.pem x509:/etc/openvpn/easy-rsa/keys/ca.crt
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thanks for taking a look at this!

Maybe it would be sufficient to enforce the SI notation and keep the logarithmic visualization enabled by default (as it was before)?

@kimheino / @ap-wtioit: what do you think?

ap-wtioit:
i would think `LOGARITHMIC=yes` as default that would keep the old behavior and still give the user the option to disable the logarithmic graph view so be the best option.

kimheino:
"yes" as default works for me. I'll probably change it to "no" in my config.

sumpfralle:
That was quick - thanks!

PR #1269, opened 2021-12-09
[minecraft-users-ram] Fix several bugs in plugin minecraft-users-ram_

HaseHarald:
Fixes:
- Make playercount work on non english systems
- Avoid getting an invalid PID if not run with sufficient permissions
- Fix getting CPU usage

Enhance:
- Clearify that CPU usage is in percent


sumpfralle:
Thank you!

Just some notes:
* for `grep` patterns starting with a hyphen, I would suggest to use the *safe* approach: `grep ... -- PATTERN`
    * `--` indicates the end of keyword arguments)
* `LANG=en LC_LANG=C ...` feels redundant to me - but I do not know these details well
* I think, `:` (colon) does not require escaping for `grep`

PR #1270, opened 2021-12-10
[gitlab_statistics] Fix urllib request decoding

codeurimpulsif:
Fix the error `AttributeError: module 'urllib' has no attribute 'request'` and then the error `TypeError: the JSON object must be str, bytes or bytearray, not HTTPResponse` by reading and decoding the urllib request.


sumpfralle:
Thank you!

PR #1272, opened 2021-12-11
added an Ignore Option to zfs_list

quotengrote:
Hi,

i have added an option to ignore certain datasets for `zfs_list`.
And updated the Header based on the Old Information within + new Variable.

I think it is useful, because in my case i have many very short lived datasets, that get destroyed and clutter the GUI.

Wishes
mg

### DIFF
--- a/plugins/zfs/zfs_list
+++ b/plugins/zfs/zfs_list
@@ -46,7 +46,10 @@ if [ "$1" = "suggest" ]; then
 	exit 0
 fi
 
+# set standard value
 ignore_datasets_pattern=${ignore_datasets_pattern:-}
+
+# set search term, based on if the searh pattern is set
 if [ -n "$ignore_datasets_pattern" ]; then
   search_term=$(zfs list -Hp -t filesystem,volume | grep -v "$ignore_datasets_pattern" | awk '{print $1}')
 else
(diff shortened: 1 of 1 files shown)


plugins/zfs/zfs_list
```diff
@@ -1,21 +1,37 @@
...
+zfs_list - Plugin to monitor ZFS Filesystems
+
+This is an extension of the zfs_fs plugin
+modified as a multigraph to graph all zfs
+filesystems it can find
+
+=head1 CONFIGURATION
+plugin config:
+[zfs_list]
+env.ignore_datasets_string <Datasets to ignore; String for grep>
```
sumpfralle, line 15: Maybe there is a way to improve the name of this settings?
I would recommend to replace `string` with `pattern` (in order to refer to a regular expression).
Regarding `datasets`: I cannot judge, whether this is the proper term.

plugins/zfs/zfs_list
```diff
@@ -1,21 +1,37 @@
...
+These magic markers are used by munin-node-configure when installing
+munin-node.
+=end comment
+  #%# family=auto
+=cut
+
+EOF
+
+
 . "$MUNIN_LIBDIR/plugins/plugin.sh"
```
sumpfralle, line 36: I would recommend to add something like `ignore_datasets_string=${ignore_datasets_string:-}` here. This would serve as an interface description and help users to identify the configuration variables and their default settings.

sumpfralle (commented):
This looks like a good change.

Thanks for preparing the perldoc header!

plugins/zfs/zfs_list
quotengrote: changed it

plugins/zfs/zfs_list
quotengrote: added

quotengrote:
Hi,

i have added your requested changes.

sumpfralle:
I squashed your commits into one and merged it. Thank you!

PR #1273, opened 2021-12-11
[btrfs_device_usage] fix division by zero when btrfs pool is unhealthy

HaseHarald:
During certain situations, a device in the btrfs pool can show a total
capacity of 0 bytes. This is aspecially true when replacing or removing a
failed disk.

This fix stops the plugin from crashing in that situation but just report
the devices percentage as unknown (U). That way other devices in the pool
stil can be monitored.


sumpfralle:
Thank you!

PR #1275, opened 2021-12-20
quake/qstat: enhancements

brknkfr:
- dash
- bots
- configurable binary


sumpfralle:
Thank you for cleaning up the plugin!

PR #1276, opened 2021-12-20
murmur-stats: Enhancements

brknkfr:
- python3 compatibility
- multigraph
- added muted/registered/unregistered users

### DIFF
--- a/plugins/mumble/murmur-stats
+++ b/plugins/mumble/murmur-stats
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # -*- coding: utf-8
 # Python Plugin for Munin
 # Copyright (C) 2010 Natenom (Natenom@googlemail.com)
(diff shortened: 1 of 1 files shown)


plugins/mumble/murmur-stats
```diff
@@ -14,31 +14,85 @@ iceslice='/usr/share/slice/Murmur.ice'
...
+  ice.destroy()
+  sys.exit(1)
+
+users = server.getUsers()
+users_muted = 0
+users_unregistered = 0
+users_registered = 0
+
+for key in list(users.keys()):
+  if users[key].userid == -1:
```
sumpfralle, line 72: Maybe the next lines could be easier to understand with a list comprehension?
```python
users_unregistered = len(user for user in users.values() if user.userid == -1)
users_registered = len(user for user in users.values() if user.userid > 0)
users_muted = len(user for user in users.values() if user.mute or user.selfMute or self.suppress)
```

plugins/mumble/murmur-stats
```diff
@@ -14,31 +14,85 @@ iceslice='/usr/share/slice/Murmur.ice'
...
+    users_unregistered += 1
+
+  if users[key].userid > 0:
```
sumpfralle, line 75: A user ID with the value zero is not allowed?

plugins/mumble/murmur-stats
```diff
@@ -14,31 +14,85 @@ iceslice='/usr/share/slice/Murmur.ice'
...
+
+  if users[key].selfMute:
+    users_muted += 1
+
+  if users[key].suppress:
+    users_muted += 1
+
+print("multigraph murmur_stats_users")
+print("users_total.value %i" % (len(users)))
+print("users_muted.value %i" % (users_muted))
```
sumpfralle, line 89: Just for the sake of compeleteness: you are adding braces around `users_muted`, thus it *looks* like a tuple (as it is recommeneded for arguments of `%`), but it is still just a number :)

I would recommend one of the following approaches:
* A) remove the braces for clarity
* B) add a trailing comma within the braces in order to turn it into a tuple expression
* C) use one of the other (less suprising) string formatting methods (`"".format()` or `f""`)

sumpfralle (commented):
Thank you for taking care for the python3-migration of this plugin!

plugins/mumble/murmur-stats
brknkfr: I'll have to add a list() for this.

plugins/mumble/murmur-stats
brknkfr: Oh, good catch. SuperUser normally has userid 0.

sumpfralle:

PR #1277, opened 2021-12-23
Add zswap Plugin

quotengrote:
Hi,

i have added a new Plugin for zswap metrics.

One Question: is it possible make the values axis logaritmic? I am asking because all values range from 0-99999 usually, but one is almost ever over 40 Million, so it would be nice to see the small lines better.

Wishes
mg


plugins/system/zswap
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
```
sumpfralle, line 1: I think, you can use `/bin/sh` for this plugin?
(just removing the `function` keywords)

plugins/system/zswap
```diff
@@ -0,0 +1,93 @@
...
+
+=head1 LICENSE
+
+GPLv3 or later
+
+SPDX-License-Identifier: GPL-3.0-or-later
+
+=head1 MAGIC MARKERS
+
+  #%# family=auto
```
sumpfralle, line 30: Please add `#%# capabilities=autoconf`.

plugins/system/zswap
```diff
@@ -0,0 +1,93 @@
...
+  exit 0
+fi
+
+if [ "$1" = "config" ]; then
+  # setze label
+  print_labels
+  exit 0
+fi
+
+echo duplicate_entry.value $(cat "/sys/kernel/debug/zswap/duplicate_entry")
```
sumpfralle, line 84: Maybe store the base directory in a variable in order to reduce visual noise? (`SYSFS_ZSWAP_DIR=/sys/kernel/debug/zswap`)

sumpfralle (commented):

plugins/system/zswap
quotengrote: changed

plugins/system/zswap
quotengrote: changed

plugins/system/zswap
quotengrote: Didn't work

sumpfralle: What exactly failed to work?
Maybe you did not add `()` after the function name? (this is optional in bash)
Maybe run `shellcheck -s dash`? Its output may be easier to understand, than the shell's syntax error messages.

We strive to keep the number of dependencies of munin plugins minimal (as long as it does not hurt maintainability). Something like `bash` renders plugins unusable on platforms like OpenWrt.
Thank you!

quotengrote: that was it...

sumpfralle:
I merged it now. Thank you for contributing this new plugin!

PR #1279, opened 2021-12-29
[ssl-certificate-expiry] add env.skip_cert_hashes to skip certificates

shtrom:
This fix allows to skip validity checks on any intermediate certificate
which subject hash appears in the list.

Let's Encrypt certificates are cross-signed by both DST Root CA X3 and
ISRG Root X1. DST Root CA X3 has expired on 2021-09-30 [0], which causes
the plugin to incorrectly report some full-chain LE certificates as
expired.

[0] https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/

### DIFF
--- a/plugins/ssl/ssl-certificate-expiry
+++ b/plugins/ssl/ssl-certificate-expiry
@@ -1,6 +1,6 @@
 #!/bin/sh -u
 # -*- sh -*-
-# shellcheck disable=SC2039
+# shellcheck disable=SC2039,SC3043
 
 : << =cut
 
@@ -70,7 +70,7 @@ uncached updates after the cache file is older than an hour.
 
 =cut
 
-# shellcheck disable=SC1090
+# shellcheck disable=SC1091
 . "${MUNIN_LIBDIR}/plugins/plugin.sh"
 
 if [ "${MUNIN_DEBUG:-0}" = 1 ]; then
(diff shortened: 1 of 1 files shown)


plugins/ssl/ssl-certificate-expiry
```diff
@@ -97,7 +98,13 @@ parse_valid_days_from_certificate() {
     input_data=$(cat)
 
     if echo "$input_data" | grep -q -- "-----BEGIN CERTIFICATE-----"; then
-        valid_until_string=$(echo "$input_data" | openssl x509 -noout -enddate \
+        cert_data=$(echo "$input_data" | openssl x509 -noout -subject_hash -enddate)
+
+        # Skip certificate if its hash is in env.skip_cert_hashes
+        hash="$(echo "${cert_data}" | head -n 1)"
+        echo "${skip_cert_hashes:-}" | grep -q "${hash}" && return
```
sumpfralle, line 105: How about `grep -iqwF` instead?

At least `-i` feels reasonable to me - the rest is just obsession with details. Pick you path :)

plugins/ssl/ssl-certificate-expiry
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh -u
 # -*- sh -*-
-# shellcheck disable=SC2039
+# shellcheck disable=SC2039,SC3043
```
sumpfralle, line 3: You do not need to add this exception, since the CI tests for `shellcheck` are executed with `-s dash`.

sumpfralle (commented):
This X3 certificate is really causing a *lot* of work :(

Thanks for adding this option!

shtrom:
@sumpfralle fixed!

PR #1280, opened 2021-12-29
[whois] Allow to specificy whois server per domain

shtrom:
Also, clean up some old code that pre-dates multi-domain handling


sumpfralle:
Thank you for cleaning up and extending this plugin!

PR #1282, opened 2021-12-31
Get statistics only for running containers. Fix #1281

### LINKED ISSUE #1281: docker_: Plugin fails for cpu, memory and networking when a container is in restart loop
The traces looks like:
```
# munin-run docker_cpu
...
  File "/etc/munin/plugins/docker_cpu", line 474, in <module>
    main()
  File "/etc/munin/plugins/docker_cpu", line 433, in main
    print_containers_cpu(client)
  File "/etc/munin/plugins/docker_cpu", line 268, in print_containers_cpu
    system_delta = (float(stats["cpu_stats"]["system_cpu_usage"])
KeyError: 'system_cpu_usage'
```

```
# munin-run docker_memory
...
  File "/etc/munin/plugins/docker_memory", line 474, in <module>
    main()
  File "/etc/munin/plugins/docker_memory", line 447, in main
    print_containers_memory(client)
  File "/etc/munin/plugins/docker_memory", line 278, in print_containers_memory
    print(container.name + '.value', stats['memory_stats']['stats']['total_rss'])
KeyError: 'stats'
```

```
# munin-run docker_network
...
  File "/etc/munin/plugins/docker_network", line 474, in <module>
    main()
  File "/etc/munin/plugins/docker_network", line 470, in main
    print_containers_network(client)
  File "/etc/munin/plugins/docker_network", line 286, in print_containers_network
    for data in stats['networks'].values():
KeyError: 'networks'
```

NITEMAN:
Stats only has sense for running container anyway


sumpfralle:
Do you think, that it could make sense to apply the same condition to `all_containers` in general?

(btw.: maybe a list comprehension could be more readable, than `filter`?)

NITEMAN:
I think the condition only has sense in stats, you'll probably want status of `all_containers`

Regarding list comprehension... I don't use to code python so I just googled, landed on filter, tested and saw it works ;)... Let me see if I can write the same using list comprehension

NITEMAN:
@sumpfralle converted to a list comprehension and tested.

BTW: It seems that [all_containers should already be doing the filtering](https://github.com/munin-monitoring/contrib/blob/master/plugins/docker/docker_#L148)... but my brand new python skills gime me no idea of what's going wrong (maybe a cache?)

PR #1284, opened 2022-01-05
Snmp  fixes

shtrom:
Small fixes in a couple of SNMP plugins:
* unify titles and category in snmp__df_ram
* add support for multiple instances of snmp__df_combined

### DIFF
--- a/plugins/router/snmp__mikrotik
+++ b/plugins/router/snmp__mikrotik
@@ -88,8 +88,8 @@ if ($ARGV[0] and $ARGV[0] eq "config") {
     if (defined $response) {
         print "multigraph flash\n";
         print "graph_args --base 1024 -l 0 --vertical-label Bytes --upper-limit " . ($response->{$sysFlashTotalOID} * 1024) . "\n";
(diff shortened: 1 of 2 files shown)


sumpfralle:
Thank you! This proposal for expanding the usage pattern for the plugin looks good to me.

Personally I would probably prefer something like `graph_name` instead of `script_name`. But besides this, I have no further comments or suggestions.

Did you mark it as "draft" on purpose?

shtrom:
> Did you mark it as "draft" on purpose?

Yes, I want to test it a bit more, but didn't want to forget about the PR (:

sumpfralle:
> Yes, I want to test it a bit more, but didn't want to forget about the PR

Ah - good. Just ping me, when you are ready.

Thanks!

shtrom:
@sumpfralle I think this is good for review. This has been running locally for me for the last two months without major hiccups. The `stackedMax` approach is not great and not foolproof, but that the best we can do (that I can think of).

sumpfralle:
Great, thank you!

PR #1286, opened 2022-01-08
[exodus_] Add plugin to monitore number of applications, reports and trackers

codeurimpulsif:
## Goal

This plugin monitore number of applications, reports and trackers of an [Exodus](https://github.com/Exodus-Privacy/exodus/) instance.

## Informations

This plugin is based on [the Exodus API documentation](https://github.com/Exodus-Privacy/exodus/blob/v1/doc/api.md) and can be used with any Exodus instance starting from [version 1.25.0](https://github.com/Exodus-Privacy/exodus/releases/tag/v1.25.0).

### DIFF
--- a/plugins/exodus/exodus_
+++ b/plugins/exodus/exodus_
@@ -38,6 +38,10 @@ You need to create a file named exodus placed in the directory
 
 Codimp <contact@lithio.fr>
 
+=head1 LICENSE
+
+GPLv3
+
 =head1 MAGIC MARKERS
 
  #%# family=manual
--- a/plugins/exodus/exodus_
+++ b/plugins/exodus/exodus_
@@ -71,7 +71,7 @@ def main():
         mode = sys.argv[1]
     except IndexError:
         mode = ""
-    wildcard = sys.argv[0].split("exodus_")[1].split("_")[0]
+    wildcard = sys.argv[0].split("exodus_")[1]
 
     if mode == "suggest":
         print("applications")
--- a/plugins/exodus/exodus_
+++ b/plugins/exodus/exodus_
@@ -58,12 +58,12 @@ import sys
 import requests
 
 
-def print_count(type):
-    endpoint = os.getenv('exodus_url') + '/api/' + type + '/count'
+def print_count(element):
+    endpoint = os.getenv('exodus_url') + '/api/' + element + '/count'
     headers = {"Authorization": "Token " + os.getenv('exodus_token')}
     count = requests.post(endpoint, headers=headers)
     if count.status_code == 200:
-        print(type + '.value', count.json()['count'])
+        print(element + '.value', count.json()['count'])
 
 
 def main():
--- a/plugins/exodus/exodus_
+++ b/plugins/exodus/exodus_
@@ -52,11 +52,11 @@ import requests
 
 
 def print_count(element):
-    endpoint = os.getenv('exodus_url') + '/api/' + element + '/count'
-    headers = {"Authorization": "Token " + os.getenv('exodus_token')}
+    endpoint = os.getenv("exodus_url") + "/api/" + element + "/count"
+    headers = {"Authorization": "Token " + os.getenv("exodus_token")}
     count = requests.post(endpoint, headers=headers)
     if count.status_code == 200:
-        print(element + '.value', count.json()['count'])
+        print(element + ".value", count.json()["count"])
 
 
 def main():
@@ -79,7 +79,7 @@ def main():
             print("graph_category exodus")
             print("applications.label Applications")
         else:
-            print_count('applications')
+            print_count("applications")
     elif wildcard == "reports":
         if mode == "config":
             print("graph_title Exodus reports")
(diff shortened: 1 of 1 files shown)


plugins/exodus/exodus_
```diff
@@ -0,0 +1,108 @@
...
+    if count.status_code == 200:
+        print(type + '.value', count.json()['count'])
+
+
+def main():
+    try:
+        mode = sys.argv[1]
+    except IndexError:
+        mode = ""
+    wildcard = sys.argv[0].split("exodus_")[1].split("_")[0]
```
sumpfralle, line 74: What is the purpose of the final `split`?

plugins/exodus/exodus_
```diff
@@ -0,0 +1,108 @@
...
+
+    if mode == "suggest":
```
sumpfralle, line 76: You should `exit` after emitting the `suggest` items.

plugins/exodus/exodus_
```diff
@@ -0,0 +1,108 @@
...
+
+=cut
+"""
+
+import os
+import sys
+import requests
+
+
+def print_count(type):
```
sumpfralle, line 61: You should avoid overriding names of `builtins` like `type`.

plugins/exodus/exodus_
```diff
@@ -0,0 +1,108 @@
...
+
+ #%# family=manual
+ #%# capabilities=suggest
+
+=cut
+"""
+
+import os
+import sys
+import requests
```
sumpfralle, line 58: `request` is not part of the standard library.
You could replace it with `urllib` (part of the standard library), e.g.
```python
request = urllib.request.Request(endpoint, method="POST", headers=headers)
with urllib.request.urlopen(request) as connection:
    count = json.loads(connection.read())["count"]
```
Or you should use `try/except` for the import and emit an error message to `stderr`, if the module is missing.
It depends on how kind you want to be towards the users :)

plugins/exodus/exodus_
```diff
@@ -0,0 +1,108 @@
...
+    ln -s /usr/share/munin/plugins/exodus_ /etc/munin/plugins/exodus_reports
+    ln -s /usr/share/munin/plugins/exodus_ /etc/munin/plugins/exodus_trackers
+
+=back
+
+After the installation you need to restart your munin-node:
+
+=over 2
+
+    systemctl restart munin-node
```
sumpfralle, line 27: I would recommend to omit this systemd-specific command. A generic hint  (e.g. "restart the munin-node service") should be sufficient and works everywhere - now and in the future.

plugins/exodus/exodus_
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3
+"""
+=head1 NAME
+
+exodus_ - Exodus wildcard-plugin to monitor an L<Exodus|https://github.com/Exodus-Privacy/exodus/>
+instance.
+
+This wildcard plugin provides at the moment only the suffixes C<applications>, C<reports>
```
sumpfralle, line 8: Maybe remove `only` here? Or state, which other parts you consider to be missing.

plugins/exodus/exodus_
```diff
@@ -0,0 +1,108 @@
...
+
+=over 2
+
+    [exodus_*]
+    env.exodus_url https://reports.exodus-privacy.eu.org
+    env.exodus_token your-api-token
+
+=back
+
+=head1 AUTHORS
```
sumpfralle, line 44: Please add a *License* section, thanks!

plugins/exodus/exodus_
```diff
@@ -0,0 +1,108 @@
...
+        else:
+            print_count('reports')
+    elif wildcard == "trackers":
+        if mode == "config":
+            print("graph_title Exodus trackers")
+            print("graph_vlabel trackers")
+            print("graph_category exodus")
+            print("trackers.label Trackers")
+        else:
+            print_count('trackers')
```
sumpfralle, line 104: Maybe run `black` in order to unify the quoting style (single or double quotes) in the script?

sumpfralle (commented):

Regarding the name: do you think, that the plugin should cover the whole `exodus_` namespace? Or maybe something like `exodus_api_` would be more specific?
(I do not know the exodus application, thus I cannot compare the scope of your plugin with the full scope of exodus)

Please take a look at my comments. Most are just style hints. Thank you!

plugins/exodus/exodus_
codeurimpulsif: my bad, it was from another plugin for another use, I remove it

codeurimpulsif:
> Regarding the name: do you think, that the plugin should cover the whole `exodus_` namespace? Or maybe something like `exodus_api_` would be more specific? (I do not know the exodus application, thus I cannot compare the scope of your plugin with the full scope of exodus)

Yes, I think it should cover the whole namespace (I talked about that with one of the main developers of Exodus)

sumpfralle:
Thank you for going through my nitpicks and for contributing this new plugin!

PR #1292, opened 2022-02-08
Update apt_ubuntu to display integers instead of floats

cookieguru:
Half a package can't be updated, so make the graph display integers instead of floats that always end in `.00`

### DIFF
--- a/plugins/ubuntu/apt_ubuntu
+++ b/plugins/ubuntu/apt_ubuntu
@@ -77,6 +77,7 @@ def config():
     print('graph_category security')
     print('graph_title %s' % (title))
     print('graph_vlabel %s' % (vlabel))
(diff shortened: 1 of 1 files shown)


sumpfralle:

> Half a package can't be updated, so make the graph display integers instead of floats that always end in .00

I do not use this plugin, but I could imagine, that the `printf` setting also applies to the averages displayed below the graph.
Thus I think, the detailed number format could be useful.

What do you think?

cookieguru:
I don't quite understand the question but here's the graph in use with the table beneath the graphs

image

sumpfralle:
Thank you for providing the screenshot, which proved my assumption to be wrong.

Thus I merged your proposed improvement - thank you!

PR #1293, opened 2022-02-09
certificate_file_expiry: enable monitoring inline openvpn certificates

ap-wtioit:
This enables monitoring certificates for openvpn where the client has the certificates defined directly in the client.conf
as described in the openvpn manual https://community.openvpn.net/openvpn/wiki/Openvpn23ManPage#lbAV

Also i restored the default warning and critical level, i accidentially removed in https://github.com/munin-monitoring/contrib/commit/76170d27453f4b2df4e42a6445a6f2be5589213a

Info @wt-io-it

### DIFF
--- a/plugins/ssl/certificate_file_expiry
+++ b/plugins/ssl/certificate_file_expiry
@@ -55,6 +55,9 @@ env.LOGARITHMIC "yes" enables the logarithmic display of values which is useful 
 long lived in respect to the warning level. e.g. a ca.crt that is valid for 10 years together with a crl.pem that is
 valid for only a few months combined with warning levels of 5 days. default is "yes" to disable set it to "no".
 
+env.IGNORE_UNEXPANDED_PATTERNS "yes" ignores patterns that did not expand to any files. this is useful to define one
+config that handles multiple types of certs where only one pattern is used. default is "no".
+
 =head1 Dependencies
 
 Dependencies: openssl
@@ -72,6 +75,7 @@ GPLv2
 . "$MUNIN_LIBDIR/plugins/plugin.sh"
 
 LOGARITHMIC=${LOGARITHMIC:-yes}
+IGNORE_UNEXPANDED_PATTERNS=${IGNORE_UNEXPANDED_PATTERNS:-no}
 
 if [ "$1" = "config" ] ; then
   echo "graph_title Certificate validity"
@@ -127,6 +131,16 @@ for cert in ${CERTS}; do
   cert_type=${cert%:*}
   cert_pattern=${cert#*:}
   for cert_file in $cert_pattern; do
+    # note: if file contains a * (e.g. /etc/letsencrypt/live/*/cert.pem) it might be an unexpanded pattern
+    # to supress errors see IGNORE_UNEXPANDED_PATTERNS above
+    # shellcheck disable=SC2063
+    if [ "$IGNORE_UNEXPANDED_PATTERNS" = "yes" ] \
+      && [ "$cert_file" = "$cert_pattern" ] \
+      && ! [ -e "$cert_file" ] \
+      && echo "$cert_file" | grep -q "*" ; then
+        # skip unexpanded patterns when IGNORE_UNEXPANDED_PATTERNS is set to yes
+        continue
+    fi
     if [ "$cert_type" = "openvpn_inline" ] ; then
       for type in "ca" "cert"; do
         cert_name=$(clean_fieldname "$cert_file-$type")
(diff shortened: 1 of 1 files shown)


plugins/ssl/certificate_file_expiry
```diff
@@ -72,23 +87,70 @@ if [ "$1" = "config" ] ; then
...
+  warning="5:"
+fi
+if [ "${critical:-}" = "" ] ; then
+  # as default when certificate is only valid for 1 day or less emit a critical
+  critical="1:"
+fi
+
 now=$(date +%s)
+get_validity() {
+  openssl_type=$1
```
sumpfralle, line 101: I would suggest to use `local` here.
(please note, that you need to run `shellcheck` with `-s dash` as a parameter in order to receive proper hints)

plugins/ssl/certificate_file_expiry
```diff
@@ -72,23 +87,70 @@ if [ "$1" = "config" ] ; then
...
+  file=$2
+  if [ "$file" != "--" ] ; then
```
sumpfralle, line 103: A single dash is a more common convention for specifying "read from stdin". Maybe this would be easier to read?

plugins/ssl/certificate_file_expiry
```diff
@@ -72,23 +87,70 @@ if [ "$1" = "config" ] ; then
...
+    validity=$(/usr/bin/openssl "$openssl_type" -text -noout -in "$file" | grep -E '(Next Update|Not After)')
+  else
+    # when file is set to -- read from stdin
+    validity=$(/usr/bin/openssl "$openssl_type" -text -noout | grep -E '(Next Update|Not After)')
+  fi
+  validity=${validity#*:}
```
sumpfralle, line 109: I think, I would use multiple variables here. It is usually a rich source of confusion to use the same variable with different content along the flow of the code ...

plugins/ssl/certificate_file_expiry
```diff
@@ -72,23 +87,70 @@ if [ "$1" = "config" ] ; then
...
-    elif [ "$1" = "" ] ; then
-      validity=$(/usr/bin/openssl "$cert_type" -text -noout -in "$cert_file" | grep -E '(Next Update|Not After)')
-      validity=${validity#*:}
-      validity=$(date --date="$validity" +%s)
-      validity=$((validity - now))
-      validity=$(echo "$validity" | awk '{ print ($1 / 86400) }')
-      echo "${cert_name}.value $validity"
+    if [ "$IGNORE_UNEXPANDED_PATTERNS" = "yes" ] \
+      && [ "$cert_file" = "$cert_pattern" ] \
+      && echo "$cert_file" | grep -q "*" ; then
```
sumpfralle, line 134: Are you sure, that you want to `grep` for `*`?
Maybe you meant the dot instead ("any character")? Or maybe simple use `[ -n "$cert_file" ]` in this case?

sumpfralle (commented):
Thank you for your contributions!

plugins/ssl/certificate_file_expiry
ap-wtioit: thanks for the hint, yes my shellcheck is still configured for posix as my IDE currently doesn't support configuring dash for sh shebang: https://github.com/BashSupport-Pro/bashsupport-pro/issues/91

plugins/ssl/certificate_file_expiry
ap-wtioit: yes i'm sure (as sure as a i can be, say 95%), grepping for `*` seemed the safest and most compatible way to check if the `cert_file` variable contains an unexpanded pattern e.g. `/etc/letsencrypt/live/*/cert.pem` which is the core of the `IGNORE_UNEXPANDED_PATTERNS` flag.

this enables us to use one config for multiple nodes that have not all files present at all times (e.g. in a pre prepared docker image where files are mapped into the container or not):
```
[certificate_file_expiry]
user root
env.CERTS crl:/etc/monitor_certs/*/crl.pem x509:/etc/monitor_certs/*/*.crt openvpn_inline:/etc/monitor_certs/*/client.conf x509:/etc/monitor_certs/*/cert.pem
env.IGNORE_UNEXPANDED_PATTERNS yes
```

i added the flag for the behaviour without the flag to stay the same (showing errors) in case somebody depends on those errors (although that is totally not adviced)

i added a few lines of comments to make it clear the `grep -q "*"` is not an accident

ap-wtioit:
@sumpfralle thanks for the feedback, i fixed the stuff you mentioned and explained why i think the `grep -q "*"` is an ok solution. if you know a better way to check that in POSIX / dash i'll be happy to implement it differently.

plugins/ssl/certificate_file_expiry
sumpfralle: Thanks for the explanation. The comments will be useful.
Probably it would have been more obvious to me, if you either used the `-F` flag or escaped the star with a backslash.

PR #1295, opened 2022-02-18
Renew the configuration for the default paths of puppet7

coacx:


sumpfralle:

Did the log file location change recently? Or maybe you are just using a different distribution?
(in this case it would be better, if you would just use a local configuration for this different path)

Regarding the new log line pattern: was this introduced somewhen recently in puppet? When exactly did this happen?
(in case of older versions still being popular, I would prefer a solution that works for both old and new versions)

coacx:
Hey Lars,

I have now had a cursory look at it.
Apparently the difference in log location and line structure is not in versions 5,6 and 7 but in whether you use the Puppetlabs repo or the distribution package.

I'll take a closer look and add a query so that all versions are covered.

coacx:
_Debian 11, package from debian repo:_
```
grep seconds /var/log/daemon.log
Feb 18 17:47:12 privacy puppet-agent[56759]: Applied catalog in 6.50 seconds
[...]
puppet --version
5.5.22
```

_Debian 11, package from puppetlabs repo:_
```
grep seconds /var/log/daemon.log
Feb 14 17:23:50 mcp puppet-agent[61020]: Applied catalog in 3.41 seconds
[...]
puppet --version
7.14.0
```
So it looks like it's up to the distribution. Do you have other examples?

coacx:
Next, I looked at rhel and fedora. With sandard configuration the plugin doesn't work with both anyway, because nothing is logged to a file, but can only be found in the journal.
With these distributions the user must adjust his puppet.conf in any case.
```
[21:16:29] root@fed.kar.vm.loki.tel:~ grep -r "Applying configuration" /var/
grep: /var/log/journal/aac5be625d604965a6ae0be083784925/system.journal: binary file matches
```

coacx:
I have found a way to do this that is independent of the operating system and puppet version used.

sumpfralle:
> So it looks like it's up to the distribution. Do you have other examples?

I use Debian, too, thus I cannot offer more examples. Thank you for taking a deeper look!
As far as I can see, you the `puppet_logfile` is unchanged in your current set of commits, right?
(in the end I would leave it up to you, to specify a default location, as long as you thought a bit about its implications)

I had a quick look at your changes and are happy about the documentation updates.
Thus I would be glad to merge your changes.
Could you please squash your commits a bit - twelve commits are a bit too much for the history :)
(one to four commits are probably reasonable for this amount of changes)

Thank you!

coacx:
I basically rewrote the plugin. Specifying the log or report file is no longer necessary. 
This is determined with the command `puppet config print lastrunreport` regardless of operating system and puppet version.

After that the determined ayml-file is read out and a graph with four values is created. I also added a png in example-graphs to make it clear.

It is tested with all combinations you can see in the documentation.

The whole thing was not planned this way, hence the many commits. I will close this pull request. The new one with clean and traceable commit can be found here: #1298

PR #1298, opened 2022-02-27
More dynamic

coacx:
Enable dynamic reading of puppet report (log file), independent of operating system and puppet version.

The following configuration is mandatory `/etc/munin/plugin-conf.d/puppet.conf`
``` 
[puppet_runtime]
user root
```

Test and make sure it works: `munin-run puppet_runtime`
```
total.value 6.952583373
fact_generation.value 4.8415459030075
catalog_application.value 1.060418801032938
plugin_sync.value 0.6275155210169032
```


plugins/puppet/puppet_runtime
```diff
@@ -1,44 +1,137 @@
...
+=head2 Configuration
+
+The following configuration is mandatory. It must be ensured that 
+the plugin is executed by the root user.
+
+  $ cat /etc/munin/plugin-conf.d/puppet.conf 
+    [puppet_runtime]
+      user root
+
+Further configurations are not necessary. If it does not work,
```
sumpfralle, line 46: This section sounds, like the output of `puppet config print lastrunreport` would be really stored in a file.
But your code looks you are directly reading the output of this command (without writing to a file).
Maybe you could clarify this here? And maybe just omit the specific filename (`/opt/puppetlabs/puppet/cache/state/last_run_report.yaml`) from the documentation, since it sounds like something special, which is required for the plugin?

sumpfralle (commented):
Thank you for working on the plugin!

Please take a look at my comment below.

plugins/puppet/puppet_runtime
coacx: Hi, either you can't understand the code or you don't know the functions of puppet. I dont know, but the inline documentation and the code itself tells you everything you want do know.

The shell command `puppet config print lastrunreport` prints the path to the yaml file which contains all information about the last puppet run.

This can be for example `/opt/puppetlabs/puppet/cache/state/last_run_report.yaml`, which is the default for puppet via the puppetlabs repo for debian. But it can also be `/var/lib/puppet/reports/last_run_report.yaml`, ` /var/puppet/last_run_report.yaml` or whatever you have configured. Also `/etc/my/puppet/datastore/last_run_report.yaml`

The command `puppet config print lastrunreport` just tells you where puppet stored this report. Where to find the file depends on the puppet version, the used operating system and the own configuration.

Just run the command, then you know where the report is in your case. This yaml is available in every puppet version on every operating system. The command only tells you where it is actually located.

What the section in line 43 and 44 should have to do with it, I don't understand. These lines show the content of the Munin plugin configuration file. As you can see in line 42, this must be the content of the file `/etc/munin/plugin-conf.d/puppet.conf`. 

In summary:
- There is only one file that is necessary for the plugin, and that is `/etc/munin/plugin-conf.d/puppet.conf`. If I read the documentation from top to bottom, there are no misunderstandings.
- `last_run_report.yaml` is just the file that is read. The path is determined by `puppet config print lastrunreport`.

coacx: For a better understanding, here is what you would do if you wrote this in bash:
```bash
#!/bin/bash
FILE=$(puppet config print lastrunreport)
echo $FILE
/var/cache/puppet/state/last_run_report.yaml 

cat $FILE | grep -A 1 Total | head -n 5 | tail -1 | awk -F"- " '{print $2}'
59.782151992
```

sumpfralle: > The shell command puppet config print lastrunreport prints the path to the yaml file which contains all information about the last puppet run.

Thanks!. This was the part, that I misunderstood.

> This can be for example /opt/puppetlabs/puppet/cache/state/last_run_report.yaml, which is the default for puppet via the puppetlabs repo for debian.

Maybe you can clarify this in the documentation? I misunderstood it as *the* location (not one of the possible ones - depending on the deployment style).

> What the section in line 43 and 44 should have to do with it, I don't understand.

I wanted to refer to the beginning of the following section (starting with line 46). This may have been confusing for you.

Regarding the location of the specific file (which is created via `puppet config print lastrunreport`): this can be problematic, if certain hardening features (e.g. systemd (`ProtectSystem=strict`), apparmor or SELinux) are configured for the `munin-node` service.
Thus you may want to check, whether it is possible to configure the default location for `print lastrunreport` (e.g. something below `/run` or `/tmp`). This would reduce friction for users, I guess.

Anyway: I am not a puppet user, thus I cannot tell, which of the above mentioned details are obvious for the casual puppet user. Thus if you feel like the current state of the plugin is sufficient, then I will just merge it as it is. Thanks!

coacx: > Maybe you can clarify this in the documentation? I misunderstood it as the location (not one of the possible ones - depending on the deployment style).

> Regarding the location of the specific file (which is created via puppet config print lastrunreport)

https://github.com/coacx/contrib/blob/more-dynamic/plugins/puppet/puppet_runtime#L41-L49

To make it clear once again. The `puppet config print lastrunreport` command only prints you the path to a file that has been in the system for a long time. This lugin creates files, not even temporarily.

This plugin also does not execute a puppet run. 

It basically does nothing else than a grep on a file and writes the result to the variable for Munin.
Therefore system hardening is no problem. Because nothing is written.

I have only improved a plugin with a hardcoded path so that it works regardless of the used Puppet version and operating system.

For example, here you can see that the codebase has been running successfully for weeks. https://mcp.loki.tel/munin/vie.exit.tor.loki.tel/05.vie.exit.tor.loki.tel/puppet_runtime.html

So I have no concerns when it is released.

sumpfralle: > The puppet config print lastrunreport command only prints you the path to a file that has been in the system for a long time.

OK, this part was confusing me before. Now your description in the plugin documentation should make it clear. Thanks!

sumpfralle:

plugins/puppet/puppet_runtime
coacx: As a non-native English speaker, not even in school we were teached that, I have my problems with some phrases.
Thanks for merging!

PR #1300, opened 2022-03-13
nginx_unit: use more human readable scale

kimheino:
Value "2.34M seconds" doesn't mean anything  but "27 days" does.  Use days for process age.


sumpfralle:
Looks good, thank you!

PR #1302, opened 2022-03-21
[system/pressure] Add plugin to monitor pressure stall information (psi)

HaseHarald:
Adds a plugin to monitor the pressure stall information (psi) as reported by the Linux kernel.

- groups averages per resource
- rate/derive totals for ease of reading
- resources, intervals and scopes configurable

See: https://www.kernel.org/doc/html/latest/accounting/psi.html

I've written this plugin because I read about the metric and wanted to get a feel for what's "normal" on these values. Therefore I did not include any warnings and am not certain that all metrics are actually useful. For example I'm not sure if it helps to monitor the 10 and 60 second averages long term. So I made this stuff configurable. Still this might help someone else, or someone has an idea on an improvement. So here it is for everyone to use.

### DIFF
--- a/plugins/system/pressure
+++ b/plugins/system/pressure
@@ -1,0 +1,331 @@
+#!/bin/bash
+
+
+: << =cut
+
+=head1 NAME
+
+pressure - Plugin to monitor the pressure stall information for CPU, Memory and
+IO as reported by the Linux kernel.
+
+This plugin monitors the pressure stall information (psi) as reported by the
+Linux Kernel. By default it reports all average intervals (10 seconds,
+60 seconds and 300 seconds) as well as the total values as a rate of change
+(DERIVE) for all resources (cpu, memory, io). The average intervals can be
+configured if you only deem some of them useful. See CONFIGURATION for
+explanations on that.
+
+This is a multigraph plugin that, by default, will create six detail graphs and
+one summary graph (so seven in total). The summary graph will contain the 300
+seconds average percentages of all resources. The detail graphs are split in two
+graphs per resource. One combining all average intervals and one for the
+"totals" (rate of change) for the given resource.
+
+There are no defaults for warnings and criticals, because this highly depends on
+the system, so you need to configure them yourself (if you want any). It is
+recommended that you first lookup the meaning of the different values.
+
+For more information on psi see:
+https://www.kernel.org/doc/html/latest/accounting/psi.html
+
+=head1 CONFIGURATION
+
+Simply create a symlink in your plugins directory like with any other plugin.
+No additional configuration needed, no specific user required (typically).
+
+If you want to configure alerts, just add "warn_" or "crit_" in front of the
+internal name.
+
+Optional configuration examples:
+
+[pressure]
+env.resources cpu io memory       - Specify the resources to monitor. Leave one
+                                    out if you don't want this one to be
+                                    monitored.
+env.intervals avg10 avg60 avg300  - Sepcify the average intervals to monitor.
+                                    Leave one out if you don't want this one to
+                                    be monitored
+env.scopes some full              - Specify the scopes to monitor. Leave one out
+                                    If you don't want it to be monitored.
+env.summary_interval avg300       - Specify the interval to be used for the
+                                    summary-graph.
+env.warn_psi_cpu_avg300_some 5    - Set a warning-level of 5 for
+                                    "psi_cpu_avg300_some"
+env.crit_psi_io_total_full 2000   - Set a critical-level of 2000 for
+                                    "psi_io_total_full"
+
+=head1 AUTHOR
+
+2022, HaseHarald
+
+=head1 LICENSE
+
+LGPLv3
+
+=head1 BUGS
+
+=head1 TODO
+
+=head1 MAGIC MARKERS
+
+ #%# family=auto
+ #%# capabilities=autoconf
+
+=cut
+
+
+# This file contains a munin-plugin to graph the psi (pressure) for CPU, Memory
+# and IO, as reported by the Linux kernel.
+#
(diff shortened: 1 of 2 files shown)


plugins/system/pressure
```diff
@@ -0,0 +1,331 @@
...
+pressure_resources=${resources[@]:-${resource_defaults[@]}}
+pressure_intervals=${intervals[@]:-${interval_defaults[@]}}
+pressure_scopes=${scopes[@]:-${scope_defaults[@]}}
+summary_interval=${summary_interval:-avg300}
+
+check_autoconf() {
+  if [ -d "${pressure_dir}" ]; then
+    printf "yes\n"
+  else
+    printf "no (%s not found)\n" ${pressure_dir}
```
sumpfralle, line 107: Please add quotes. See the output of `shellcheck` for such issues.

plugins/system/pressure
```diff
@@ -0,0 +1,331 @@
...
+  fi
+}
+
+get_pressure_value() {
+  resource=$1
```
sumpfralle, line 112: Please use `local` for variables inside of functions in order to keep the namespace clean.

plugins/system/pressure
```diff
@@ -0,0 +1,331 @@
...
+  interval=$2
+  scope=${3:-some}
+  grep "$scope" ${pressure_dir}//${resource} | grep -o -E "${interval}=[0-9]{1,}(\.[0-9]{1,}){0,1}" | cut -d '=' -f 2
```
sumpfralle, line 115: Just one slash is sufficient for the `grep` path, oder?

Please quote the path.

Regarding `{1,}` in the regular expression: wouldn't be `+` an easier readable way of expressing this?

plugins/system/pressure
```diff
@@ -0,0 +1,331 @@
...
+      esac
+      ;;
+    
+    *)
+      printf "ERROR: Could not determine kind %s ! Must be one of 'interval' 'scope' 'resource'\n" $kind >&2
+      exit 2
+      ;;
+  esac
+  
+  printf "%s\n" $printable_name
```
sumpfralle, line 184: I guess, the linebreak is not relevant here?
(trailing linebreaks are automatically removed in a *Command Substitution*)

plugins/system/pressure
```diff
@@ -0,0 +1,331 @@
...
+  for scope in ${pressure_scopes[@]}; do
+    if [ ${resource} == "cpu" ] && [ ${scope} != "some" ]; then
+      continue
+    else
+      printf "psi_%s_%s_%s.value %s\n" $resource $interval $scope $(get_pressure_value $resource $interval $scope)
+    fi
+  done
+}
+
+output_usage() {
```
sumpfralle, line 292: Maybe simply redirect `output_usage` to *stderr* when it is called?
(in order to remove all these redirects here)

plugins/system/pressure
```diff
@@ -0,0 +1,331 @@
...
+  printf >&2 "%s - munin plugin to graph pressure stall information for CPU, Memory and IO as reported by the Linux kernel.\n" ${0##*/}
```
sumpfralle, line 293: Maybe suggest to run `munin-doc PLUGIN_NAME` instead of repeating the content of the documentation header here?

plugins/system/pressure
```diff
@@ -0,0 +1,331 @@
...
+  
+  1)
+    case $1 in
+      auto|autoconf)
+        check_autoconf
+        ;;
+      config)
+        iterate_config
+        ;;
+      *)
```
sumpfralle, line 320: `fetch` is also an acceptable alternative for "no argument".

plugins/system/pressure
```diff
@@ -0,0 +1,331 @@
...
+}
+
+case $# in
+  0)
+    iterate_values
+    ;;
+  
+  1)
+    case $1 in
+      auto|autoconf)
```
sumpfralle, line 314: `auto` is not specified. Please treat it as an error.

sumpfralle (commented):
Thank you for contributing this plugin!

In addition I have the following suggestions:
* please take a look at the issues reported by `shellcheck`
* I would recommend to use a less generic name for this plugin (not just "pressure"). How about `pressure_stall`, `linux_pressure` or `linux_psi`?

Thank you!

HaseHarald:
Hello @sumpfralle,

thanks for the feedback. Sorry it took so long to answer, I've been busy. I'm going to go through the issues and resolve them if needed. On some of them I've got a different opinion though. I'll add them right next to your code-quotes so they are easier to relate.

plugins/system/pressure
HaseHarald: Well, that depends on your definition of "easy". The way it currently is set, the regex works the same way throughout. It's all the same pattern. Whereas your solution (using `+`) brings in a different notation, where one reading this will need to think about it again.

Off course one could argue that this is still common regex that most everybody knows by heart, if they know regex at all, and that regular changes can help to keep you awake. But still its opinion either way around.

The double slash is a typo by the way. I'm going to fix that.

HaseHarald:
Hello,

I've fixed several of the issues in 2ee1161a. The once not changed I elaborated above.

Also I renamed the plugin to `linux_psi` which really is the better name.

Have a nice day

HaseHarald

plugins/system/pressure
sumpfralle: OK, you thought about the possible ways of expressing this. So I am happy with your choice.

sumpfralle:
Your changes look good to me. Thank you!

I merged your plugin now.

btw.: just today another contributor proposed another munin plugin for this topic in [munin#1462](https://github.com/munin-monitoring/munin/pull/1462). Maybe you want to take a look at it.

PR #1303, opened 2022-03-27
chrony: don't resolve IP addresses to hostnames

antonv6:
If we don't want to show any host names in graphs/legend, it makes sense to not resolve them.

I occasionally notice the issue with name resolution of IPv6 addresses on certain networks, which makes `chronyc tracking` take 20+ seconds. Adding -n makes it almost instantaneous.


sumpfralle:
Looks great - thanks!

PR #1305, opened 2022-04-04
[arris-sb8200] Add plugin to support Arris SB8200 cable modem

utopiabound:
This adds a plugin that supports the Arris SB8200 cable modem
- Uptime
- Downstream error counts
- Downstream Signal-to-Noise ratio (dB)
- Downstream & Upstream Power (dBmV)


sumpfralle:
This new plugin looks good to me. Thank you!

PR #1306, opened 2022-04-06
Update memory_osx plugin

sdarwin:
A new version of memory_osx.

Inspired by the answer in this stackexchange thread: https://apple.stackexchange.com/questions/423717/is-there-a-command-line-tool-that-accurately-describes-the-amount-of-used-memory

Screenshot:
memory-week

It's more colorful and interesting than the memory plugin from the earlier version: https://github.com/munin-monitoring/contrib/pull/1173

The vm_stat command lists many memory regions.  An important question was whether the selected values would consistently add up to 100% of total memory and not overlap, which would cause the graph to grow or shrink.  Luckily, that seems to be correct, because the dividing line between memory and swap space is flat.

> Please avoid using bc

In this particular case, I didn't see how to simply and conveniently replace all instances of `bc`.
Mac OSX includes `/usr/bin/bc`.   The command is available and so that might not be a problem that needs to be solved.


plugins/memory/memory_osx
```diff
@@ -35,33 +34,68 @@ if [ "$1" = "autoconf" ]; then
...
+        echo "graph_args --base 1024 -r --lower-limit 0"
         echo 'graph_vlabel Bytes'
         echo 'graph_category system'
-        echo 'used.label used (not incl. wired)'
-        echo 'used.draw AREA'
-        echo 'used.min 0'
-        echo "used.info Used memory, not including wired"
         echo 'wired.label wired'
-        echo 'wired.draw STACK'
+        echo 'wired.draw AREA'
```
sumpfralle, line 48: You can use `AREASTACK` everywhere (i.e. for the first and for all subsequent entries).
This would make it easier to read such a patch.

plugins/memory/memory_osx
```diff
@@ -35,33 +34,68 @@ if [ "$1" = "autoconf" ]; then
     fi
 fi
 
-TOTALMEM=$(sysctl hw.memsize | cut -d" " -f2)
-graphlimit=$TOTALMEM
-
 dehumanise() {
     echo "$1" | sed -e "s/K/*1024/g;s/M/*1024*1024/;s/G/*1024*1024*1024/;s/T/*1024*1024*1024*1024/"
```
sumpfralle, line 38: Just as a recommendation for clearer code (and for avoiding `bc`):
```shell
local value="$1"
case "$value" in
    *K) echo "$((value * 1024))";;
    *M) echo "$((value * 1024 * 1024))";;
    *G) echo "$((value * 1024 * 1024 * 1024))";;
    *T) echo "$((value * 1024 * 1024 * 1024 * 1024))";;
    *) echo "$value";;
esac
```

But feel free to keep it unchanged, if you prefer.

sumpfralle (commented):

Please take a look at my comments (just recommendations). I would also merge it as it is, if you prefer.

plugins/memory/memory_osx
sdarwin: unfortunately it won't work exactly that way because the output contains decimal places, floating point values, such as 1.23.

sumpfralle:

PR #1307, opened 2022-04-22
Fix mikrotik_system -  temp + voltage

quotengrote:
Hi,

with the latest versions of RouterOS some Names have changed.

That PR fixes the Errors for "temperature" and "voltage".

It is tested on RB5009, hex, CRS309, CRS305 with RouterOS 7.2.1. It is NOT tested with older Versions of RouterOS because i dont have any devices left with it.


sumpfralle:
Thank you!

PR #1308, opened 2022-05-03
[irc2] Add port and SSL/TLS support, add IPv6 env variable

codeurimpulsif:
## Goal
- Add port support (defaults to `6667`)
- Add SSL/TLS support (defaults to `0`)
- Add IPv6 environment variable
- Improve documentation
- Remove `host_name` config to avoid localhost IRC monitoring limit

### DIFF
--- a/plugins/ircd/irc2
+++ b/plugins/ircd/irc2
@@ -8,11 +8,16 @@ ircstats  - Plugin to graph data about an IRC network and a single IRC server
 =head1 CONFIGURATION
 
 - ENV{SERVER} to point to the server to connect to, defaults to localhost.
+- ENV{PORT} port to use, defaults to 6667.
 - ENV{NICK} nickname to use, defaults to munin-$HASH.
+- ENV{USESSL} 0 or 1 values to enable SSL/TLS, defaults to 0.
+- ENV{USEIPV6} 0 or 1 to enable IPv6 use, defaults to 0.
 
 =head1 USAGE
 
 This plugin connects to an IRC server.
+
+It require POE::Component::IRC and POE::Component::SSLify if you use SSL/TLS.
 
 =head1 AUTHOR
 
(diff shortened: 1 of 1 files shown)


plugins/ircd/irc2
```diff
@@ -7,13 +7,18 @@ ircstats  - Plugin to graph data about an IRC network and a single IRC server
...
+- env.PORT port to use, defaults to 6667.
+- env.NICK nickname to use, defaults to munin-$HASH.
+- env.USESSL 0 or 1 values to enable SSL/TLS, defaults to 0.
+- env.USEIPV6 0 or 1 to enable IPv6 use, defaults to 0.
 
 =head1 USAGE
 
 This plugin connects to an IRC server.
 
+It require POE::Component::IRC and POE::Component::SSLify if you use SSL/TLS.
```
sumpfralle, line 20: *It requires* (the `s` was missing)

plugins/ircd/irc2
```diff
@@ -70,8 +78,10 @@ my $irc = POE::Component::IRC->spawn(
    nick => $nickname,
    ircname => $ircname,
    server => $server,
+   port => $port,
    raw => 0,
-   useipv6 => 0,
+   usessl => $usessl,
```
sumpfralle, line 83: Are you sure, that the string (taken from `$ENV{USESSL}`) is handled properly? Or should you convert it into a numeric value? Is `USESSL=0` really interpreted as *no SSL*?

sumpfralle (commented):

I added two remarks below.

plugins/ircd/irc2
codeurimpulsif: fixed ;)

plugins/ircd/irc2
codeurimpulsif: Yes it's really interpreted as *no SSL* as [mentioned in the documentation](https://metacpan.org/pod/POE::Component::IRC#spawn), and maybe I can add and int conversion to be sure?

codeurimpulsif:
Is it ok?

sumpfralle:
Sorry for my late response - I was away for some days ...

Regarding the string `"0"` for `usessl` being interpreted by Perl as `false`: yes, this seems to be one more quirky language feature surprising me (I rarely write Perl). Thank you for clarifying that.

Your contribution looks fine to me. Thank you!

PR #1309, opened 2022-05-04
smtp_hello_ improvements

ap-wtioit:
This contains the following improvements:
* enable running the plugin with bash time keyword if time executable is not found
* enable using curl with a proper SMTP client implementation instead of `echo | nc` to enable monitoring servers that do not automatically terminate the connection after `HELO localhost` is answered.
* enable configuring the SMTP test command (e.g. to `QUIT` or `HELP`)
* enable configuring warning and critical levels
* check status when using curl to detect if mailserver is not responding as expected / not connectable
* enable monitoring multiple ports on one host and ports different from 25

Info @wt-io-it

### DIFF
--- a/plugins/network/smtp_hello_
+++ b/plugins/network/smtp_hello_
@@ -41,14 +41,8 @@ if [ "${MUNIN_DEBUG:-0}" == "1" ]; then
 fi
 
 SMTP_COMMAND=${SMTP_COMMAND:-HELO localhost}
-SMTP_PORT=${SMTP_PORT:-25}
+SMTP_PORTS=${SMTP_PORTS:-25}
 TIMEOUT_SECONDS=${TIMEOUT_SECONDS:-120}
-# if we have curl all non-zero exit codes indicate an error
-status_warning=${status_warning:-0:0}
-# default warning time begins at half the timeout
-host_warning=${host_warning:-:$((TIMEOUT_SECONDS/2))}
-# default when we get values bigger than 99% of timeout we mark the value as critical
-host_critical=${host_critical:-:$((TIMEOUT_SECONDS * 99 / 100))}
 
 host=$(basename "$0" | sed 's/^smtp_hello_//g')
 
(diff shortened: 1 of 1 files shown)


plugins/network/smtp_hello_
```diff
@@ -20,37 +17,227 @@
...
+elif [ -x /usr/bin/time ]; then
+  # use time executable found in /usr/bin
+  time_executable="/usr/bin/time"
+fi
+if [ "$time_executable" != "" ]; then
+  function take_the_time() {
+    "$time_executable" -f "%e" "$@" 2>&1 | tail -n 1
+    return "${PIPESTATUS[0]}"
+  }
+elif (time echo) >/dev/null 2>&1; then
```
sumpfralle, line 148: I think, you can omit the brackets here. At least I do not see a reason for spawning a sub-shell.

plugins/network/smtp_hello_
```diff
@@ -20,37 +17,227 @@
...
+  # if we cannot find time executable but time keyword is successful we use the keyword
+  function take_the_time() {
+    # we use LC_ALL=C to force . as decimal separator on systems where LC_ is set to a language that uses ,
+    (
+      TIMEFORMAT='%R'
+      LC_ALL=C
+      time "$@" 1>/dev/null
+    ) 2>&1
```
sumpfralle, line 156: I think, you can reduce this down to the following:
```shell
{ TIMEFORMAT='%R' LC_ALL=C time "$@" 1>/dev/null; } 2>&1
```

At least for me it is easier to read (no complications with `$?`, no sub-shell).
But this may be a matter of taste.

sumpfralle (commented):

These changes look good and useful to me. Please take a look at my minor comments.
(but feel free to ignore them ...)

plugins/network/smtp_hello_
ap-wtioit: Ok i need to read up on the `{...;}`, this works as well to achive the stderr redirection mentioned above.

But: (sorry)
On systems without the time executable (usually `/usr/bin/time`)
```bash
{ TIMEFORMAT='%R' LC_ALL=C time echo "OK" 1>/dev/null; } 2>&1
```
gives an error:
```
-bash: time: command not found
```
whereas on systems with the time executable it clearly shows the output from the time command (ignoring the variables for the time keyword):
```
0.00user 0.00system 0:00.00elapsed ?%CPU (0avgtext+0avgdata 1452maxresident)k
0inputs+0outputs (0major+67minor)pagefaults 0swaps
```

eventually you missed the intentional newlines:
```bash
TIMEFORMAT='%R'
LC_ALL=C
time "$@" 1>/dev/null
```
is not equivalent to
```bash
TIMEFORMAT='%R' LC_ALL=C time "$@" 1>/dev/null
```
but should be
```bash
TIMEFORMAT='%R'; LC_ALL=C; time "$@" 1>/dev/null
```
which would work:
```bash
{ TIMEFORMAT='%R'; LC_ALL=C; time echo "OK" 1>/dev/null; } 2>&1
```
gives
```
0.000
```
this then would set `TIMEFORMAT` and `LC_ALL` for the following script when used with `{...;}` instead of `(...)`, so i think the subshell is ok here (even if i do not know of any sideeffects those 2 variables would have for the rest of the script, i would like to make sure they are contained in the subshell)

ap-wtioit:
> Thank you for your contribution!
> 
> These changes look good and useful to me. Please take a look at my minor comments. (but feel free to ignore them ...)

Thanks for the fast review. I replaced the first test for the keyword `(time echo)` with `{ time echo; }` to not have a subshell here. 

But i prefer having the second subshell containing the `LC_ALL=C` just to make sure.

sumpfralle:
Sorry for my late response ...

Thank you for your helpful responses, which indeed clarified my misunderstandings!

Transparently tolerating both a shell builtin and an external command is really a pain :(

I *could* imagine, that a quick summary of the above subtle details ("builtin vs. command" for redirection and for variable passing) as a comment would be helpful for the next reader of these code pieces. But I leave it up to you, whether you are in the mood or not. Otherwise I would just merge your useful contribution as it is.

Thank you!

ap-wtioit:
i added a comment inside of `function take_the_time()` to mention that the effect of the assigned values for LC_ALL and TIMEFORMAT should be limited (as well as to make stderr redirection possible for the keyword / builtin).

i also rebased onto the current master branch.

sumpfralle:
Thank you for improving this plugin with its subtle details!

PR #1310, opened 2022-05-12
Update vsftpd-rel: use mktemp

xkill:
Use mktemp instead tempfile, as tempfile is deprecated.


sumpfralle:

PR #1313, opened 2022-05-24
Nutups2 multiplier

spacelama:
Add ability to provide multiplier to common_fetch to manufacture a reading out of an already existing raw reading (eg, power when the UPS only outputs load).

I've supplied an example multiplier for my particular UPS, but that example belongs in the configuration file.  I just don't know how to get that optional multiplier from the config file to this function in an appropriate way.

### DIFF
--- a/plugins/power/nutups2_
+++ b/plugins/power/nutups2_
@@ -147,8 +147,8 @@ my %config = (
 		args => '--base 1000 -l 0',
 		vlabel => 'Watt',
 		config => \&common_config,
-                multiplier => 6.349,  # 400 watts at 63% load
-		fetch => \&multiplier_fetch,
+		multiplier => 6.349,  # 400 watts at 63% load on powershield commander 1000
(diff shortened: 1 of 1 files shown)


sumpfralle:
Thank you for working on this possible extension of the plugin!

I would recommend to read this *multiplier* value from the environment. This will allow separate values for different power supply devices. Please try to find a *descriptive* name for this environment variable (e.g. "foo_something_ratio") and describe it in the plugin documentation header.

Regarding your branch: please ensure, that it does not contain merge commits. Use `rebase` in order to achieve this.
Currently your branch contains the `tasmota`-related changes, too. Instead there should be only *one* simple commit.

Thank you!

spacelama:
Who knows how I screwed up that original git commit history, and I don't know why I didn't push this a year earlier when I had apparently done all the work already...

Force-pushed a revised commit history in order not to have to redo this pull request.  Hopefully that's OK...

PR #1319, opened 2022-06-19
puppet_runtime: fix perlpod

kenyon:
`perlpod` needs these blank lines in order to detect and format the documentation.


sumpfralle:
Thanks!

PR #1322, opened 2022-07-09
[exodus_] Add warning and critical configurations

codeurimpulsif:
## Goal
- Add `env.warning` and `env.critical` configurations for each suffix
- Improve documentation accordingly


codeurimpulsif:

PR #1324, opened 2022-07-11
[rabbitmq_messages*] Fix rabbitmqctl output filter

codeurimpulsif:
## Goal

- Fix rabbitmqctl output filter to avoid bad values

## Informations

Like #1323 

By default `rabbitmqctl list_queues -p / name messages` give this kind of output:
```
Timeout: 60.0 seconds ...
Listing queues for vhost / ...
name	messages
celery	0
```

Then in this plugin it result to:
```
Timeout:.value 60.0
name.value messages
celery.value 0
```

This merge request fix the grep command to get this plugin result:
```
celery.value 0
```


codeurimpulsif:

PR #1336, opened 2022-10-06
Allow dovecot_stats_ cope with Dovecot 2.3

### LINKED ISSUE #1327: dovecot_stats_ only works for dovecot 2.1 and 2.2
I got confused after a big migration that dovecot_stats_ was acting up. Turns out, as mentioned on https://doc.dovecot.org/configuration_manual/stats/ , that dovecot stats using doveadm were completely redone in dovecot 2.3 (first release in December 2017, last 2.2 security release in august 2019). The command executed in dovecot_stats_ is therefore no longer valid. I wonder whether the plugin should be marked to only be useable with 2.1 and 2.2, or whether it should be removed altogether.

darac:
Dovecot has a new "stats" command, and the version that this plugin works with is now called as "doveadm oldstats", so try to adapt the command based on the dovecot version.

Fixes #1327

### DIFF
--- a/plugins/dovecot/dovecot_stats_
+++ b/plugins/dovecot/dovecot_stats_
@@ -36,8 +36,24 @@ if [[ "$1" == "autoconf" ]]; then
     exit 0
 fi
 
+# Dovecot 2.3 changes the stas format, but we can still access the older version with "doveadm oldstats".
+dovecot_version=$(/usr/sbin/dovecot --version | awk '{print $1}')
+
+verlte() {
+    [ "$1" = "$2" ] && return 1 || [  "$2" = "`echo -e "$1\n$2" | sort -V | head -n1`" ]
+}
+
+verlt() {
+    [ "$1" = "$2" ] && return 1 || verlte $2 $1
+}
+
+# The stats command is "stats" unless the version is NOT less than 2.3, in which case it's "oldstats".
+stats_command="stats"
+verlt $dovecot_version 2.3 || stats_command="oldstats"
+
+
 if [[ "$1" == "suggest" ]]; then
-    doveadm stats dump domain|awk 'NR!=1 {print $1}'
+    doveadm $stats_command dump domain|awk 'NR!=1 {print $1}'
     exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


bertvandepoel:
Does it actually work with oldstats? I've been trying to get this working after receiving notification that this was fixed, but running "doveadm oldstats dump domain" gives me "Info: no statistics available" and modules "stats" and "imap_stats" no longer exist in 2.3. Did you fully check this? Is certain preconfiguration required that should be added to the header of this plugin script? I feel like I'm doing something wrong but I'm not quite sure what.

darac:
Now that you mention it, I see that I have the following in `/etc/dovecot/conf.d/plugin-stats.conf`:
```
# Enable Statistics
mail_plugins = $mail_plugins old_stats
protocol imap {
	mail_plugins = $mail_plugins imap_old_stats
}

plugin {
	old_stats_refresh = 30 secs
	old_stats_track_cmds = yes
}

service old-stats {
	fifo_listener old-stats-mail {
		user = vmail
		mode = 0600
	}
}
```

If you find that works, I'll update the plugin documentation.

bertvandepoel:
Took me a bit of time to figure out. Another config file already defined mail_plugins so I had to add old_stats over there. I suggest describing that the old_stats plugin should be added to mail_plugins globally, that should be enough to help any sysadmin figure out themselves where to put the relevant term.

I see you are only using `doveadm oldstats dump domain` in the dovecot_stats_ plugin. If I understand things correctly, imap_old_stats and old_stats_track_cmds are both used for per-command statistics, available using `doveadm oldstats dump command`. I think that that's therefore not necessary, or is that used to enable the "Commands" metric in one of the graphs?

Finally, the user, group and mode for the fifo_listener I had hoped to make things night and tight when it comes to privileges (vmail for dovecot internals, but the munin group for example). To my frustration, it seems that doveadm insists on running doveconf over the entire config and verifying everything, even when requesting stats. This results in failure due to it not being able to read my SSL/TLS certificate. This can be solved by running the plugin as root, but I prefer not to do that to be honest. I'm taking the matter up with the Dovecot mailing list, but for now we probably want to advise running as root in the munin-node plugin config. Another solution could be to talk to the socket directly instead of using doveadm, but such a change would probably make sense when also switching from oldstats to the new 2.3 stats. I'm not sure if you have any plans for that?

PR #1340, opened 2022-10-20
Update ssl-certificate-expiry to allow legacy renegotiation

simontuckwell:
This small change can be used to enable SSL legacy renegotiation.  It is sometimes required for older software or hardware - in my case a cluster of Aruba APs.


Neustradamus:
@simontuckwell: What is the situation of your PR?

PR #1371, opened 2023-04-27
lxc_guests: Enhanced and fixed lxc_guests plugin

brknkfr:
- Working with systemd and cgroup version 2 (tested on debian bullseye and debian booksworm)
- Fixed processes (with cgroup version 2)
- Simplified labels
- Simplified memory usage graph
- Added CPU usage in percent (using systemd-cgtop)
- Added Tasks
- No cgrouppath guessing
- Removed support for old lxc versions and systems running on cgroup version 1


kenyon:
Why not update the existing `lxc_guests` plugin?

brknkfr:
> Why not update the existing `lxc_guests` plugin?

It would be possible (as everything in informatics), but I think there will be too much so called "workarounds" included to make this plugin compatible to `lxc_guests`. To name a few "changes" or "problems":

- the old cgroup version 1 has nanoseconds in cpuacct.stat in whereas cgroup version 2 has miliseconds in cpu.stat for usage_usec
- with systemd-cgtop it's easier to get cpu usage in percet whereis older versions of systemd-cgtop can't do that yet
- counting the actual processes is more difficult to do in cgroup2 than in cgroup1 (relates to all the cgrouppath stuff)

so, that's why.

kenyon:
Yeah, figured `lxc_guests2` one is not compatible, but if `lxc_guests` doesn't work on modern systems, is it even worth keeping?

brknkfr:
In my opinion, `lxc_guests` could be replaced by `lxc_guests2`, but there are probably still old systems out there which use the old `lxc_guests`. In case of a replacement, I don't know if it would be sufficient to just add a note, that old systems should use an older version of `lxc_guests` or would it be better to give a new name to the `lxc_guests2` plugin, something like `lxc_containers`. What do you think?

kenyon:
I would just replace `lxc_guests` with your update. Users should be pinning to specific commits, so I don't think it should break old systems.

PR #1374, opened 2023-05-22
dhcp/kea: add plugin to monitor Kea DHCP server

kimheino:
Kea is ISC's recommended DHCP server, old dhcpd is not maintained anymore. 

See https://www.isc.org/kea/ for more information.


kimheino:
Allocation is not working as intended. I'll create a new pr later.

PR #1380, opened 2023-07-19
dnsresponse_: enable verifying responses

ap-wtioit:
add option verify to enable graphing the number of successful responses. if verify is enabled (default it is not) then a warning is emitted when we do not get a response for all queries and critical if no query succeeded.

add option timeout to be able to limit the time the plugin takes for unresponsive servers. old max time was 75 seconds times 20 (25minutes). new default max is 30 seconds timeout times 20 (10 minutes). with setting a timeout of a few seconds this can now be reduced even more.

added warnings if average and or mean get close to the timeout.

info @wt-io-it


ap-wtioit:
Updated to be compatible with libnet-dns-perl 1.10-2 on Ubuntu 18.04

PR #1409, opened 2024-02-01
asterisk: enable pjsip call channel support and more default codecs

zebastian:
- add pjsip channel retrieval through ami command "pjsip show channelstats"
- also add more common codecs g723 and g729 to default list

- tested against:
  - an asterisk 18.19.0 with pjsip calls running, which now includes the pjsip channels
  - an old asterisk 16.6.2 running sip calls


zebastian:
thanks @kenyon

PR #1439, opened 2024-06-25
Add dnsperftest plugin to monitor response times of multiple DNS servers

cweiske:
Includes an example graph.

DNS server IPs are read from `/etc/resolv.conf`. This path can be overridden with `env.resolvconf`.
A list of domains is tested that can be overriden as well with `env.domains`.

Based on https://github.com/cleanbrowsing/dnsperftest

Bash script was shellcheck'ed, and all relevant issues were resolved.
Requires `dig` and `bc`.


kenyon (approved):

cweiske:
Thanks for the quick merge!

PR #1444, opened 2024-07-26
Add CrowdSec plugin

d0m84:
Add support for [CrowdSec](https://www.crowdsec.net/) using `cscli`


d0m84:
The question is if it makes sense to put it into the folder `security/` or if I should create a new tool specific folder `crowdsec/`

kenyon (approved):
The `security` directory is fine.

PR #1450, opened 2024-09-09
First version of the resolvectl munin plugin

SAH62:


plugins/resolvectl/resolvectl
```diff
@@ -0,0 +1,121 @@
...
+if [ "$1" = "autoconf" ]; then
+	if ( resolvectl statistics >/dev/null 2>&1 ); then
+		echo yes
+		exit 0
+	else
+		if [ $? -eq 127 ]; then
+			echo "no (could not run \"resolvectl\")"
+			exit 0
+		else
+			echo "no (unknown resolvectl return value $?)"
```
kenyon, line 55: Minor problem here with reporting the return value of the `resolvctl` command: https://www.shellcheck.net/wiki/SC2319

kenyon (changes requested):

plugins/resolvectl/resolvectl
SAH62: Thanks. I just fixed that code and committed the change.

kenyon (approved):

PR #1452, opened 2024-09-11
feature: ADB temperatures

mdepa91:
read Android temperatures using ADB


kenyon (changes requested):
Instead of having its own directory, can you move these files to `plugins/sensors`?

mdepa91:
> Instead of having its own directory, can you move these files to `plugins/sensors`?

oh can be in that way.. done

kenyon (approved):

PR #1453, opened 2024-09-14
journald support of dovecot inspired by the sshd_log plugin

skleber:


plugins/dovecot/dovecot
```diff
@@ -53,76 +78,52 @@ if [ "$1" = "config" ]; then
...
+        CURSOR_FILE=
 fi
-echo -n
+
+if [ "$LOGFILE" = "journald" ]; then
+        # shellcheck disable=SC2086
+        if [ "$TYPE" = "ABSOLUTE" ]; then
+                journalctl --no-pager --quiet --show-cursor ${CURSOR:+"--after-cursor=$CURSOR"} $JOURNALCTL_ARGS
+        else
+                journalctl --no-pager --quiet --since=$(date -dlast-sunday +%Y-%m-%d) $JOURNALCTL_ARGS
```
kenyon, line 103: This looks indented too much:

```suggestion
        journalctl --no-pager --quiet --since=$(date -dlast-sunday +%Y-%m-%d) $JOURNALCTL_ARGS
```

plugins/dovecot/dovecot
```diff
@@ -53,76 +78,52 @@ if [ "$1" = "config" ]; then
...
+        # shellcheck disable=SC2086
+        if [ "$TYPE" = "ABSOLUTE" ]; then
+                journalctl --no-pager --quiet --show-cursor ${CURSOR:+"--after-cursor=$CURSOR"} $JOURNALCTL_ARGS
+        else
+        journalctl --no-pager --quiet --since=$(date -dlast-sunday +%Y-%m-%d) $JOURNALCTL_ARGS
+        fi
+else
+        cat "$LOGFILE"
+fi | \
+    awk -v cursor_file="$CURSOR_FILE" -v connected="$(doveadm who -1 | expr $(wc -l) - 1)" 'BEGIN{c["login_total"]=0;c["connected"]=connected;c["login_tls"]=0;c["login_ssl"]=0;c["login_imap"]=0;c["login_pop3"]=0; }
```
kenyon, line 108:
```bash
connected="$(doveadm who -1 | expr $(wc -l) - 1)"
```

Is this part correct? Piping to `expr`?

kenyon (commented):
Sorry, looks like that was wrong feedback about the indentation. This file is kind of a mess because of the mix of tabs and spaces used for indentation, and the trailing whitespace.

plugins/dovecot/dovecot
skleber: Yes, it is. I sutract 1 from the number of lines (wc -l) because (coveadm who -1) contains one line of headers.

skleber:
> Sorry, looks like that was wrong feedback about the indentation. This file is kind of a mess because of the mix of tabs and spaces used for indentation, and the trailing whitespace.

Sorry for that. I provide a new commit fixing this.

plugins/dovecot/dovecot
kenyon: Oh I see. I couldn't find any docs on what `expr` does with its standard input. I would usually do this with `tail`, like 

```bash
connected="$(doveadm who -1 | tail +2 | wc -l)"
```

kenyon (approved):

plugins/dovecot/dovecot
skleber: Your solution would have worked just as well. I simply used the more "arithmetic" approach.
Here is the man page of `expr`: https://man7.org/linux/man-pages/man1/expr.1.html
Thanks for merging.

skleber: Oh, now I see your confusion. It could also have been
`connected="$(expr $(doveadm who -1 | wc -l) - 1)"`
would have even been more readable.

PR #1456, opened 2024-09-26
letsencrypt_weekly: enable monitoring multiple directories

ap-wtioit:
Enable specifying `letsencrypt_dirs` to be able to monitor different / multiple directories.

This is useful when having the certificates in a docker volume used by docker containers without sharing those directories in /etc/letsencrypt/.

Info @wt-io-it

### DIFF
--- a/plugins/ssl/letsencrypt_weekly
+++ b/plugins/ssl/letsencrypt_weekly
@@ -25,6 +25,15 @@ You can configure the warning and critical limits for this plugin:
   # critical when more than 50 certificates have been requested in the last week
   env.critical :50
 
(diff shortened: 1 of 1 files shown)


plugins/ssl/letsencrypt_weekly
```diff
@@ -25,6 +25,15 @@ You can configure the warning and critical limits for this plugin:
   # critical when more than 50 certificates have been requested in the last week
   env.critical :50
 
+If you have multiple letsencrypt directories e.g. one running in docker, you can configure them seperated by spaces:
```
kenyon, line 28:
```suggestion
If you have multiple letsencrypt directories e.g. one running in docker, you can configure them separated by spaces:
```

kenyon (approved):

PR #1459, opened 2024-10-12
syslog_ng_stats: mark total counter stats as DERIVE / min 0

iasjennen:
the data produced by the underlying `syslog-ng-ctl stats` gives the stats values in absolute numbers since the start of the syslog-ng daemon. In order to visualize in a more user friendly way those values are now marked in the config as type DERIVE.

here a comparison.

before the change with absolute values:
syslog-ng-abs

here with derive:
grafik

### DIFF
--- a/plugins/syslog/syslog_ng_stats
+++ b/plugins/syslog/syslog_ng_stats
@@ -117,6 +117,7 @@ if( exists $ARGV[0] and $ARGV[0] eq "config" ) {
         printf "%s.label %s: %s, %s\n", $id,
             $graph->{source_name},$graph->{source_id}, $graph->{type};
         printf "%s.min 0\n", $id;
+        printf "%s.type DERIVE\n", $id;
     }
 }
 


kenyon (changes requested):
`graph_vlabel` should also change to `messages/${graph_period}`

iasjennen:
@kenyon 

when i change

...
    print "graph_vlabel count\n";
...
to
...
    print "graph_vlabel messages/${graph_period}\n";
...

and i run:
 munin-run syslog_ng_stats config

i get the error:
Global symbol "$graph_period" requires explicit package name (did you forget to declare "my $graph_period"?) 

Please give more details how to change.

kenyon:
You need to either use single quotation marks, or escape the dollar sign so that perl doesn't try to do variable interpolation. The variable `${graph_period}` is for munin.

kenyon (approved):

PR #1460, opened 2024-10-15
foomuuri: plugin to graph Foomuuri statistics

kimheino:
Add new plugin to graph statistics from Foomuuri. Foomuuri is a multizone bidirectional nftables firewall.

See: https://github.com/FoobarOy/foomuuri

I'm the author of Foomuuri.

### DIFF
--- a/plugins/foomuuri/foomuuri
+++ b/plugins/foomuuri/foomuuri
@@ -1,0 +1,228 @@
+#!/usr/bin/env python3
+
+"""Munin plugin to graph Foomuuri statistics.
+
+=head1 NAME
+
+foomuuri - graph Foomuuri statistics.
+
+=head1 APPLICABLE SYSTEMS
+
+Linux systems with Foomuuri running.
+
+=head1 CONFIGURATION
+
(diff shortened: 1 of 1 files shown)


kenyon (commented):
What do you think about putting this under the `network` directory rather than starting a new directory containing a single plugin?

kimheino:
I was about to do that first, but then I saw that there is `shorewall` directory (another firewall generator) with only one plugin in it. Actually there are *lot* of these one plugin directories.

I'm not sure what is the best way. Having lot of plugins in `network` directory is not optimal either, especially with these "specific to single program only" -type plugins.

I'll let you to decide. I'm happy to move it to network, or keep it in foomuuri.

kenyon:
Yeah, I don't really like the proliferation of single-plugin directories, but I'm also probably not going to do the work to consolidate them. I think that the directory should tend to match the `graph_category`, but plugins can also have multiple categories so that pattern breaks down a bit. No strong feelings about this and it's easy enough to search regardless of the layout, so I'll just merge this.

PR #1466, opened 2024-12-29
add naxsi exceptions plugin

patroqueeet:
plugin to monitor naxsi WAF


kenyon (changes requested):
Let's put this in a better directory name, like `naxsi` or `waf`.

patroqueeet:
@kenyon done and also made warning/critical levels configurable

kenyon (approved):

patroqueeet:
@kenyon thx for merging. when will the live site plugin gallery be updated, so we can link it in the naxsi waf docs?

kenyon:
@patroqueeet not sure about that. Will see if I can find out. For now, linking to this github repository should be fine.

PR #1469, opened 2025-01-08
feat: add turbostat plugin

pimlie:
See embedded plugin docs for description etc

It might be weird that this is a multigraph plugin with only one graph, but wanted to be prepared for if I overlooked something other useful to monitoring using turbostat. Feel free to let me know if you have suggestions which graphs to add, but things like temperatures can already be monitored using the `sensors_` plugin which uses the same source so imho not really useful to add those in this plugin too


plugins/sensors/turbostat_
```diff
@@ -0,0 +1,217 @@
...
+  for e; do [[ "$e" == "$match" ]] && return 0; done
+  return 1
+}
+
+# Join array elements by separator
+function join {
+  local IFS="$1"
+  shift
+  echo "$*"
+}
```
kenyon, line 173: Indentation style for these two functions doesn't match the rest of the file. Do you want to fix that?

kenyon (approved):

PR #1476, opened 2025-01-26
[jellyfin_sessions] Simple plugin to check active sessions on a jellyfin media system

brknkfr:


plugins/jellyfin/jellyfin_sessions
```diff
@@ -0,0 +1,86 @@
...
+
+    config)
+        echo "graph_title Sessions on Jellyfin system"
+        echo "graph_vlabel sessions"
+        echo "sessions.label sessions"
+        exit 0
+    ;;
+esac
+
+if active_sessions=$(curl -s -f -m "$max_time" -H "Authorization: MediaBrowser Token=$api_key" -X GET "$url_sessions"?activeWithinSeconds="$interval" | jq '. | length') | grep -E "[0-9]*"; then
```
kenyon, line 81: Don't you want to match one or more digits? The `*` matches zero or more digits.

```suggestion
if active_sessions=$(curl -s -f -m "$max_time" -H "Authorization: MediaBrowser Token=$api_key" -X GET "$url_sessions"?activeWithinSeconds="$interval" | jq '. | length') | grep -E "[0-9]+"; then
```

PR #1485, opened 2025-03-02
nvidia_gpu_: fix utilization when using 570.x driver

tofurky:
The utilization used to be on a line beginning with "Gpu", but has since been capitalized to "GPU". There's already several other instances of that in the nvidia-smi output, so narrow down the search by returning only the line after "Utilization".


tofurky:
old (from the web):
```
    Utilization
        Gpu                               : 0 %
        Memory                            : 0 %
        Encoder                           : 0 %
        Decoder                           : 0 %
```

new:
```
    Utilization
        GPU                               : 33 %
        Memory                            : 3 %
        Encoder                           : 0 %
        Decoder                           : 0 %
        JPEG                              : 0 %
        OFA                               : 0 %
```

PR #1494, opened 2025-04-18
[disk/file_age] Add support for host_name

cweiske:
.. so the plugin can be grouped into a different host, e.g. backup file timestamps to the "backup" server host.


kenyon (approved):

cweiske:
Thank you for merging!

PR #1521, opened 2026-03-06
postfix_mail_stats1: Fix date grepping

### LINKED ISSUE #1520: postfix_mail_stats1 does't work on German server in march
`postfix_mail_stats1` worked fine for february 2026, but stopped working on first of march. It also does not work on days that have a single digit (1-9).

Reason is that the log file has dates in this format:
```
Mar 06 16:57:22 mail postfix/smtpd[990077]
```

and the plugin uses `date` with the wrong parameters:
```
$ date '+%b %e %H'
Mär  6 16
```
- the locale is used, so "Mar" and "Mär" don't match
- a single-digit day name is used, so "06" does not match

# Screenshot
Image

cweiske:
- use english month names, so that German ones "Mär" do not collide with english ones in the log file "Mar"
- use a two-digit day name as the logs contain.

This is with Debian 13 and postfix 3.10.5-1~deb13u1

Resolves: https://github.com/munin-monitoring/contrib/issues/1520

### DIFF
--- a/plugins/postfix/postfix_mail_stats1
+++ b/plugins/postfix/postfix_mail_stats1
@@ -21,7 +21,7 @@
 
 
 LOGFILE=${logfile:-/var/log/maillog}	# Allow user to specify logfile through env.logfile
-DATE=`date '+%b %e %H'`
+DATE=`LC_ALL=C date '+%b %d %H'`
 MAXLABEL=20
 
 if [ "$1" = "autoconf" ]; then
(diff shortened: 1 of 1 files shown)


cweiske:
This date-format-changed problem also occured with `pflogsumm`, tool to summarize postfix logs: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1068425 and https://athena.linxnet.com/downloads/ChangeLog-1.1.6

kenyon (approved):

cweiske:
Thanks for reviewing.

